PR #1, opened 2019-12-10
A name has been chosen!

Mullefa:
Repository name chosen to be "gateway" (Jess); I've updated the files accordingly. Note domain name profile.theguardian.com hasn't been changed yet, as domain names haven't been decided.


paddyo:
All Hail The Gateway!

PR #11, opened 2020-06-19
Set up AWS Cloudwatch Metrics

coldlink:
## What does this change?
- Add ability to track metrics to cloudwatch from gateway
- Use non-blocking promise to send metrics and catch and log failures
- Add metrics for send password and change password


src/server/lib/AWS.ts
```diff
@@ -0,0 +1,50 @@
...
+export const trackMetric = (
+  metricName: Metrics,
+  dimensions: MetricDimensions = { Stage },
+) =>
+  CloudWatch.putMetricData({
+    Namespace: 'Gateway',
+    MetricData: [
+      {
+        MetricName: metricName,
+        Dimensions: Object.entries(dimensions).map(([Name, Value]) => ({
```
QuarpT, line 38: in the Scala services we have default dimensions ApiMode (which is the name of the service) and Stage which always get added to any metric sent. It'd be cool to do the same here. The dimensions parameter should probably append to these rather than override them.

coldlink: thanks, i've added the ApiMode, as well as the ability to append to (and override if necessary) the dimensions
https://github.com/guardian/gateway/pull/11/commits/9c71369d5e90fd14117372a572f3c1cc71a3dce4

PR #17, opened 2020-06-30
Documentation

coldlink:
## What does this change?
Adds more comprehensive documentation for developers.

Documentation:
- Update README
- Add Architecture doc
- Add Setup doc
- Add Development help doc

Other changes:
- Add recommended VSCode extensions
- Change default port to `8861` for nginx
- Abstract client routes into it's own file
- Add ability to change page title through the server side renderer


.vscode/settings.json
```diff
@@ -0,0 +1,3 @@
+{
```
QuarpT, line 1: Putting local environment IDE settings in source control is often objected. Is this common for VS Code projects?

docs/architecture.md
```diff
@@ -0,0 +1,53 @@
...
+
+Many other client side applications at The Guardian now are React app based, which allows for easier code resuse and knowledge sharing.
+
+## Technology
+
+Gateway is primarily a [TypeScript](https://www.typescriptlang.org/), [React](https://reactjs.org/), and [Express.js](https://expressjs.com/) application, utilising [The Guardian Source Design System](https://theguardian.design/) components, and [Emotion](https://emotion.sh) CSS-in-JS library for UI/design. We use [Jest](https://jestjs.io/) for unit testing, and [Cypress](https://www.cypress.io/) for integration tests (and hopefully E2E tests in the future).
+
+## Browser Support
+
+**Our core line in the sand is that core functionality and features must NOT require any client-side JavaScript.** To this end Gateway is set up as a completely SSR (Server-side Rendered) React application.
```
QuarpT, line 23: except password registration is requires recaptcha in prod right now 😆

docs/development.md
```diff
@@ -0,0 +1,337 @@
...
+res.type('html').send(html);
+```
+
+When a browser makes a request to gateway, routes must render an output, this goes for all request methods (`GET`, `POST` etc.). This is so that redirects can be avoided since it makes it difficult to persist state, and there are some bugs with redirects on certain browsers and devices.
+
+## Routing
+
+Due to SSR, routing is a bit complicated than simply using client-side routing. We do routing through a combination of Express.js and React Router.
+
+To make it easier to keep track of routes/paths, we define then in a `Routes` enum in the [shared/mdel/Routes.ts](../src/shared/model/Routes.ts) file. The enum is then consumed by both Express and React Router so we can share the paths between them.
```
QuarpT, line 30: typo on `mdel`

.vscode/settings.json
coldlink: good point, i did it with the best intentions, but i see why it could be a problem, i'll remove it since it's not that important to have anyway

PR #18, opened 2020-07-13
Dependency Update - July 13 2020

coldlink:
## What does this change?
Update as many node dependencies as possible to the latest version to stay up to date on functionality, security etc.

All functionality still works as expected.

Breaking change to the functionality, will be caught by unit tests, and integration tests. In production monitoring for errors.


.eslintrc.json
```diff
@@ -13,7 +13,8 @@
   },
   "rules": {
     "react/prop-types": 0,
-    "@typescript-eslint/explicit-function-return-type": 0
+    "@typescript-eslint/explicit-function-return-type": 0,
+    "@typescript-eslint/explicit-module-boundary-types": 0,
```
coldlink, line 17: https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/explicit-module-boundary-types.md

Although not required to disable this rule, since it only shows a warning, have done so as compiler should implicitly infer the types, and we have so many methods without explicity typing that it makes it difficult to fully implement.

webpack.config.js
```diff
@@ -23,7 +23,7 @@ const server = {
   entry: "./src/server/index.ts",
   externals: [
     nodeExternals({
-      whitelist: [/^@guardian/]
+      allowlist: [/^@guardian/]
```
coldlink, line 26: breaking change when updating `webpack-node-externals`

#BLM

PR #27, opened 2020-08-12
Update CMP for TCFv2/CCPA

coldlink:
## What does this change?
- Latest version of CMP for TCFv2
- Add `getCountryCode` which is required for the new version of the CMP
- `copyNodeModules` script to stop including `@guardian` libraries from the production build `node_modules`


src/client/static/index.tsx
```diff
@@ -12,8 +21,12 @@ import './analytics/ophan';
 // initialise google analytics
 gaInit();
 
-// load cmp if it should show
-// use dymamic import to reduce delivered bundle size
-if (shouldShow()) {
-  import('./cmp');
+// don't load this if running in cypress
+if (!window.Cypress) {
```
sndrs, line 25: you could use [webpack define plugin](https://webpack.js.org/plugins/define-plugin/) to only output this code if `process.env.NODE_ENV === 'test'` if that's useful

src/client/static/countryCode.ts
```diff
@@ -0,0 +1,66 @@
...
+              `getCountryCode | An api call returned HTTP status ${response.status}`,
+          );
+        }
+        return response;
+      })
+      .then((response) => response.json())
+      .then((json) => json.country);
+
+    // What's this setTimeout business?
+    // localStorage calls are syncronous and we don't need to wait for this
```
paddyo, line 42: Is this really necessary? (Not a blocker to release)

PR #30, opened 2020-08-21
Newsletters page

paddyo:
## Description
Adds initial styling and newsletter data retrieval functionality to the new consents flow. This uses placeholder assets for images that are not ready. Does not feature geolocated newsletters, so the default newsletter set is used.

<p align="center">

</P>

## Changes
* Add `Newsletters` page
* Add `server/lib/idapi/newsletters` adapter library for the IDAPI newsletters endpoint.
* Add `/util/idapi-proxy.js` for allowing easy development vs code IDAPI.
* Add `NewslettersCard` component.
* Fix bug in grid on IE11 that caused grid items to span too much.
* Fix bug in Lines component that prevented display in IE11


src/shared/model/Routes.ts
```diff
@@ -8,6 +8,7 @@ export enum Routes {
   VERIFY_EMAIL = '/verify-email',
   VERIFY_EMAIL_TOKEN = '/:token',
   CONSENTS = '/consents',
+  CONSENTS_NEWSLETTERS = '/consents/newsletters',
```
coldlink, line 11: didn't think we want to commit the routes at the moment

coldlink (approved):

PR #39, opened 2020-09-14
only allow TLSv1.2+

QuarpT:
## What does this change?

Stop supporting SSL TLS versions 1.0 and 1.1 and weak stream ciphers, see https://trello.com/c/NgUVNH05/2072-weak-ssl-tls-ciphers-insufficient-transport-layer-protection-s

See https://docs.aws.amazon.com/elasticloadbalancing/latest/classic/elb-security-policy-table.html for the ciphers and SSL protocols supported

## How to test

Test the API is still accessible behind Fastly

Stop using insecure SSL protocols


mchv:
@QuarpT I think the change on the trello card is regarding `profile.theguardian.com`and as such was mostly about Fastly. We have already disabled `TLS 1.0` and `1.1` (and enabled TLS `1.3`) on all domains for which `guardian.map.fastly.net` applies which includes `profile.theguardian.com`. I think it is still worth to only allow TLS 1.2+ on the ELB however 👍

QuarpT:
👍  @mchv
yeah, speaking to Dom we decided to do this for our load balancers as well as the Fastly changes that were already made 😄

rebecca-thompson (approved):

PR #42, opened 2020-10-02
Make error links customisable and change password link

rebecca-thompson:
## What does this change?
Currently all errors have a 'Report this error' link which links to the `/info/tech-feedback page`, but in cases such as when a user's passwords don't match in the password reset flow, this is the wrong link and should instead point to the `/help/identity-faq` page. This PR makes the help link and link text customisable for each error, defaulting to the tech-feedback page where no link or link text has been specified.

## How to test
This can be tested in `CODE` by going through the reset flow and entering non-matching passwords

Fewer people should contact user help when they can't log in because of user input error.


coldlink (approved):
Nice work!

PR #43, opened 2020-10-08
Add Ophan tracking to Gateway

rebecca-thompson:
## What does this change?
This PR expands on the Ophan tracking in Gateway by adding the `gateway` to the `experiences` parameter and adding a static gateway `abTestRegister` entry. When we start running actual AB tests the latter field will be populated by the variants of those tests, but for now we are using it to differentiate between gateway and non-gateway traffic.

These parameters correspond to the `experiences` and `ab` fields in the Event module in Thrift - https://dashboard.ophan.co.uk/docs/thrift/event.html

There is also a small change to the docker compose file to get the local DEV environment working again.

## How to test
You can run gateway locally and monitor the network tab in developer tools for ophan traffic to see the parameters correctly coming through on the requests

### DIFF
--- a/src/client/static/analytics/ophan.ts
+++ b/src/client/static/analytics/ophan.ts
@@ -1,0 +1,56 @@
+import 'ophan-tracker-js';
+
+// the 'guardian.ophan' object is added by ophan-tracker-js
+// we extend the Window interface here to keep Typescript happy
+interface OphanWindow extends Window {
+  guardian: {
+    ophan: {
+      setEventEmitter: () => void;
+      trackComponentAttention: (
+        name: string,
+        el: Element,
+        visiblityThreshold: number,
+      ) => void;
+      record: ({}) => void;
+      viewId: string;
+      pageViewId: string;
+    };
+  };
+}
+
+declare let window: OphanWindow;
+
+interface OphanEvent {
+  experiences: string;
+  abTestRegister?: { [testId: string]: TestData };
+}
+
+interface TestData {
+  variantName: string;
+  complete: boolean;
+  campaignCodes?: Set<string>;
+}
+
+const record = (event: OphanEvent) => {
+  if (
+    window.guardian &&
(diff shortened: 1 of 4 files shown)


rebecca-thompson:
[Add gateway to the experiences section of the pageview event](https://trello.com/c/m3tG6RLA/2512-add-gateway-to-the-experiences-section-of-the-pageview-event)

coldlink (approved):

src/client/static/analytics/ophan.ts
```diff
@@ -0,0 +1,56 @@
...
+
+interface TestData {
+  variantName: string;
+  complete: boolean;
+  campaignCodes?: Set<string>;
+}
+
+const record = (event: OphanEvent) => {
+  if (
+    window.guardian &&
```
QuarpT, line 36: having called `import 'ophan-tracker-js';` I think these are set synchronously so we shouldn't have to test if record exists 🤔

QuarpT: ah I see this is done in dotcom rendering. Maybe this is what we're supposed to do.
We shouldn't throw exceptions since failure to init ophan shouldn't prevent the page loading.

I don't know whether it's possible for `record` to throw exceptions, which we'd need to recover from

rebecca-thompson: I checked this locally by removing the import line. An error is thrown in the console, but it does not prevent the page from loading so it should not affect the user

QuarpT (approved):

src/client/static/analytics/ophan.ts
QuarpT: an exception would prevent the JavaScript after `ophanInit();`from being executed

rebecca-thompson: I take your point. If the Error is not adding any value, I think it can be removed

PR #44, opened 2020-10-09
Hydrate react 

QuarpT:
## What does this change?

- Hydrate react in preparation for reactive clientside password validation
- run lint fix

## Why?

We need to run component JavaScript on the browser for password validation. Hyrdating the react components allows us to do this.


coldlink:
We have some hydrating code already in [https://github.com/guardian/gateway/blob/main/src/client/index.tsx](https://github.com/guardian/gateway/blob/main/src/client/index.tsx) so it's worth removing the file i linked as an addition to this PR

coldlink:
it may also be worth updating the development readme sections about SSR, Client Side Scripts etc too: https://github.com/guardian/gateway/blob/main/docs/development.md

coldlink:
Another thing, how much does client side hydration add to the client bundle size? So the size before and after adding in hydration. That would be useful for comparison.

coldlink:
A wider conversation outside of this PR is that should we switch to [Preact](https://preactjs.com/) instead of using React. [DCR](https://github.com/guardian/dotcom-rendering) is a large project which already uses Preact as a drop in replacement.

src/client/components/GlobalState.tsx
```diff
@@ -7,7 +7,7 @@ type GlobalProps = {
   globalState: GlobalState;
 };
 
-export const GlobalStateProvider: FunctionComponent<GlobalProps> = props => {
+export const GlobalStateProvider: FunctionComponent<GlobalProps> = (props) => {
```
dominickendrick, line 10: I've tried looking for `prettier` and `editorconfig` rules to specify bracket style, but I've not had any luck. Is there a way we can force the requirements for brackets for function arguments ?

QuarpT: I think they are enforced by prettier. `yarn lint` fails when the brackets aren't present.

dominickendrick: ah I see, at least it will be consistent going forward 👍

QuarpT:
> Another thing, how much does client side hydration add to the client bundle size? So the size before and after adding in hydration. That would be useful for comparison.

old bundle size: 149 KiB
new bundle size: 368 KiB

src/client/components/GlobalState.tsx
QuarpT: we could add `yarn lint` to CI

dominickendrick: ahh I see, I assumed it was run on a `pre-commit` hook or something. Yeah it should be in CI and a pre-commit hook if possible. Maybe one for another PR tho

QuarpT:
> it may also be worth updating the development readme sections about SSR, Client Side Scripts etc too since this is adding in the hydration: https://github.com/guardian/gateway/blob/main/docs/development.md
> 
> Specifically to mention that core functionality should work without client side scripts.
> e.g. for the password validation stuff youre doing, you can validate the password client side, but if scripts aren't enabled then the api in gateway should catch there's an issue with the password validation and return an error that way, so although having script enabled will have a better UX (e.g. with the live pattern matching/show hide password), it should still be validated if they dont

👍  so for the password validation work, only render the validating components on the client. If JavaScript isn't enabled then they don't get shown.

coldlink:
> > Another thing, how much does client side hydration add to the client bundle size? So the size before and after adding in hydration. That would be useful for comparison.
> 
> old bundle size: 149 KiB
> new bundle size: 368 KiB
> 
> Discussing the use of preact is a good idea. Also, we should decide whether the new bundle size is a blocker

Yeah, let's discuss this at the engineering meeting after standup on Monday. There may be ways around this, e.g. dynamic importing of certain components etc. Or maybe there might be an less impactful way of doing client side stuff.

coldlink (approved):
We're approving this for now, with a view to further increase optimisations at a later date.

PR #47, opened 2020-10-13
Add cloudwatch metrics for verify, consents and login middleware

QuarpT:
## What does this change?

Add CloudWatch metrics for verify, consents and login middleware

## Why?

These metrics can be added to the Identity health dashboard. They will gives developers an idea of whether the service is healthy or not.


coldlink (approved):
Looks great to me!

PR #48, opened 2020-10-13
Return URL Functionality

paddyo:
## Description
Adds the returnUrl functionality that returns a user to the page they were on when they began the sign up process.


coldlink (approved):
Great!

A note for the future in case we eventually need to pass on more parameters, it may be useful to have a method which generates query parameter string based on an object and append that onto the path instead.

PR #49, opened 2020-10-13
Fix flakiness on gateway startup

QuarpT:
## What does this change?

Retry get parameter store values on instance initialisation

# Why?

The SSM command can fail when called immediately on instance startup, so we keep calling it until it is successful.

This is causing RiffRaff deploys to frequently fail, and can potentially cause issues replacing unhealthy instances or scaling up.

### DIFF
--- a/cloudformation.yaml
+++ b/cloudformation.yaml
@@ -172,6 +172,16 @@ Resources:
             touch /var/log/identity-gateway.log
             chown identity-gateway:identity-gateway /var/log/identity-gateway.log
 
+            # Try multiple times to get parameter store values. The SSM command can fail when called immediately on instance startup.
+            while true; do
+                if \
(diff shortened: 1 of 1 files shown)


coldlink (approved):

cloudformation.yaml
```diff
@@ -172,6 +172,16 @@ Resources:
             touch /var/log/identity-gateway.log
             chown identity-gateway:identity-gateway /var/log/identity-gateway.log
 
+            # Try multiple times to get parameter store values. The SSM command can fail when called immediately on instance startup.
+            while true; do
```
paddyo, line 176: Maybe a sleep and a kill switch?

QuarpT: I've added a sleep.

If parameter store isn't available it's probably okay for this to loop forever, writing to stderr until the instance is deemed unhealthy

paddyo (approved):

PR #50, opened 2020-10-14
Onboarding | Add Ophan interaction tracking for newsletters and consents

coldlink:
## What does this change?
As part of the onboarding flow we want to see what users consent/subscribe to in the data lake. This PR adds an ophan interaction event to fire for each consent/newsletter after clicking "Save and continue".

We also fix a small bug on the newsletters page where it was not possible to uncheck a newsletter after saving it.

We also make a global window declaration file to extend the window object for typescript to stay happy.

### Changes list
- Add `client/lib/ophan` file for handling ophan events
   - We still need the `client/static/ophan` file to initialise ophan and send the initial pageview event before hydration occurs
- Add ophan types from guardian/types repository
- Add `OphanInteraction` type
- Add `consentsTracking` file with methods for ophan tracking on the onboarding flow
- Add `onSubmit` handler to form on consents layout to perform the tracking
- Add `global` `window` declaration file
- Add `nameId` to the `NewsLetter` interface which refers to the kebab cased name of the newsletter from IDAPI
- Fix newsletter card checkbox bug
- Linting


src/client/layouts/ConsentsLayout.tsx
```diff
@@ -59,6 +60,9 @@ export const ConsentsLayout: FunctionComponent<ConsentsLayoutProps> = ({
         css={form}
         action={`${Routes.CONSENTS}/${page}${returnUrlQuery}`}
         method="post"
+        onSubmit={(e) => {
```
dominickendrick, line 63: can `e` be changed to `event` ?

src/client/lib/consentsTracking.ts
```diff
@@ -0,0 +1,79 @@
...
+export const onboardingFormSubmitOphanTracking = (
+  page: string,
+  pageData: PageData,
+  event: React.FormEvent<HTMLFormElement>,
+) => {
+  const form = event.target as HTMLFormElement;
+  const inputElems = form.querySelectorAll('input');
+
+  switch (page) {
+    case Routes.CONSENTS_COMMUNICATION.slice(1):
```
dominickendrick, line 71: is it worth putting a comment that you are using the `slice` to remove the slash from the beginning of the route path. ? Or could you 
```
switch ("/" + page) { 
     case Routes.CONSENTS_COMMUNICATION
     ...
}`

File: src/client/lib/consentsTracking.ts [hunk 3]
```diff
@@ -0,0 +1,79 @@
...
+    case Routes.CONSENTS_DATA.slice(1):
+      return consentsFormSubmitOphanTracking(inputElems, pageData);
```
dominickendrick, line 73: Does this need to `return`?

src/client/lib/consentsTracking.ts
```diff
@@ -0,0 +1,79 @@
+import { PageData } from '@/shared/model/GlobalState';
+import { Routes } from '@/shared/model/Routes';
+import { sendOphanInteractionEvent } from './ophan';
+
+const trackInputElementInteraction = (
+  inputElem: HTMLInputElement,
+  component: string,
+  consentName: string,
+) => {
```
dominickendrick, line 9: Should the return type be `void` for all of these functions ?

src/client/lib/consentsTracking.ts
```diff
@@ -0,0 +1,79 @@
...
+// handle newsletter form submit event
+const newslettersFormSubmitOphanTracking = (
+  inputElems: NodeListOf<HTMLInputElement>,
+  pageData: PageData,
+) => {
+  const newsletters = pageData.newsletters;
+
+  if (newsletters) {
+    inputElems.forEach((elem) => {
+      const newsletter = newsletters.find((c) => c.id === elem.name);
```
dominickendrick, line 52: could you destructure here ?  `find ( ({id}) => id === elem.name` 

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/find#Using_arrow_function_and_destructuring

src/client/layouts/ConsentsLayout.tsx
coldlink: Destructuring `e` to `{ target: form }`, so that the `form` variable is the `event.target` which is what we're interested in

coldlink: and passing the `form` variable like so:

```ts
onboardingFormSubmitOphanTracking(
            page,
            pageData,
            form as HTMLFormElement,
          );
```

src/client/lib/consentsTracking.ts
coldlink: i like that solution:

i'll switch (😉) to using
```ts
  switch (`/${page}`) {
    case Routes.CONSENTS_COMMUNICATION:
   ...
```

src/client/lib/consentsTracking.ts
coldlink: No, I'll add a `void` to everything as the expected return

src/client/lib/consentsTracking.ts
coldlink: Yes, I'll add this to the type

src/client/lib/consentsTracking.ts
coldlink: Great idea!

src/client/lib/consentsTracking.ts
coldlink: i use return, as it returns `void`, thus avoiding the `break` statement

dominickendrick (approved):

PR #51, opened 2020-10-15
Webpack/Babel client bundle changes

coldlink:
## What does this change?
- Transpile code from `node_modules` on client to ES5 too which makes it compatible with IE11
- Use `core-js` to polyfill functionality as required, e.g. `Object.entries`
- Use split chunk and runtime chunk to seperate out vendor code from client code, and runtime chunk for shared code
- Adds `contenthash` to output client js bundle for better caching and cache busting
- Use `assets-webpack-plugin` to generate json with location and name of hashed bundles to be used by the server side renderer
- Read from file this generated json on the server to get links to the hashed bundles.

## Why
The previously generated bundle was not compatible with IE11 as it was only transpiling code from our `src` directly, and not anything that we imported from `node_modules` (e.g. the CMP banner, Source Design System). This would result in code being pulled into the bundle that was incompatible with older browsers that didn't support ES6 syntax (like IE11).

This wasn't too much of problem before as we were not doing client side hydration, but as some non-core functionality now requires javascript to run (e.g. analytics/password validation), we need this to be compatible with these browsers too.

We also split out vendor and client code, and add hashes onto the file names for better caching.

## Notes
As a result of these changes the total bundle size has increased to 518KB, with the vendors bundle accounting for 418KB. So further optimisations will be needed in the future.


QuarpT (commented):
lgtm

dominickendrick (approved):

PR #53, opened 2020-10-15
Password eye

QuarpT:
## What does this change?

- Allow user to show and hide password by clicking an eye symbol
- Only show the symbol on certain browsers which don't already show an overlay

## Why?

users might want to know what they've typed

## Video

ezgif com-gif-maker (2)

## notes

target branch is set to `password-reset-validation`
release `password-reset-validation` first, then change target branch to `main`


src/server/lib/renderer.ts
```diff
@@ -23,12 +25,20 @@ interface RendererOpts {
...
-  opts = {},
-) => {
+export const renderer: (
+  url: string,
+  request: Request,
+  opts?: RendererOpts,
+) => string = (url, request, opts = {}) => {
   const { globalState = {}, queryParams = {}, pageTitle = 'Gateway' } = opts;
 
+  try {
```
QuarpT, line 35: set browserName on the server to avoid adding bowser as a clientside dependency
(it would add 25KB to the bundle)

src/server/lib/renderer.ts
```diff
@@ -23,12 +25,20 @@ interface RendererOpts {
 
 const { gaUID } = getConfiguration();
 
-export const renderer: (url: string, opts?: RendererOpts) => string = (
-  url,
-  opts = {},
-) => {
+export const renderer: (
+  url: string,
+  request: Request,
```
coldlink, line 30: Rather than pass the Request in every `renderer` call, it's worth adding it to the `RendererOpts` as an optional value, and then only passing in the `request` in when we need the `browserName` on the global state on a particular route.

coldlink: Even better would be to add parse and add the browserName to the globalState only on the route that its needed directly, and have the logic for getting the user agent/brower in its own lib file that can be called.

QuarpT: if we make it a property on the global state, is there any advantage to only setting it for specific routes? 
The overhead for setting it is negligible

coldlink: yeah it's negligible, imo it would provide more clarity to whats in the globalState, and opt in as needed

the main concern I have at the moment is that i feel that the renderer should not need the whole request object to render the html, it should only rely on things we pass into it in it's options parameter

PR #54, opened 2020-10-16
Marketing consents to use checkboxes.

paddyo:
## Description
Uses the checkbox style marketing consents with the corresponding text from IDAPI

## Changes
* Replace radio components with checkbox components on Data as well as Communications pages.
* Marketing consents on Data and Communications pages now uses description from IDAPI.
* Change `ConsentsLayout` to hide the save and continue button if the page is in error state.

### DIFF
--- a/src/client/layouts/ConsentsLayout.tsx
+++ b/src/client/layouts/ConsentsLayout.tsx
@@ -77,14 +77,16 @@ export const ConsentsLayout: FunctionComponent<ConsentsLayoutProps> = ({
         </main>
         <ConsentsBlueBackground>
           <div css={[gridItem(gridItemColumnConsents), controls]}>
-            <Button
-              iconSide="right"
-              nudgeIcon={true}
-              icon={<SvgArrowRightStraight />}
-              type="submit"
-            >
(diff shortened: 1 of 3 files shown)


src/client/layouts/ConsentsLayout.tsx
```diff
@@ -77,14 +77,16 @@ export const ConsentsLayout: FunctionComponent<ConsentsLayoutProps> = ({
...
           <div css={[gridItem(gridItemColumnConsents), controls]}>
-            <Button
-              iconSide="right"
-              nudgeIcon={true}
-              icon={<SvgArrowRightStraight />}
-              type="submit"
-            >
-              Save and continue
-            </Button>
+            {!error && (
```
coldlink, line 80: i'm worried this could mean that users get stuck on this page if there is an error?

coldlink (approved):

src/client/layouts/ConsentsLayout.tsx
paddyo: Previously the button posted to `/consents/` which doesn't exist.

PR #61, opened 2020-10-22
Onboarding | Integration Tests

coldlink:
## Description
First set of integration tests and utilities to support them. All further tests to done in separate PRs and tickets


cypress/integration/onboarding_flow.spec.js
```diff
@@ -0,0 +1,333 @@
...
+      cy.url().should('include', `returnUrl=${returnUrl}`);
+
+      ReviewPage.getBackButton().should('not.exist');
+      ReviewPage.getSaveAndContinueButton().should('not.exist');
+
+      ReviewPage.getNewslettersSection().contains('N/A');
+      ReviewPage.getConsentsSection().contains('N/A');
+
+      ReviewPage.getMarketingResearchChoice().contains('Yes');
+      ReviewPage.getMarketingAnalysisChoice().contains('Yes');
```
coldlink, line 239: If we're opting out of all consents in this flow, should this be `No` instead?

coldlink:
apart from that I approve 👍

cypress/integration/onboarding_flow.spec.js
paddyo: Yes good catch. Fixed now

paddyo (approved):

PR #80, opened 2020-10-30
Add CSRF token validation for all Gateway forms

QuarpT:
## What does this change?

Add CSRF token validation for all Gateway forms. Any POST requests will fail unless the form includes the `<CsrfFormField />` component. I've added this component to all existing forms.

I've add the config option `IS_HTTPS` to set the secure cookie attribute depending on whether we are developing locally or not. I do not rely on the `STAGE` parameter since this is set to production on the CI tests.

## Why

We want to prevent CSRF attacks on Gateway

https://owasp.org/www-community/attacks/csrf
https://cheatsheetseries.owasp.org/cheatsheets/Cross-Site_Request_Forgery_Prevention_Cheat_Sheet.html


src/server/routes/changePassword.ts
```diff
@@ -167,6 +167,7 @@ router.get(
   (_: Request, res: ResponseWithLocals) => {
     const html = renderer(Routes.RESET_RESEND, {
       pageTitle: PageTitle.RESET_RESEND,
+      locals: res.locals,
```
QuarpT, line 170: we need to propagate res.locals for all routes because the CSRF token is set on locals

src/server/lib/middleware/errorHandler.ts
```diff
@@ -0,0 +1,17 @@
+import { NextFunction, Request, Response } from 'express';
+
+export const routeErrorHandler = (
```
QuarpT, line 3: maybe I should remove this and return the error immediately from src/server/lib/middleware/csrf.ts

src/server/lib/middleware/csrf.ts
```diff
@@ -0,0 +1,29 @@
...
+  next: NextFunction,
+) => {
+  res.locals.csrfToken = req.csrfToken();
+  next();
+};
+
+export const addCsrfMiddleware = (server: Express) => {
+  server.use(csrfMiddleware);
+  server.use(csrfLocalsMiddleware);
+};
```
coldlink, line 29: middleware can just be an array, so don't really need a function to do this

from the express [docs](https://expressjs.com/en/guide/using-middleware.html):

```js
function logOriginalUrl (req, res, next) {
  console.log('Request URL:', req.originalUrl)
  next()
}

function logMethod (req, res, next) {
  console.log('Request Type:', req.method)
  next()
}

var logStuff = [logOriginalUrl, logMethod]
app.get('/user/:id', logStuff, function (req, res, next) {
  res.send('User Info')
})
```

src/server/lib/middleware/errorHandler.ts
```diff
@@ -0,0 +1,17 @@
...
+  // eslint-disable-next-line
+  err: any, // ErrorRequestHandler uses type any
+  req: Request,
+  res: Response,
+  next: NextFunction,
+) => {
+  if (err.code === 'EBADCSRFTOKEN') {
+    // handle CSRF token errors here
+    res.status(403);
+    res.send('csrf token error');
```
coldlink, line 13: we shouldn't just send back a 'csrf token error', need to think about a better way of handling this and displaying it on a page, possibly through the global error somehow 🤔

lets discuss next week!

.env.example
```diff
@@ -7,3 +7,4 @@ IDAPI_BASE_URL=
 PLAY_SESSION_COOKIE_SECRET=
 DEFAULT_RETURN_URI=
 SIGN_IN_PAGE_URL=
+IS_HTTPS=false
```
coldlink, line 10: think in the example it should be left blank as you can use `localhost` non https, or `thegulocal` with https

you'll also want to update the dev config .env files in the s3 bucket to reflect these new parameters if you haven't done so already

src/server/lib/middleware/errorHandler.ts
QuarpT: I was thinking for genuine users, with the current implementation this error should never propagate.

Although it could get returned if a user was in the middle of a form when we release the change 🤣

src/server/lib/middleware/csrf.ts
```diff
@@ -0,0 +1,27 @@
...
+
+const isHttps = getConfiguration().isHttps;
+
+export const csurfMiddleware = csrf({
+  cookie: {
+    key: '_csrf',
+    sameSite: true,
+    secure: isHttps,
+    httpOnly: true,
+    signed: false, //TODO create a cookie secret for cookie-parser and set to true
```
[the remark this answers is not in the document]
QuarpT: done

src/server/lib/middleware/csrf.ts
QuarpT: done

.env.example
QuarpT: done

.env.example
```diff
@@ -7,3 +7,5 @@ IDAPI_BASE_URL=
 PLAY_SESSION_COOKIE_SECRET=
 DEFAULT_RETURN_URI=
 SIGN_IN_PAGE_URL=
+IS_HTTPS=
+APP_SECRET=
```
QuarpT, line 11: app secret required for generating signed csrf tokens

QuarpT:
changing to draft since we need to think about WIP  https://github.com/guardian/gateway/pull/80/commits/b235cbea0796defd0977f53c74b122f519513164

cloudformation.yaml
```diff
@@ -203,6 +204,8 @@ Resources:
             Environment=PLAY_SESSION_COOKIE_SECRET=$PLAY_SESSION_COOKIE_SECRET
             Environment=DEFAULT_RETURN_URI=${DefaultReturnUri}
             Environment=STAGE=${Stage}
+            Environment=IS_HTTPS=true
+            APP_SECRET=$APP_SECRET
```
coldlink: is there a reason that isn't prefixed with `Environment=`?

src/server/lib/middleware/csrf.ts
```diff
@@ -0,0 +1,49 @@
+import csrf from 'csurf';
+import { NextFunction, Request } from 'express';
+import { ResponseWithLocals } from '@/server/models/Express';
+import { getConfiguration } from '@/server/lib/configuration';
+
+const isHttps = getConfiguration().isHttps;
```
coldlink, line 6: Prefer destructuring:

```suggestion
const { isHttps } = getConfiguration();
```

src/server/lib/middleware/errorHandler.ts
```diff
@@ -0,0 +1,36 @@
...
+const appendQueryParameter = (url: string, parameters: string) => {
+  if (url.split('?').pop()?.includes(parameters)) {
+    return url;
+  }
+  if (url.includes('?')) {
+    return `${url}&${parameters}`;
+  } else {
+    return `${url}?${parameters}`;
+  }
+};
```
coldlink, line 13: this is fine for now, just making a note to see if this is something we can abstract to a library at some point

src/server/lib/middleware/index.ts
```diff
@@ -4,18 +4,24 @@ import compression from 'compression';
 import { helmetMiddleware } from './helmet';
 import { loggerMiddleware } from './logger';
 import { applyRoutes } from './routes';
+import { csrfMiddleware } from '@/server/lib/middleware/csrf';
+import { getConfiguration } from '@/server/lib/configuration';
+
+const appSecret = getConfiguration().appSecret;
```
coldlink, line 10:
```suggestion
const { appSecret } = getConfiguration();
```

coldlink (changes requested):
thanks, just a couple of comments and suggestions before i approve

looking great otherwise

src/client/layouts/ConsentsLayout.tsx
```diff
@@ -84,6 +85,7 @@ export const ConsentsLayout: FunctionComponent<ConsentsLayoutProps> = ({
       >
         <main css={[mainBackground, ieFlexFix]}>
           <ConsentsContent>
+            <CsrfFormField />
```
QuarpT, line 88: I will check how this looks in CODE before releasing

coldlink (approved):

PR #93, opened 2020-11-10
Refactoring state management

coldlink:
## Problem
Our server side rendering method (`renderer`) relies on multiple parameters to be passed in, all of which could be described as a form of a "state". This leads to confusion about how state is managed, and how state is reasoned about between the server and the client.

The `renderer` method relies on 4 parameters:
 - `url` - Route of the page to render, used by React [Router] to determine which path to render.
- `pageTitle` (optional) - Sets the current title of the document/page (what you see in the tab name).
- `GlobalState` (optional) - State that is exposed to the client, and is able to be accessed by React.
- `locals` (optional) - Derived from Express' [`res.locals`](https://expressjs.com/en/api.html#res.locals) to manage a form of server-side state for that request, allows data to be passed between middleware. Adds some items to the `GlobalState` in the `renderer` method (e.g. `geolocation`, `csrf`), and manages tangential things like `queryParams`. 

## Solution
`renderer` method should only rely on 2 inputs, the path/url to render, and some form of a state, the state should be standalone, so we can render without needing a dependency on express.

To achieve this we propose on the server adding all state related stuff to `res.locals`, which is an typed object. This can also be created independently from express if required.

In `renderer` method, derive the `ClientState` from the `ServerState` which is whats exposed to the client

This means we only have a single place for the state on the server (`res.locals`) and a single object is exposed to the client which we can control (`ClientState`)

The new renderer options `opts` in `renderer(url, opts)` looks like:

```ts
interface RendererOpts {
  pageTitle: string;
  serverState: ServerState;
}
```
These are both now required in every call to `renderer`. It was not (easily) possible to decouple the `pageTitle` from the option, as we sometimes need to dynamically set this depending on a condition in the Express route/logic.

## Other Changes
- Renaming `Locals` to `ServerState`
- Renaming `GlobalState` to `ClientState`
- Refactored the `ClientState` and `ServerState` types to make their properties more related, e.g. group anything related to the page in `pageData`, global error/success messages in `globalMessage`
- Add middleware to set a default `res.locals` server state to save on having to create the object each times we need to use it.
- Refactor everything using `ServerState` and `ClientState` to use these new interfaces


src/server/lib/renderer.ts
```diff
@@ -22,47 +21,55 @@ const favicon =
...
+  // checking if csrf error exists in query params, and attaching it to the
+  // forms field errors
+  if (queryParams.csrfError) {
+    // global state page data will already exist at this point
+    // this is just a check to get typescript to compile
+    if (!globalState.pageData) {
+      globalState.pageData = {};
+    }
 
-const appendCsrfFieldError = (globalState: GlobalState) => {
```
QuarpT, line 38: I don't think there's an advantage to inlining this

QuarpT (approved):
looks great

dominickendrick (approved):

dominickendrick (approved):

PR #101, opened 2020-11-16
Implement AB Testing Framework

coldlink:
## What does this change?
Setup [The Guardian's AB Testing Library](https://github.com/guardian/ab-testing) for use in Gateway. Library information on how it works is available there.

## How it works

1. **Define the AB test**: Each AB test and their variants are defined in code with configuration such as audience size & offset and impression & success listeners etc
2. **Initialise the library**: The AB Test library is initialised with configuration values such as a user's MVT ID, an array of the above defined A/B tests etc
3. **Use the AB Test API**: The intialisation returns an API that can be used to check if the current user is in a variant of a test along with a variety of other API methods

## Usage in Gateway

There are 2 ways to use AB Tests on Gateway.

1. **Client Side**: Runs on the client/browser, e.g. for showing different components or taking different actions depending on the test in question. Works with SSR too. This is the easiest way to setup and run tests.
2. **Per Request**: The AB test information and API are also available on the ServerState (res.locals) to be untilised within routes/middleware. This is more complex to setup and run, but for example it gives the option of directing users to different routes/flows based on the test in question. 

## More information

See the new `ab-testing` readme in this PR.


docs/ab-testing.md
```diff
@@ -0,0 +1,74 @@
+# AB Testing Guide
```
dominickendrick, line 1: 👏  great docs !

gtrufitt: +1! And I am very excited to see this being implemented in another platform 🎊

src/client/main.tsx
```diff
@@ -29,6 +49,8 @@ export const Main = (props: ClientState) => {
         `}
       />
       <ClientStateProvider clientState={props}>
+        {/* Uncomment below line for AB Test Demo */}
```
dominickendrick, line 52: Should this comment be there ? It might just be worth adding these instructions to the readme or `ABTestDemo.tsx`

src/client/static/hydration.tsx
```diff
@@ -11,10 +15,24 @@ export const hydrateApp = () => {
 
   const clientState = routingConfig.clientState;
 
+  const {
```
dominickendrick, line 18: What is this line doing ?

src/client/static/hydration.tsx
```diff
@@ -11,10 +15,24 @@ export const hydrateApp = () => {
...
+
   hydrate(
-    <StaticRouter location={`${routingConfig.location}`} context={{}}>
-      <Main {...clientState} />
-    </StaticRouter>,
+    <ABProvider
+      arrayOfTestObjects={tests}
+      abTestSwitches={switches}
+      pageIsSensitive={false}
+      mvtMaxValue={1000000}
```
dominickendrick, line 27: I assume this has to be the same value as that of the mvt config in vcl ? is it worth noting this as a comment so people don't accidentally change it.

src/client/main.tsx
```diff
@@ -1,11 +1,31 @@
...
+// Uncomment below line for AB Test Demo
+// import { ABTestDemo } from './components/ABTestDemo';
 
 export const Main = (props: ClientState) => {
+  // initalise the AB Test Framework:
+  // load the AB Hook
+  const ABTestAPI = useAB();
+
+  // use effect to initialise and register events if needed
+  useEffect(() => {
```
oliverlloyd, line 18: I only about 50% sure here but I'm wondering if this `useEffect` is needed?

src/server/lib/middleware/abTests.ts
```diff
@@ -0,0 +1,110 @@
...
+  runnableTests.forEach((test) => {
+    res.locals.abTesting.participations[test.id] = {
+      variant: test.variantToRun.id,
+    };
+  });
+
+  next();
+};
+
+export const abTestDemoMiddleware = (
```
dominickendrick, line 52: Could this be in another file ?

src/server/lib/middleware/abTests.ts
```diff
@@ -0,0 +1,110 @@
...
+  // {
+  //   ExampleTest: 'variant'
+  // }
+  runnableTests.forEach((test) => {
+    res.locals.abTesting.participations[test.id] = {
+      variant: test.variantToRun.id,
+    };
+  });
+
+  next();
```
dominickendrick, line 49: if only `next()` tool `res` as an argument, we wouldn't need this mutation .....

src/server/lib/middleware/mvtId.ts
```diff
@@ -0,0 +1,26 @@
...
+
+const { stage } = getConfiguration();
+
+export const mvtIdMiddleware = (
+  req: Request,
+  res: ResponseWithServerStateLocals,
+  next: NextFunction,
+) => {
+  // defaults to 0 already
+  let mvtId = res.locals.abTesting.mvtId;
```
dominickendrick, line 13:
```
const mvtId (req) => {
if (req.cookies['GU_mvt_id']) {
    return Number(req.cookies['GU_mvt_id']) || 0;
  }

  if (req.cookies['GU_mvt_id_local'] && stage === 'DEV') {
    return Number(req.cookies['GU_mvt_id_local']) || 0;
  }

 res.locals.abTesting.mvtId = mvtIdFromCookie

next();
}

```

Then you won't have to live with the shame of using a `let` ;)

src/server/lib/renderer.tsx
```diff
@@ -70,16 +80,24 @@ export const renderer: (url: string, opts: RendererOpts) => string = (
 
   const location = `${url}${queryString ? `?${queryString}` : ''}`;
 
+  const {
+    abTesting: { mvtId = 0, forcedTestVariants = {} } = {},
+  } = clientState;
+
   // Any changes made here must also be made to the hydration in the static webpack bundle
```
dominickendrick, line 87: Is it possible to abstract this definition so that is only needs to be changed in a single location ?

src/shared/model/experiments/abTests.ts
```diff
@@ -0,0 +1,41 @@
+import { AB, ABTest, Participations } from '@guardian/ab-core';
+import { switches } from './abSwitches';
+import { exampleTest } from './tests/example-test';
+
+interface ABTestConfiguration {
+  abTestSwitches: Record<string, boolean>;
+  arrayOfTestObjects: ABTest[];
+  pageIsSensitive: false;
+  mvtMaxValue: 1000000;
```
dominickendrick, line 9: As this mvtMaxVaule is used in multiple places, can it be defined as a `CONST` somewhere ?

dominickendrick: Maybe just in this file !

```
export const MVT_MAX_VALUE = 1000000

interface ABTestConfiguration {
  abTestSwitches: Record<string, boolean>;
  arrayOfTestObjects: ABTest[];
  pageIsSensitive: false;
  mvtMaxValue: MVT_MAX_VALUE;
 ...
```

src/client/static/hydration.tsx
coldlink: extracting (destrucuring) the `mvtId` and `forcedTestVariants` from the client state. These values are from the the users cookie and query parameters respectively, which we pass through the server state to the client state.

Since the client state is never guaranteed we give them default values too

src/shared/model/experiments/abTests.ts
coldlink: The interface itself already enforces this:
```ts
interface ABTestConfiguration {
  abTestSwitches: Record<string, boolean>;
  arrayOfTestObjects: ABTest[];
  pageIsSensitive: false;
  mvtMaxValue: 1000000;
}
```

so the mvtMaxValue cannot be set to any other value or else typescript will complain

src/server/lib/renderer.tsx
coldlink: it's just the nature of doing both SSR and CSR, we have to duplicate things in the 2 places so react is able to hydrate correctly after the initial server render.

src/client/main.tsx
coldlink: I mainly followed the guide from https://github.com/guardian/ab-testing#initialising-in-a-react-project

From what I can tell the `useEffect` prevents the tracking/completion/impression events from firing more than once, as we only need them on the initial load.

src/client/static/hydration.tsx
dominickendrick: I see, would a comment explaining that help? Also, how can you just reference `mvtId` later rather than `abTesting.mvtId` 🤯  ??

```
  hydrate(
  <ABProvider
      arrayOfTestObjects={tests}
      abTestSwitches={switches}
      pageIsSensitive={false}
      mvtMaxValue={1000000}
      mvtId={mvtId}
```

dominickendrick: Ah thanks for explaining what about 
```
const mvtId = clientState.abTesting?.mvtId || 0
const forcedTestVariants = clientState.abTesting?. forcedTestVariants || {}
```

dominickendrick (approved):

PR #102, opened 2020-11-16
fix high latency alarm reference

QuarpT:
## What does this change?

Fixes the high latency alarm used by auto scaling. The parameter expects the LB name, not the ARN

### DIFF
--- a/cloudformation.yaml
+++ b/cloudformation.yaml
@@ -363,6 +363,7 @@ Resources:
           Protocol: email
   HighLatencyAlarm:
     Type: AWS::CloudWatch::Alarm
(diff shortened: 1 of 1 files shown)


cloudformation.yaml
```diff
@@ -363,6 +363,7 @@ Resources:
           Protocol: email
   HighLatencyAlarm:
     Type: AWS::CloudWatch::Alarm
+    Condition: IsProd
```
QuarpT, line 366: Send alerts to TopicSendEmail on scaling. 
TopicSendEmail only exists in prod

rebecca-thompson (approved):

PR #103, opened 2020-11-16
Instantiate new default server state on each request

coldlink:
## What does this change?
- Instantiate new server state on each request to default state from being mutated


dominickendrick (approved):

katebee:
Are there any tests that we could add as an extra safeguard against regressions?

PR #111, opened 2020-11-24
Further integration tests

paddyo:
## Description
Further integration tests for testing email validation and login checks using the recommended cypress method for checking 'external' redirects. Doing it the standard way of following the redirect caused problems in non-headless mode such as losing snapshots and test reruns failing.

This is also includes the individual page tests, which mostly focus on preselection of checkboxes and errors, but also the geolocation functionality of the newsletters page.


coldlink (approved):
👍

PR #120, opened 2020-12-02
Middleware functionality to library function migration

paddyo:
## Description
Part I of the migration to an immutable object setup. This involves spinning out as much functionality that extracts values from `req` in a middleware into their own library functions so that the functionality is testable and easier to migrate. This will also fix filenaming and attempt to add tests where missing.

## Changes
* Create `getGeolocationRegion` request parser.
* Create `getMvtId` request parser.
* Create `getCsrfPageUrl` request parser.
* Create `getABForcedVariants` request parser.
* Use request parsers where applicable in middleware.
* Remove unused file/feature `abTestDemo`
* Move `routes` helper library out of `middleware` folder.
* Rename any file exporting a single function to be the name of that function.


src/server/lib/__tests__/getMvtId.test.ts
```diff
@@ -0,0 +1,83 @@
+import { getMvtId } from '@/server/lib/getMvtId';
```
dominickendrick, line 1: 👏  Great to be able to test all this functionality now

src/server/lib/middleware/abTestDemo.ts
```diff
@@ -1,64 +0,0 @@
-import { ResponseWithServerStateLocals } from '@/server/models/Express';
```
coldlink, line 1: this isn't being used, as it's an example file on how to possibly run an test for a given route, so we may still want to keep this, or move it into the docs

paddyo: Good shout, I've moved this code to the A/B Test docs

coldlink (approved):

PR #122, opened 2020-12-04
Centralise res.locals manipulation.

paddyo:
## Description
Part II of the migration to an immutable object setup. This involves making a single place that `res.locals` mutation occurs, as well as potential state mutation which will be exempt from the mutation rules.

## Changes
* Create `requestStateMiddleware` middleware as a single place for generating `res.locals` based state.
* Create `fourZeroFourMiddleware` for handling 404 errors.
* Create `getABTesting` library function for setting up ABTesting.
* Rename type `ServerState` and `ResponseWithServerStateLocals` to `RequestState` and `ResponseWithRequestState` to better reflect the nature of that type (specific state per request).
* Rename `clientStateFromServerStateLocals()` to `clientStateFromRequestStateLocals()`
* Remove `applyRoutes`
* Remove `abTestsMiddleware`
* Remove `geoLocationMiddleware`
* Remove `mvtIdMiddleware`
* Remove `queryParamsMiddleware`
* Remove `serverStateLocalsMiddleware`
* Remove `updateCsrfPageUrlMiddleware`
* Remove `updateCsrfTokenMiddleware`
* `routeErrorHandler` no longer uses `res.locals`.
* Various documentation changes.

## Depends on
* #120


src/server/lib/middleware/requestState.ts
```diff
@@ -0,0 +1,28 @@
...
+import { getConfiguration } from '../getConfiguration';
+import { tests } from '@/shared/model/experiments/abTests';
+import { getABTesting } from '../getABTesting';
+
+export const requestStateMiddleware = (
+  req: Request,
+  res: Response,
+  next: NextFunction,
+) => {
+  const state = getDefaultRequestState();
```
QuarpT, line 17: Can we avoid mutating an object returned by `getDefaultRequestState`?
Instead we could have a function `createDefaultRequestState` that takes the middleware state as parameters, and sets them on object creation.

paddyo: Yes, that's the plan for when I implement the immutability linting (I'll have to do this for other bits of code too)

src/server/lib/renderer.tsx
```diff
@@ -25,20 +25,20 @@ const favicon =
...
-const clientStateFromServerStateLocals = (
+const clientStateFromRequestStateLocals = (
   {
     csrf,
     globalMessage,
     pageData,
     queryParams,
     abTesting,
-  } = getDefaultServerState(),
+  } = getDefaultRequestState(),
```
QuarpT, line 41: is this default ever used?

src/server/lib/middleware/requestState.ts
QuarpT: 👍  I pointed it how here because this is new code.

src/server/lib/renderer.tsx
paddyo: I'm not convinced it is. But because of the scope of changes here, I've made a note of it and I'll look at this when I do AB testing ticket.

QuarpT (approved):
looks great

coldlink (approved):

src/server/lib/renderer.tsx
coldlink: I think this is a leftover from our inital state management, where the state was optional

PR #123, opened 2020-12-04
Implement breached password validation on password reset

QuarpT:
## What does this change?
 - Implement breached password validation on password reset
- see reset password section in: https://www.figma.com/file/HrS0AJGKiUzV83dHr1UZP3/Reset-password-screen?node-id=160%3A260
- Requires an IDAPI release updating the password reset endpoint to check if a password is breached: https://github.com/guardian/identity/pull/1853

## How to test

Test reset password flow in CODE

## Videos showing reset form behaviour

ezgif com-gif-maker (8)
ezgif com-gif-maker (7)
ezgif com-gif-maker (6)


util/mock-server.js
```diff
@@ -42,6 +47,12 @@ app.post('/mock', (req, res) => {
   res.sendStatus(204);
 });
 
+// Mock an endpoint with a pattern
+app.post('/mock/permanent-pattern', (req, res) => {
```
QuarpT, line 51: I've implemented stubbing patterns in our mock server. 

An improvement would be to move away from `idapiMockNext`, where it's very easy to stub something unintended.

src/client/pages/ChangePassword.tsx
```diff
@@ -10,15 +18,240 @@ import { PageBox } from '@/client/components/PageBox';
...
+    isPasswordConfirmSelectedCurrently,
+  ] = useRefState(false);
+
+  // the latest validation result
+  const [
+    validationResult,
+    setValidationResult,
+  ] = useState<PasswordValidationResult>(PasswordValidationResult.AT_LEAST_8);
+
+  // redError is used for the first password input box
```
QuarpT, line 147: I've attempted to describe the some of the UX behaviour, and the reasoning behind the state. I will attach a video to the PR to make it more clear.

src/client/lib/ThrottledBreachedPasswordCheck.ts
```diff
@@ -0,0 +1,73 @@
+import { PasswordValidationResult } from '@/shared/lib/PasswordValidation';
+
+export class ThrottledBreachedPasswordCheck {
+  checkInterval = 500; //ms
```
QuarpT, line 4: throttle requests to breached password endpoint to twice per second, timing out after 1 second

package.json
```diff
@@ -17,7 +17,7 @@
     "test": "yarn type-check && yarn test:unit",
     "test:unit": "jest --ci",
     "type-check": "tsc --noEmit --skipLibCheck",
-    "watch:server": "nodemon ./build/server.js",
+    "watch:server": "nodemon --inspect ./build/server.js",
```
QuarpT, line 20: --inspect added for node debugging

cypress/support/pages/change_password_page.js
```diff
@@ -32,11 +35,19 @@ class ChangePasswordPage {
...
     cy.get('input[name="password"]').type(password0);
     cy.get('input[name="password_confirm"]').type(password1);
+  }
+
+  submitPasswordChange(password0, password1) {
+    this.typePasswordChange(password0, password1);
     this.clickPasswordChange();
   }
+
+  stubBreachedCountEndpoint(response) {
```
paddyo, line 48: I'm not sure Page Objects are the right place for API based stubbing, they are really only to abstract out the page and Cypress' interaction with it. I think mocking should be either restricted to the actual test file or to a dedicated mocking helper file, much like the fixtures are off in their own files.

QuarpT: 👌  
moved in https://github.com/guardian/gateway/pull/123/commits/a0bc6ab714e0be6976cb358fb172040393d186d9

package.json
paddyo: Might be worth exposing the debugging port in the `docker-compose.yml` file too

QuarpT: https://github.com/guardian/gateway/pull/123/commits/0087f1e623c0297e7ac399b715f31060c4b661ad

coldlink:
Looks good to me.
I haven't checked it on `CODE` though, so I'll do that now before approving.

coldlink:
i think the error message needs a bit of padding/margin:
chrome_xvUVjIRcez

similar to how the message on the repeat password is:
chrome_uKYoR030Z3

src/client/components/ClientState.tsx
```diff
@@ -1,7 +1,9 @@
 import React, { createContext, FunctionComponent } from 'react';
 import { ClientState } from '@/shared/model/ClientState';
 
-export const ClientStateContext = createContext({});
+export const ClientStateContext = createContext({
+  clientHosts: { idapiBaseUrl: '' },
```
coldlink, line 5: if you've already added the defaults to server/models/Express.ts then i'm not sure you need to set the default here?

coldlink (approved):

src/client/components/ClientState.tsx
QuarpT: I have to set it because client hosts field is not optional
```
 Type '{}' is not assignable to type 'ClientState'.
```

PR #143, opened 2021-01-05
handle uncaught errors in gateway to avoid tcp connection leaks

QuarpT:
## What does this change?

- Create a helper function to call `next()` when the router async handler function fails. Synchronous errors are already handled by express 4.

- Return an error page using the default error handler and log the error.

## Why?

In Express 4, async handlers which fail to call `next()` (or specific functions on the response) leave the TCP connection open (indicating a leak) and don't return any data. 

## Notes

This error handling is provided by default in express 5 (which has been in alpha for years).

I'll add unit tests if developers approve this approach.

My initial plan was to wrap methods such as router.get. The approach in this PR is more simple to implement since the router http methods are very broadly typed.

## Images

image

### DIFF
--- a/src/server/lib/logger.ts
+++ b/src/server/lib/logger.ts
@@ -9,8 +9,21 @@ export const logger: Logger = {
   log(level: LogLevel, message: string) {
     winstonLogger.log(level, message);
   },
-  error(message: string) {
-    logger.log(LogLevel.ERROR, message);
+  // errors can be anything
+  // eslint-disable-next-line
+  error(message: string, error?: any) {
+    if (error && error.stack && typeof error.message === 'string') {
+      return logger.log(
+        LogLevel.ERROR,
+        `${message} ${error.message} ${error.stack}`,
+      );
+    }
+
+    if (error) {
+      return logger.log(LogLevel.ERROR, `${message} ${error}`);
+    }
+
+    return logger.log(LogLevel.ERROR, message);
   },
   warn(message: string) {
     logger.log(LogLevel.INFO, message);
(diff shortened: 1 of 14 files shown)


src/server/lib/logger.ts
```diff
@@ -9,8 +9,19 @@ export const logger: Logger = {
   log(level: LogLevel, message: string) {
     winstonLogger.log(level, message);
   },
-  error(message: string) {
-    logger.log(LogLevel.ERROR, message);
+  // errors can be anything
+  // eslint-disable-next-line
+  error(message: string, error?: any) {
+    if (error) {
+      if (error.stack && error.message) {
```
QuarpT, line 16: log stack traces if the error contains them

src/server/lib/logger.ts
```diff
@@ -9,8 +9,21 @@ export const logger: Logger = {
   log(level: LogLevel, message: string) {
     winstonLogger.log(level, message);
   },
-  error(message: string) {
-    logger.log(LogLevel.ERROR, message);
+  // errors can be anything
```
paddyo, line 12: I feel that this could be addressed, but not in this PR, might be worth something to think about as a team.

paddyo:
I think the approach looks sensible and offers little disruption, once that unit test is in place I think this is good to go.

src/server/lib/logger.ts
QuarpT: This PR needs a mechanism to log the stack traces if they exist. I'm happy to implement an alternative or have a chat about it.

paddyo: Oh this is out of scope of the PR, but I wanted to flag that you had to do this in order for this PR to work and that we can see, as a separate investigation if there's a nicer way we can deal with error shapes.

QuarpT: 👍  I misunderstood

paddyo (approved):

PR #144, opened 2021-01-06
fix flakey change password test

QuarpT:
## What was wrong

The breached password cypress test (which checks the long red error message appears when the password is breached and the user has selected the confirm password input) fails when the user inputs a breached password, then very quickly types in the repeated password field and unselects the field, before the breached promise returns.

## The fix
Now display the long red error message if either the user has selected the repeat password field, or the user has input some text into the repeat password field


coldlink (approved):
working for me today on local CI, so approving

PR #186, opened 2021-01-21
Fix invalid token redirect issue

coldlink:
## What does this change?
According to a [bug report](https://trello.com/c/kMyIHr2q/2707-bad-sign-in-redirect-when-token-invalidated-on-gateway-s), when a users identity token becomes invalid, Gateway redirects users to the wrong page. It redirects to the verify email page rather than the login page.

After doing some research, when an invalid token is submitted to IDAPI, it returns a JSON with the following format:
```js
{
  emailValidated: undefined,
  status: 'notSignedIn',
  redirect: {
    url: *URL OF LOGIN PAGE*
  }
}
```

From this we see that there is an `emailValidated` is `undefined` and `status` is `notSignedIn`. In the flow, Gateway would first run a check for `if (!auth.emailValidated) { ...`, this would mean that the above JSON would match, as `undefined` is a falsey value.

However  I noticed that we were not doing any checks for when the status was `notSignedIn` which should precede any other check for the response.

So I added a check in for ` if (auth.status === IDAPIAuthStatus.SIGNED_OUT) { ...` to catch this issue for when an invalid token was supplied, and redirect to the url if it was returned in the response or a default login url.

Since this redirect code was being used by the `if (auth.status === IDAPIAuthStatus.RECENT) {` check, I refactored this into a `redirectAuth` method so that it could be shared by both checks.

I then also added in an additional metric that would be fired if the signed out check passed.

### DIFF
--- a/src/server/lib/middleware/login.ts
+++ b/src/server/lib/middleware/login.ts
@@ -1,6 +1,6 @@
 import { NextFunction, Request, Response } from 'express';
 import { read } from '@/server/lib/idapi/auth';
-import { IDAPIAuthStatus } from '@/shared/model/IDAPIAuth';
+import { IDAPIAuthRedirect, IDAPIAuthStatus } from '@/shared/model/IDAPIAuth';
 import { getConfiguration } from '@/server/lib/getConfiguration';
 import { getProfileUrl } from '@/server/lib/getProfileUrl';
 import { Routes } from '@/shared/model/Routes';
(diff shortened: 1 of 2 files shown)


src/server/lib/middleware/login.ts
```diff
@@ -47,14 +61,7 @@ export const loginMiddleware = async (
       next();
     } else {
       trackMetric(Metrics.LOGIN_MIDDLEWARE_NOT_RECENT);
-      if (auth.redirect) {
```
alexduf, line 50: Is there a case here where the email is validated, there is a redirect URL, but the status isn't `SIGNED_OUT`? 

And if so, is it okay to fall into the `LOGIN_MIDDLEWARE_NOT_RECENT` case?

alexduf (approved):
I think it looks good but, putting my extra-paranoid hat on, is there a case that's been missed?

src/server/lib/middleware/login.ts
coldlink: There are only [3 statuses](https://github.com/guardian/gateway/blob/21763dbea01af34b1a6cdf5ceece32f716fc1c16/src/shared/model/IDAPIAuth.ts#L1):
- Signed Out
- Recently Signed In
- Not Recently Signed In

So all 3 cases have been covered.

PR #344, opened 2021-03-11
Initialise Google Analytics only if we have CMP consents

QuarpT:
## Changes

Initialise google analytics only if we have the google-analytics CMP consent

## Why

We need consents from the user before we can start collecting Google Analytics

### DIFF
--- a/src/client/static/index.tsx
+++ b/src/client/static/index.tsx
@@ -35,8 +35,6 @@ hydrateApp();
 // initalise ophan
 ophanInit();
 
-initGoogleAnalyticsWhenConsented();
-
 // don't load this if running in cypress
 if (!window.Cypress) {
   // load cmp if it should show
@@ -46,3 +44,5 @@ if (!window.Cypress) {
     cmp.init({ country });
   })();
 }
+
+initGoogleAnalyticsWhenConsented();
--- a/src/client/static/index.tsx
+++ b/src/client/static/index.tsx
@@ -1,7 +1,11 @@
 import { hydrateApp } from '@/client/static/hydration';
 
 // method to check if the cmp should show
-import { cmp } from '@guardian/consent-management-platform';
+import {
+  cmp,
+  getConsentFor,
+  onConsentChange,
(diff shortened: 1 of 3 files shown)


src/client/static/index.tsx
```diff
@@ -15,13 +19,23 @@ import { init as ophanInit } from './analytics/ophan';
...
+    ) {
+      gaInit();
+    }
+  });
+};
+
 hydrateApp();
 
 // initalise ophan
 ophanInit();
```
coldlink, line 36: i know the ticket was only for ga, but do we want to consider something similar for ophan?

approving regardless

coldlink (approved):

src/client/static/index.tsx
mchv: @coldlink We don't treat the ophan tracker and GA tracker in the same way for few different reasons:
-   `Ophan` is a `first-party` tracker while GA is `third-party` tracker. We can't send personal data to any third-party without disclosure, an updated privacy policy and a technical integration that protect the privacy of reader data _by design_.
- While we may intend to use `GA` with legitimate interest as lawful basis for processing, we found it too difficult to correctly configure `GA` to ensure that the data collected will be used only for that purpose. As such we only send data to GA for reader who consented for all purposes.
- For `Ophan` tracker, we have architected our data warehouse to ensure that data collected for legitimate interest purpose will never be integrated and joined with data collected for other purposes and  have put the right control in place.

We may use in the future different trackers to make this more seamless, but in the meantime the ophan tracker should not be removed as the data is collected and processed with legitimate interest as a lawful basis for processing.

mchv: @paulbrown1982 @dominickendrick are the right people to talk too about the use cases for the data collected by the ophan tracker under legitimate interest which is likely to include some customer journey analysis and optimisation.

coldlink: Very interesting to learn. Thanks @mchv 👍

PR #369, opened 2021-03-18
AB Test: Follow on consent

paddyo:
## Description

AB Test: followupConsent. This AB test shows an opportunity to subscribe to a newsletter (Variant A) or supporter emails (Variant B). Users are redirected here from `identity-frontend`. If a user is already subscribed they will be immediately redirected back to the Guardian/Return Url.

Both variants work in IE11.

## Changes
* Add the `ConsentFollowUp` page to handle both variants.
* Add post/get routes, `/consents/follow-up` & `/consents/follow-on` for both variants, as well as private utility functions for them.
* Modified the `EnvelopeImage` to allow an inverted colour scheme.
* Add `redirectRenderer` for rendering a page whose purpose is simply to redirect. This is to deal with the very poor consistency of CSP form submit implementations across browsers, most notably Chrome.

## Screenshots

Note copy in the screenshots outdated.

### Variant One
<p align="center">

</p>

### Variant Two
<p align="center">

</p>


src/server/routes/consents.ts
```diff
@@ -261,6 +269,226 @@ router.get(Routes.CONSENTS, loginMiddleware, (_: Request, res: Response) => {
...
+  return [IDAPIEntity.CONSENTS, consents];
+}
+
+async function setNewsletterEntity(
+  ip: string,
+  sc_gu_u: string,
+  body: { [key: string]: string },
+  geocode?: GeoLocation,
+) {
+  const { update } = consentPages[1]; // Can reuse newsletter updater function.
```
alexduf, line 368: The access with a hardcoded index makes me feel uncomfortable. 

It might be worth pulling the definition of each page out of the array, then compose an array with these definitions. That way you can directly access the definition you fancy without having to care about the order of each page in the array.

src/server/routes/consents.ts
```diff
@@ -261,6 +269,226 @@ router.get(Routes.CONSENTS, loginMiddleware, (_: Request, res: Response) => {
...
+      [status, state] = getErrorResponse(e, state);
+    }
+
+    const html = renderer(route, {
+      pageTitle,
+      requestState: state,
+    });
+    res
+      .type('html')
+      .status(status ?? 500)
```
alexduf, line 453: Why the `??` ?, can status be null-ish here?

alexduf (commented):
Only minor remarks or questions from me

src/client/components/EnvelopeImage.tsx
```diff
@@ -3,51 +3,73 @@ import React from 'react';
 
 interface Props {
   width?: string;
-  cssOverrides?: SerializedStyles;
+  cssOverrides?: SerializedStyles | SerializedStyles[];
```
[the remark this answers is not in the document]
paddyo: Yes, `cssOverrides` is the name that Source uses to allow Emotion styles passed in. We follow that convention too so that Source would find it easier to extract our components and it should be easier to understand what's going on.

src/server/routes/consents.ts
paddyo: It can't be, and that's a problem ... it shouldn't default to a success! Same with the other handler, I will fix both.

src/client/pages/ConsentsFollowUp.tsx
```diff
@@ -0,0 +1,395 @@
...
+  ${from.desktop} {
+    -ms-grid-row: 1;
+    grid-row: 1;
+  }
+`;
+
+function getVariantValue<A, B, C>(
+  A: A,
+  B: B,
+  none: C,
```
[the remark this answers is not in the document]
paddyo: Agreed, will fix.

src/server/routes/consents.ts
paddyo: As discussed, I agree, but don't have the capacity to handle this refactor. It would be great to see this done in future though.

src/client/components/EnvelopeImage.tsx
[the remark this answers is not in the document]
alexduf: Great, I didn't particularly like how it was both an array and an object but if we have no control over it, so be it 🤷

alexduf (approved):

PR #384, opened 2021-03-24
Bundle/browser optimisations

coldlink:
## What does this change?
- Update webpack configuration to output two client bundles, one for browsers with `<script type="module" src="..." ></script>` support, and a legacy bundle for browsers without module support (target IE11)
  - This will mean that we can provide a smaller bundle size to modern browsers as we don't have to rely on polyfills/transforms etc. and only include these in browsers that need them (legacy)
  - The renderer method in the server will include both the modern and legacy bundles in the html, in modern browsers anything with `nomodule` attribute is ignored, and in legacy browsers `type="module"` scripts are not executed.
  - Update the documentation to reflect this
- Set a performance budget/size for the bundles
  - `modern`
    - assets: `384KB`
    - entry: `512KB`
  - `legacy`
    - assets: `512KB`
    - entry: `768KB`  
- Webpack optimisations mean that we can remove `unsafe-eval` in the CSP **(big yay)**.
- Update `makefile` to use `.PHONY`
- Remove documentation on dynamic imports, as this doesn't work with the `SplitChunksPlugin` in webpack
  - We can investigate dynamic imports in the future if needed especially with [Loadable Components](https://loadable-components.com/), as we can then only provide the needed js for each component, rather than all the js. But this will eat into developer time
- Update dependencies
- Add documentation to expect errors on the first launch of the application, this is due to `webpack --watch` and `nodemon` running in parallel, but `nodemon` complains it cannot find the right files as webpack hasn't finished building them.
  - Since neither applications exit until you stop them, it's not possible to serially run one after the other

## Bundle changes
Current:
```
Entrypoint main [big] 549 KiB = c1002448fe0bcdddcad8.bundle.js 1.56 KiB 06f2d1246df616da9a37.bundle.js 481 KiB 159006e35acc7c54e7f3.bundle.js 66.1 KiB 8 auxiliary assets
```

This PR
Legacy
```
Entrypoint main 573 KiB (417 KiB) = runtime.legacy.9703a449539a0467a9bd.js 1.56 KiB vendors.legacy.1fca4ccdce44cf780bed.js 500 KiB main.legacy.96b9d5ae5309da17dda3.js 72.1 KiB 9 auxiliary assets
```

Modern
```
Entrypoint main 315 KiB (417 KiB) = runtime.92126914a3f2d4b99fd2.js 1.18 KiB vendors.e0ddde461e9c393a7e7f.js 248 KiB main.8ea0ed2ee9048d51dd51.js 65.3 KiB 9 auxiliary assets
```


rebecca-thompson (approved):
From what I can see this looks fine, but maybe get a more experienced client-side dev to take a look?

PR #491, opened 2021-05-04
NotFound story

oliverlloyd:
## What?
Adds a story for the `NotFound` page

Screenshot 2021-05-04 at 17 33 47

I also added configuration for displaying stories at different breakpoints

## Why?
So that we have a stepping stone for building out more test coverage and understanding for what the application does. Adding these stories helps me learn the codebase by seeing what each page and component does and how it looks when you change the props. Hopefully, as well as giving us regression testing, this visual documentation will be useful for others going forward


coldlink (approved):
LGTM

wrt to breakpoints for the consents/onboarding flow we use a two layouts:

ConsentsLayout: https://github.com/guardian/gateway/blob/b5badc928f804a4909114d6fd572dd5147f75d34/src/client/layouts/ConsentsLayout.tsx
Used in the onboarding/consents flow, which uses the following grid system based on the design system:
https://github.com/guardian/gateway/blob/b5badc928f804a4909114d6fd572dd5147f75d34/src/client/styles/Grid.ts

Then there is also the SignInLayout, currently used by the reset password flow:
https://github.com/guardian/gateway/blob/2d0ff1d8309bd478e221bbd192e119b0d47dc0ec/src/client/layouts/SignInLayout.tsx
This is based on flexbox, and has a maxwidth of tablet (740px)

We have a list of breakpoints defined: https://github.com/guardian/gateway/blob/6303b41a2db6b4799e261fadafb9185b412b3cc5/src/client/models/Style.ts

oliverlloyd:
Breakpoint logic now refactored to lean on the local breakpoint definitions 👍

PR #494, opened 2021-05-05
5th May 2021 | Updating dependencies

coldlink:
## What does this change?
- Change dependabot to only run weekly to avoid PR spam

## Tested?
- [x] Local
- [x] CI
- [ ] CODE


.github/dependabot.yml
```diff
@@ -3,7 +3,7 @@ updates:
   - package-ecosystem: "npm"
     directory: "/"
     schedule:
-      interval: "daily"
-    open-pull-requests-limit: 10
+      interval: "weekly"
+    open-pull-requests-limit: 25
```
oliverlloyd, line 7: Bring it on!

package.json
```diff
@@ -102,18 +102,18 @@
     "@guardian/src-link": "^3.5.0",
     "@guardian/src-radio": "^3.5.0",
     "@guardian/src-text-input": "^3.5.0",
-    "aws-sdk": "2.893.0",
+    "aws-sdk": "2.900.0",
```
oliverlloyd, line 105: I presume this won't trigger the TC issue we had the other day because now we're specifying it as a dep?

oliverlloyd (approved):

package.json
coldlink: Yeah, still not entirely sure what the issue was, but now that we specify `@guardian/node-riffraff-artifact` as a dependency the build passes in TeamCity.

PR #495, opened 2021-05-05
`MaxWidth` > `Breakpoints`

oliverlloyd:
## What?
This PR removes `MinWidth` and renames `MaxWidth` to `Breakpoints`.

```ts
export enum Breakpoints {
  WIDE = 1300,
  DESKTOP = 980,
  TABLET = 740,
  MOBILE = 660,
  MOBILE_LANDSCAPE = 480,
}
```

## Why?
I removed `MinWidth` because in the one place it was used it was being applied to `max-width`. That left `MaxWidth` which made sense and was fine but I thought it might be more scalable / flexible if it were exported as `Breakpoints`. Having a property named like this means it is not so tied to the css property name and is useful as a declaration that these are the breakpoints we work to

### DIFF
--- a/.storybook/preview.js
+++ b/.storybook/preview.js
@@ -1,14 +1,14 @@
 import { breakpoints } from '@guardian/src-foundations/mq';
-import { MaxWidth } from '@/client/models/Style';
+import { Breakpoints } from '@/client/models/Style';
 
 const viewports = {};
-for (let breakpoint in MaxWidth) {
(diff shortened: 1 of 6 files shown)


oliverlloyd:
[Breakpoints](https://trello.com/c/zM8Tkh3H/2961-breakpoints)

coldlink (approved):

PR #496, opened 2021-05-05
Mock ABProvider and StaticRouter in Storybook

oliverlloyd:
## What?
Uses the `preview.js` file in storybook to add global mocks for `ABProvider` and `StaticRouter`

See: https://storybook.js.org/docs/react/writing-stories/decorators#context-for-mocking

## Is this the right approach?
Typically when an app depends on global state such as Context the best solution is to isolate the component and only render the presentational elements. Then, to add stateful logic (things like `useContext` or `useAB`) you use a separate component and pass down state as props. This is an option here (and maybe we could take this approach?) but we can also use mocking as in this PR.

I'm a little on the fence which is the best approach. By creating a separation between presentation (via props) and Context we make testing a lot easier. This will arguably also make scaling things in the future easier too. But it adds additional components/layers and would mean we'd need to do some refactoring.

### DIFF
--- a/.storybook/preview.js
+++ b/.storybook/preview.js
@@ -1,5 +1,7 @@
 import { breakpoints } from '@guardian/src-foundations/mq';
 import { Breakpoints } from '@/client/models/Style';
+import { ABProvider } from '@guardian/ab-react';
+import { StaticRouter } from 'react-router-dom';
 
 const viewports = {};
 for (let breakpoint in Breakpoints) {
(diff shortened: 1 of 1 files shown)


coldlink:
I see both sides to this argument.

I'm _**slightly**_ leaning towards the refactor though, especially if there are changes to the structure of specifically the `ClientState` which would involve updating many mocks as it's widely used throughout the application. Also in some way's we already have additional layers through our use of the layout components for specific pages.

However like you say, I see that the work to refactor everything would take some time, and there would be quite a few additional component layers. Though if we can spend the time now before we've fully steamed ahead on the sign in migration it would make things easier going forward.

oliverlloyd:
> However like you say, I see that the work to refactor everything would take some time and may be tricky to do, and there would be quite a few additional component layers. Though if we can spend the time now before we've fully steamed ahead on the sign in migration it would make things easier going forward.

I think this fairly succintly sums it up. It might take longer to do the refactor but maybe now is the time to put the effort in.

I really like the idea of investing extra effort at the beginning of projects, trying different appraoches, being willing to refactor, and focussing on dev ex, etc. In all likelyhood, this code will live for a long time and we want our future selves to be happy.

Let me try a little spike to see how the separation might go and if it's not too bad it might turn out to be a nice way for me to learn all the components!

oliverlloyd:
> Let me try a little spike to see how the separation might go and if it's not too bad it might turn out to be a nice way for me to learn all the components!

Wasn't hard at all. If we're happy with https://github.com/guardian/gateway/pull/500 then I can close this PR and extend this logic separation approach instead

oliverlloyd:
Closed based on discussions

PR #497, opened 2021-05-05
`ConsentConfirmation` Stories

oliverlloyd:
## What?
Adds stories for the `ConsentConfirmation` page.

** Draft pending #496 **

### `ClientStateProvider`
I'm using this component to set the Context so that we can render the component in different states.

I _think_ I've got all the possible states covered but I had to read through the code to work out the various possible presentation states so I might have missed some!

## Why?
Having a visual snapshot is useful not just for if we want to do more development going forward but it also enables us to snapshot this page for visual regression in the future and also provided a visual document for the things Gateway does.


oliverlloyd:
Closed in favour of #501

PR #500, opened 2021-05-05
Refactor `ResetPassword`

oliverlloyd:
## What?
This refactor is born from the discussion [over here](https://github.com/guardian/gateway/pull/496#issuecomment-832675262) where we talked about seprating the state logic out from what gets rendered.

To achieve this it's was actually pretty simple. I just dropped the `SignInLayout` component down one level, which meant we had one 'Page' component which dealt with state and another 'Presentation' component to handle rendering.

I then did a bunch of file renames to account for this. The files are now called:

`ResetPasswordPage` - this is called by Router and is responsible for state
`ResetPassword` - this is purely presentational and just takes props and show them

## Why?
To make testing easier basically. We no no longer need to setup any mocks and don't have any magical global state to consider when testing.

## Don't be dogmatic
There used to be a dogma around that said you must _always_ have this presentation/state separation and to not have it was wrong. I don't hold with that and think it's better to just do whatever feels right in each situation. I think here there's a good case to have the separation but maybe that changes in future and we move away from it - that will also be fine.


oliverlloyd:
[ResetPassword Stories](https://trello.com/c/8g5swuia/2962-resetpassword-stories)

coldlink (approved):
Really liking the look of this approach. The rest of the reset password flow should be straightforward to refactor, but the onboarding/consents flow might prove more difficult.

PR #511, opened 2021-05-06
Refactor `ConsentsData`

oliverlloyd:
## What?
Breaks out the presentational aspects of `ConsentsData` and adds `ConsentsDataPage` as the container for state.

## Why?
Allows stories to be easily created (included)

## What's with no description?
There's an edge case where the state passed in _could_ contain an empty consent. The previous code we refactored from checked for this state and displayed nothing in that case. We're doing the same thing here now after initially not worrying about it and discovering that that caused Cypress to complain

We might want to think through if we need to handle this edge case better. It seems to only happen if there's an error from an api call but I presume this is a possibility?


coldlink (approved):
Looks good. I wonder if it's better to combine a few of these refactor PRs going forward. Happy to review either way

PR #512, opened 2021-05-07
Eslint/Prettier/Codeowner Changes

coldlink:
## What does this change?
- Disable `functional/immutable-data` for `*.stories.tsx` in eslint so that we don't need to include `/* eslint-disable functional/immutable-data */` at the top of every file
- Add `build` folder to `.prettierignore` so that it isn't checked by prettier when running the husky git hook
- Add identity team as CODEOWNERS

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -31,6 +31,14 @@
     "no-param-reassign": "error",
     "prefer-const": "error"
   },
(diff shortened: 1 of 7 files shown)


oliverlloyd (approved):
Nice!

PR #515, opened 2021-05-07
Fix: Unsubscribe issue on newsletter forms

coldlink:
## What does this change?
The newsletters onboarding page (during the single newsletter test) as well as the (now switched off test) follow-up newsletter test could potentially cause a user to unsubscribe from certain newsletters that they were already subscribed to.

This was caused by the newsletter subscriptions update method expecting four specific newsletters to be included in the body to update, whereas both those tests would only include one newsletter at most.

This meant that if a user was previously subscribed to the other 3 newsletters not shown the method would unsubscribe them, as they were not defined in the HTML form body.

To fix this issue the first change was to change the `newslettersSubscriptionFromPage` method to accept a generic range of newsletters, rather than the 4 it was expecting on the page. This was done by abstracting into the `newslettersSubscriptionsFromFormBody` method. This method compared all the possible valid newsletter ids to those returned in the form body. If a newsletter id existed in the form body it would mark it as to possibly subscribe or unsubscribe, otherwise we ignore it if it wasn't in the morning. This lets us update an arbitrary range of newsletters at once.

The other change was for each newsletter checkbox to add in a `hidden` input element with the same name as the newsletter checkbox, but with an empty value. This way when the form is submitted, the form body would include the key with an empty string value if the checkbox was unchecked. This meant that the `newslettersSubscriptionsFromFormBody` method could see if the checkbox was unchecked or checked. It would be unchecked if the form body had an empty value for that newsletter and checked if it didn't have an empty value. See the [tests](https://github.com/guardian/gateway/blob/fa49d89f4211d76c961ee8fe01c00f35e59e2150/src/shared/__tests__/newsletter.test.ts) for the `newslettersSubscriptionsFromFormBody` to see what the body looks like.

As was previously the case, after the form body is parsed using `newslettersSubscriptionsFromFormBody` we check this against the current users subscriptions (`userNewsletterSubscriptions`), if there is a difference then a user has subscribed or unsubscribed from a newsletter (`newsletterSubscriptionsToUpdate`).

## Other changes
- Add a list of all newsletter ids (`ALL_NEWSLETTER_IDS`)
- Move the newsletter images out of the `src/shared/model/Newsletter.ts` to `src/client/model/Newsletter.ts`, as the `shared` directory should not inherit anything from the `client` or `server` directory (this was causing the new jest tests to fail, as jest didn't know what to do with the `.jpg`s)

[Trello Card](https://trello.com/c/irygPfZy/2935-gateway-newsletter-consent-implementation-can-unsubscribe-users-from-newsletters-s)


src/shared/lib/newsletter.ts
```diff
@@ -50,3 +54,21 @@ export const NewsletterMap = new Map<GeoLocation | undefined, Newsletters[]>([
...
+    // then mark this newsletter as to potentially update (subscribe / unsubscribe)
+    // otherwise return undefined
+    if (id in body) {
+      return {
+        id,
+        subscribed: !!body[id],
+      };
+    }
+    // filter out the undefined values from the array to return the correct type
+  }).filter(Boolean) as NewsletterPatch[];
```
oliverlloyd, line 74: You could use a simple forEach here and push to a separate array if id in body? Horses for courses and not a big deal or a blocker but it would be a way to avoid having to cast the type / filter out undefined?

I know the pattern of defining a new array and using forEach to populate it gets a bad rep but sometimes it does the job 🤷

oliverlloyd (approved):
In my experience this kind of state syncing, where you need to know when to add or remove things, is surprisingly hard! I feel like you've got Cache invalidation, Naming things, and then things like this (soft/hard deletes basically)

But this code looks great. I made a few non blocking comments but mainly because this is my first time seeing the code, nothing major

src/server/routes/consents.ts
```diff
@@ -168,44 +171,40 @@ export const consentPages: ConsentPage[] = [
         throw error;
       }
     },
-    update: async (ip, sc_gu_u, body, geo) => {
-      const newslettersPage = NewsletterMap.get(geo) as string[];
-
+    update: async (ip, sc_gu_u, body) => {
```
QuarpT, line 174: can we update the caller to stop passing in `geo`?

src/shared/lib/newsletter.ts
QuarpT: flatMap might be an option https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/flatMap#for_adding_and_removing_items_during_a_map

QuarpT (approved):

src/server/routes/consents.ts
```diff
@@ -168,44 +171,40 @@ export const consentPages: ConsentPage[] = [
...
+            if (subscription) {
+              // if previously subscribed AND now wants to unsubscribe
+              // OR if previously not subscribed AND wants to subscribe
+              // then include in newsletterSubscriptionsToUpdate
+              if (
+                (subscription.subscribed && !subscribed) ||
+                (!subscription.subscribed && subscribed)
+              ) {
+                return true;
+              }
```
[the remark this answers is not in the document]
coldlink: i've updated the variable names and moved the comments a little to make it easier to follow and read

src/shared/lib/newsletter.ts
coldlink: updated to use `flatMap`, works perfectly 😁

src/server/routes/consents.ts
coldlink: 👍 removed the `geo` in all the callers for `update`

PR #525, opened 2021-05-11
Complete Page Refactors

oliverlloyd:
## What does this change?
Refactors the following components to split out page level state with the local presentation logic

`ChangePassword`
`ConsentsFollowUp`
`ConsentsNewsletters`
`ConsentsCommunication`

This completes the work to document all the Gateway pages in Storybook

## Why?
As before, the goal is to create a presentational component that is easy to load into Storybook (without the need for mocking context). This is useful for testing but also improves the readability and documentation for each page as we now see their dependecies written out (and typed) as props.

### `ConsentsFollowUp` looks off
It does, this component is accepting arrays but only rendering a single item. If you pass it multiple items only the last one is rendered. After discussing this with Ana, I've got a separate PR ready to refactor this component to be more semantic


src/client/pages/ConsentsCommunication.stories.tsx
```diff
@@ -0,0 +1,123 @@
...
+
+export const MultipleConsentsAndMarketing = () => (
+  <ConsentsCommunication
+    isUserInTest={false}
+    marketResearchOptout={{
+      id: '0',
+      name: 'mr-name',
+      description:
+        'I do NOT wish to be contacted by The Guardian for market research purposes.',
+    }}
```
coldlink, line 79: this consent is not showing up in the story

coldlink: when the user is not in test `isUserInTest = false`, this consent should show

src/client/pages/ConsentsCommunication.stories.tsx
```diff
@@ -0,0 +1,123 @@
...
+      {
+        id: '1',
+        name: 'My Other Consent Name',
+        description: 'Other consent description',
+      },
+    ]}
+  />
+);
+MultipleConsentsAndMarketing.story = {
+  name: 'with multiple consents',
```
coldlink, line 95: this should be 
```suggestion
  name: 'with multiple consents and marketing',
```

src/client/pages/ConsentsFollowUp.stories.tsx
```diff
@@ -0,0 +1,117 @@
...
+        description: 'Another newsletter description',
+        name: 'Another Newsletter Name',
+      },
+    ]}
+    consents={[]}
+  />
+);
+MultipleNewsLetters.story = {
+  name: 'with multiple newsletters',
+};
```
coldlink, line 57: ah this is why in standup you mentioned why this was an array, yeah i guess we should refactor this to only take a single newsletter or use one of the particular newsletters in this array

src/client/pages/ConsentsFollowUp.stories.tsx
```diff
@@ -0,0 +1,117 @@
...
+        name: 'Newsletter Name',
+      },
+    ]}
+    consents={[
+      {
+        id: '0',
+        description: 'Description for this consent',
+        name: 'Consent name',
+      },
+    ]}
```
coldlink, line 112: almost feels like we should combine these two into a single prop, but this test isn't on right now, so we can discuss this later 🤔

src/client/pages/ConsentsNewsletters.stories.tsx
```diff
@@ -0,0 +1,104 @@
...
+};
+
+export const SingleNewsletter = () => (
+  <ConsentsNewsletters
+    newsletters={[
+      {
+        id: '0',
+        nameId: 'n0',
+        description: 'Newsletter description',
+        name: 'Newsletter Name',
```
coldlink, line 33: for one of these stories, it might be worth using an id of a newsletter that has an image, just so we can see what it's like with one e.g. using one of the id's from https://github.com/guardian/gateway/blob/e3b3f2d6d66efe04d03195ac30030a3b1edaa7bb/src/shared/model/Newsletter.ts#L24

src/client/pages/ConsentsNewsletters.stories.tsx
```diff
@@ -0,0 +1,104 @@
...
+};
+
+export const SingleNewsletterInTest = () => (
+  <ConsentsNewsletters
+    newsletters={[
+      {
+        id: '0',
+        nameId: 'n0',
+        description: 'Newsletter description',
+        name: 'Newsletter Name',
```
coldlink, line 50: again as in the previous comment, and probably more so for "in test", would be useful to get a newsletter with an image

src/client/pages/ConsentsFollowUp.stories.tsx
oliverlloyd: Ha, yep. I already have this PR queued up and it's doing exactly this 👍

src/client/pages/ConsentsCommunication.stories.tsx
coldlink: ah interesting, it's showing up now, i have a feeling that it was due to the name clashing with the other story before

coldlink (approved):

PR #528, opened 2021-05-12
Base css for Storybook

oliverlloyd:
## What does this change?
Storybok was missing some base reset css so I added it

### Before
Screenshot 2021-05-12 at 09 26 49

### After
Screenshot 2021-05-12 at 09 27 36

### DIFF
--- a/.storybook/preview-head.html
+++ b/.storybook/preview-head.html
@@ -1,0 +1,17 @@
+<style>
+  html {
+    height: 100%;
+  }
+  body {
+    height: 100%;
+  }
+  #app {
+    min-height: 100%;
+    height: 100%;
+    display: flex;
+    flex-direction: column;
+  }
+  * {
+    box-sizing: border-box;
+  }
+</style>


coldlink (approved):

coldlink (approved):

coldlink:
we have some reset code here: https://github.com/guardian/gateway/blob/3da8f3d0d2307a7b275f7ae620cd3bd57c592c8f/src/client/main.tsx#L29

is this where you got it from?

oliverlloyd:
> we have some reset code here:
> 
> https://github.com/guardian/gateway/blob/3da8f3d0d2307a7b275f7ae620cd3bd57c592c8f/src/client/main.tsx#L29
> 
> is this where you got it from?

Yup. But those globals aren't applied in SB because that file is never executed.

Note. I did try using that same `Global` component to apply them via a decorator but it didn't work and rather than put a foot in a potential rabbit hole I just went with the simple solution

PR #532, opened 2021-05-13
Refactor layout

oliverlloyd:
## What does this change?
This PR refactors the layout, header and sub header elements to:

1. Call them `Layout`, `Header` and `SubHeader`
2. Change their sematics of our layout html tags
3. Make the title text a prop that can be passed through

## Why?
1. We want too align naming with the wider discussion around these components
2. It's valid having two `header` tags on a page and `header` better represents the content than `nav` but these elements [can't be siblings](https://dequeuniversity.com/rules/axe/4.1/landmark-no-duplicate-banner) so this PR refactors from

```ts
<header/>
<nav>SubHeader</nav>
<main>
  {children}
</main>
<footer/>
```

to

```ts
<header/>
<main>
  <header>SubHeader</header>
  <section>{children}</section>
</main>
<footer/>
```

3. We don't want to always use 'Sign in' as the text, sometimes we want more context sensitive text shown, such as 'Registration'


oliverlloyd:
[Refactor Layouts](https://trello.com/c/qflXlIcZ/2985-refactor-layouts)

coldlink:
profile thegulocal com_reset(Laptop with HiDPI screen)

seems like the footer isn't being pushed to the bottom of the page anymore 🤔

coldlink:
profile thegulocal com_consents_data_returnUrl=https%3A%2F%2Fm code dev-theguardian com%2F

also affecting the ConsentsLayout too, but only on super high resolution screens, as there's usually enough content to keep the footer at the bottom

oliverlloyd:
> seems like the footer isn't being pushed to the bottom of the page anymore

Great spot!

coldlink:
i wonder if this is a case for using specific heights for breakpoints in storybook? to make it easier to spot issues like this

oliverlloyd:
> also affecting the ConsentsLayout too

Hmm, I'm also seeing the same on `main`

Screenshot 2021-05-14 at 09 41 05

coldlink:
> > also affecting the ConsentsLayout too
> 
> Hmm, I'm also seeing the same on `main`
> 
> Screenshot 2021-05-14 at 11 40 44

oh interesting, i checked the `Layout` (`SignInLayout`) on `main` and that's definitely affected

i'll dig up the designs for the `ConsentsLayout` and see what's happening there

oliverlloyd:
@coldlink I've pushed a couple of commits to fix both `Layout` and `ConsentsLayout` to ensure they stretch vertically.

I also made a small change to the Storybook config to make sure we're rendering pages with the correct global css (we were missing `height: 100%` on `body` and `html`)

coldlink (approved):

PR #550, opened 2021-05-18
Use guardian/libs getLocale method for getting country code

coldlink:
## What does this change?
The commercial team sent around an email saying that `gu.geolocation` in `localStorage` in favour of the [`getLocale`](https://github.com/guardian/libs/blob/HEAD/src/getLocale.README.md) method provided by the [guardian/libs](https://github.com/guardian/libs) library

This PR removes the old geolocation code in favour of the new method.

Context: [Trello](https://trello.com/c/TIEGnmsw/2966-use-gugeocountry-cookie-in-gateway-idfe)


QuarpT (approved):

oliverlloyd (approved):

src/client/static/index.tsx
```diff
@@ -39,9 +37,11 @@ ophanInit();
 if (!window.Cypress) {
   // load cmp if it should show
   (async () => {
-    const country = await getCountryCode();
+    const country = await getLocale();
 
-    cmp.init({ country });
+    if (country) {
+      cmp.init({ country });
+    }
```
oliverlloyd, line 44: What's the ux here if the call to getLocale fails? Do users then not get to set their consent prefs?

Just trying to work out if there's a default needed or if cmp really needs to know the country

oliverlloyd: Note. This is a bit outside this PR, just curious really

coldlink: the cmp throws an error if the country is not supplied according to: https://github.com/guardian/consent-management-platform#optionscountry

i don't think we can set a default country/value

coldlink: yeah the users don't get to set their consent perfs, in turn meaning that we have to assume they havent consented to anything

coldlink: the only thing in gateway that requires consent is google analytics

oliverlloyd: > we have to assume they havent consented to anything

Ah, this seems fair. I guess that's the best outcome 👍

PR #575, opened 2021-05-24
`Nav`

oliverlloyd:
## What does this change?
Adds the `Nav` component.

The active bar state shows on focus:

The active bar state shows on hover

Styles change above and below tablet

## This is the same as Pillar nav!
Yes, it is. The main pillar section nav control is very similar to this component. In fact I used a lot of the styling from that code here. Could we eventually centralise this logic? Maybe, this is not something we need to action straight away but should keep this possibility in mind as we learn this use case.

## We're now using `Container`
`Container` is a [layout component from Source](https://theguardian.design/2a1e5182b/p/440a83-container/b/2402e9) that centralises the logic around how to position things on the page. There's still [an open PR ](https://github.com/guardian/source/pull/820)around enhancing this to better serve our needs but for now I'm using css overrides as workarounds

## No local state
This PR has no local state included (for which tab is active). We could refactor this in later or use a parent component. I'm leaving this choice out of this PR for now because I'm not certain how we will do routing.

### DIFF
--- a/src/client/components/Nav.stories.tsx
+++ b/src/client/components/Nav.stories.tsx
@@ -1,0 +1,105 @@
+import React from 'react';
+import { Meta } from '@storybook/react';
+
+import { Header } from './Header';
+import { Nav } from './Nav';
+
+export default {
+  title: 'Components/Nav',
+  component: Nav,
+  parameters: { layout: 'fullscreen' },
+} as Meta;
+
+export const Register = () => (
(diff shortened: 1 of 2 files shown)


oliverlloyd:
[Nav Tabs](https://trello.com/c/Zqcd0HZd/3001-nav-tabs)

QuarpT (approved):

PR #576, opened 2021-05-25
Refactor `Footer`

oliverlloyd:
## What does this change?
Refactors `Footer` to add support for the new designs and to use more components from `Source`

## What's the deal with all these side padding overrides?
The [Container component in source](https://github.com/guardian/source/tree/main/src/core/components/layout/components/container) is pending [an update](https://github.com/guardian/source/pull/820) to allow us more control when dealing with side padding. Once this PR is merged and published we can import the new features and use those instead.

I've noticed this override issue is quite common so perhaps a good interim solution here would be to abstract them by creating our own version of `Container` with the expected changes built in?


oliverlloyd:
[Footer](https://trello.com/c/E4vM8aTD/3002-footer)

src/client/components/Footer.tsx
```diff
@@ -1,87 +1,178 @@
...
-        &copy; {new Date().getFullYear()} Guardian News &amp; Media Limited or
-        its affiliated companies. All rights reserved.
-      </p>
-      <ul css={ul}>
-        {link('Help', locations.HELP)}
-        {link('Terms & Conditions', locations.TERMS)}
-        {link('Contact us', locations.CONTACT_US)}
-        {link('Privacy policy', locations.PRIVACY)}
-        {link('Report technical issue', locations.REPORT_ISSUE)}
-        {link('Cookie Policy', locations.COOKIE_POLICY)}
```
coldlink, line 83: Do we not need these two links anymore?

coldlink (approved):
Approving, but one small question

src/client/components/Footer.tsx
oliverlloyd: > Do we not need these two links anymore?

They're not in the designs so I assume not?
Screenshot 2021-05-25 at 15 41 41

PR #577, opened 2021-05-26
17th May 2021 | Updating dependencies 

coldlink:
## What does this change?

## Nodes
- `jest` and related deps are major version updates, however breaking changes didn't affect this project
- `@guardian/libs` is also a major version update, again the breaking change didn't affect this project

## Tested?
- [x] Local
- [x] CI
- [ ] CODE


rebecca-thompson (approved):

oliverlloyd (approved):
Thanks for the comments on jest and libs, I was worried about the major bumps but then I saw you'd already checked it 👍

PR #594, opened 2021-06-01
Storybook coverage

oliverlloyd:
## What does this change?
Completes storybook coverage for the remaining components in our library

### Refactors along the way
In some cases to be able to add the story I had to refactor some things. I also did some light renaming to make the story files makes sense (eg. Keep the name of the story the same as the file and the same as the component)

The individual refactors are hopefully well describes by each commit. The biggest area of change was around `Consents` which exported a lot of shared code used in different places. Some of this is now inlined to help with isolation and testing.

### Why is this snapshot file removed?
I replaced the jest snapshot file with some stories that, along with Chromatic, do the same thing but better because we're not snapshotting the visual appearance of the outcome of the css and html, not the css and html itself


coldlink (approved):
LGTM, I double checked the routes on the `CODE` environment too, and they look fine. 😃

PR #595, opened 2021-06-03
`returnUrl`

oliverlloyd:
## What does this change?
Moves the place where we set `pageData.returnUrl` from in the local routes to `requestState.ts`, where the state object is first built based on `req: Request`

## Why?
Because this means we will always pass through `returnUrl` into `pageData` if it is set as a query param, this means we don't need to keep adding code to different routes when we want to make `returnUrl` available


oliverlloyd:
[returnUrl](https://trello.com/c/Mi2ViZSI/3039-returnurl)

coldlink (approved):
Yeah, approving this for now, I think this is okay, especially because there are many pages that require the `returnUrl` across the app,

It might be worth refactoring this file to use the `pageData` from the `ClientStateContext` instead of the custom `useQuery` hook to get the return url. https://github.com/guardian/gateway/blob/main/src/client/pages/ChangePasswordCompletePage.tsx

PR #596, opened 2021-06-03
Sign in routing

oliverlloyd:
## What does this change?
Adds the routes and events for authenticating a reader using their email and password on the Gateway `SignIn` page

## This uses `idapi` not `okta`
This PR is using the existing `/auth` endpoint provided by `idapi`. We could - potentially - take the approach where we release this code, working against `idapi` and then later make the change to use `okta`. This has the advantage of breaking the changes up into stages. We can first go live with the frontend changes and then later push out the change to swap out the backend call by itself.

### Still pending
Social sign in
Testing on `CODE`


oliverlloyd:
[SignIn routing](https://trello.com/c/29u15rpn/3020-signin-routing)

src/client/components/Nav.tsx
```diff
@@ -153,7 +154,7 @@ export const Nav = ({ tabs }: Props) => (
       border={true}
       cssOverrides={[borderOverrides, paddingLeftOverrides]}
     >
-      <div css={tabRowStyles}>
+      <h1 css={tabRowStyles}>
```
oliverlloyd, line 157: I changed this so that there would be at least one `h1` tag on the page

src/client/pages/SignIn.tsx
```diff
@@ -29,64 +31,77 @@ const Links = ({ children }: { children: React.ReactNode }) => (
...
-              label="Password"
-              name="password"
-              type="password"
+export const SignIn = () => {
+  const clientState: ClientState = useContext(ClientStateContext);
+  const { pageData = {} } = clientState;
+  const { returnUrl } = pageData;
+  const returnUrlQuery = returnUrl
+    ? `?returnUrl=${encodeURIComponent(returnUrl)}`
+    : '';
```
oliverlloyd, line 40: At a later stage there's a case to have a `useReturnUrl` hook for this code

src/server/lib/idapi/auth.ts
```diff
@@ -45,13 +58,34 @@ export const read = async (
   };
   try {
     return responseToEntity(
-      (await idapiFetch(AUTH_URL, {
+      (await idapiFetch('/auth/redirect', {
```
oliverlloyd, line 61: I inlined this to reduce the need for scrolling

src/server/routes/signIn.ts
```diff
@@ -0,0 +1,67 @@
...
+      const html = renderer(Routes.SIGN_IN, {
+        requestState: state,
+        pageTitle: PageTitle.SIGN_IN,
+      });
+      return res.status(status).type('html').send(html);
+    }
+
+    trackMetric(Metrics.SIGN_IN_SUCCESS);
+
+    return res.redirect(303, returnUrl || defaultReturnUri);
```
oliverlloyd, line 63: In a later PR, perhaps the default uri should point to production, not code?

src/server/lib/idapi/auth.ts
coldlink: I wonder if there is a need for an enum of Identity URLs, similar to what we have for the internal routes in `shared/model/Routes.ts`. But this isn't a huge issue.

src/server/routes/index.ts
```diff
@@ -15,6 +16,9 @@ router.use(core);
 // request reset password routes
 router.use(noCache, reset);
 
+// request sign in routes
+router.use(noCache, signIn);
```
coldlink, line 20: another small thing, but not for this PR, I feel like `noCache` is used for most of our routes, so maybe we want to opt into routes that we would actually like to cache.

src/server/routes/signIn.ts
coldlink: it's just defined as an environment variable, so on `CODE` and `DEV` it points to `CODE`, while on `PROD` it points to `PROD`

src/client/pages/SignIn.tsx
```diff
@@ -29,64 +31,77 @@ const Links = ({ children }: { children: React.ReactNode }) => (
...
+
+  return (
+    <>
+      <Header />
+      <Nav
+        tabs={[
+          {
+            displayText: PageTitle.SIGN_IN,
+            linkTo: '/signin',
```
coldlink, line 49:
```suggestion
            linkTo: Routes.SIGN_IN,
```

src/client/pages/SignIn.tsx
```diff
@@ -29,64 +31,77 @@ const Links = ({ children }: { children: React.ReactNode }) => (
...
+            isActive: true,
+          },
+          {
+            displayText: PageTitle.REGISTRATION,
+            linkTo: '/register',
```
coldlink, line 54:
```suggestion
            linkTo: Routes.REGISTRATION,
```

coldlink (changes requested):
Looking good, main change request is to use the Routes enum in the navigation bar. Other comments are just general musings.

src/server/routes/index.ts
oliverlloyd: Yeah, so make not using a cache the default. This makes sense

src/server/lib/idapi/auth.ts
oliverlloyd: I actually find these kind of abstractions make it harder to read the code. I much prefer inlined, hardcoded values because it means I see right away what the value it and don't need to click around opening up new files to work it out, especially when the value is only ever used once.

Granted though, this is a subjective area!

coldlink (approved):

src/server/lib/idapi/auth.ts
coldlink: That's fair. Considering that the IDAPI endpoints are only ever specific to the route we're calling there's no real need for the abstraction, and they're only every usually being use in a single place.

PR #611, opened 2021-06-07
6th June 2021 | Updating dependencies

coldlink:
## What does this change?

## Nodes
- `@types/node` wasn't updated as this caused a clash with `express` types, so we can investigate that separately.

## Tested?
- [x] Local
- [x] CI
- [ ] CODE


oliverlloyd:
Seems like the ts bump is causing `chromatic` to throw

```/home/runner/work/gateway/gateway/node_modules/react-docgen-typescript/lib/parser.js:475
            var trimmedText = (tag.text || '').trim();
                                               ^
TypeError: (tag.text || "").trim is not a function```

coldlink:
> Seems like the ts bump is causing `chromatic` to throw
> 
> ```
>             var trimmedText = (tag.text || '').trim();
>                                                ^
> TypeError: (tag.text || "").trim is not a function```
> ```

Ah interesting. I'll delay on updating typescript too in this PR. Just trying to clean most of the dependabot PRs in one swoop. 😅

oliverlloyd (approved):

PR #612, opened 2021-06-07
Remove `form-action` CSP

oliverlloyd:
## What does this change?
Removes the Content Security Policy for `form-action`

This allows the workaround component `redirectRenderer` to be removed.

## Why?
Because it was blocking redirects in Chrom. See [this w3c discussion](https://github.com/w3c/webappsec-csp/issues/8) for more information

By removing the CSP rule we no longer need to use the workaround of forcing the browser to redirect the page using javascript, which isn't a great UX for the reader

## Why not?
There's a case to say this change remove some level of protection on the page, and perhaps that protection is more important than the improvement to UX? I'm not sure

### DIFF
--- a/src/client/pages/ChangePassword.tsx
+++ b/src/client/pages/ChangePassword.tsx
@@ -152,9 +152,8 @@ const usePasswordValidationHooks = (idapiBaseUrl: string) => {
   // errors go red color if the user has selected the confirm password input while the password criteria is not satisfied - or if the user submits the form and there are errors present
   // A user has to solve the problem indicated by the red error to make it go away
   // The red error is not necessarily the latest validation result - e.g. common password does not get solved by going beneath 8 characters
-  const [redError, setRedError, redErrorCurrently] = useRefState<
-    ErrorValidationResult | undefined
-  >(undefined);
+  const [redError, setRedError, redErrorCurrently] =
+    useRefState<ErrorValidationResult | undefined>(undefined);
 
   // store last length tick/cross error since we want to show it in green after it has been corrected (it can show either password too long, or too short)
   const [lastLengthError, setLastLengthError] =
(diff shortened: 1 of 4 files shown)


src/client/pages/ChangePassword.tsx
```diff
@@ -152,9 +152,8 @@ const usePasswordValidationHooks = (idapiBaseUrl: string) => {
   // errors go red color if the user has selected the confirm password input while the password criteria is not satisfied - or if the user submits the form and there are errors present
   // A user has to solve the problem indicated by the red error to make it go away
   // The red error is not necessarily the latest validation result - e.g. common password does not get solved by going beneath 8 characters
-  const [redError, setRedError, redErrorCurrently] = useRefState<
-    ErrorValidationResult | undefined
-  >(undefined);
+  const [redError, setRedError, redErrorCurrently] =
+    useRefState<ErrorValidationResult | undefined>(undefined);
```
oliverlloyd, line 156: Not sure why this prettier change popped up here but it's only that, a prettier diff

src/server/routes/consents.ts
```diff
@@ -433,7 +433,10 @@ function getABTestPOSTHandler(
 
     try {
       await entitySetter(req.ip, sc_gu_u, req.body);
-      const html = redirectRenderer(url);
+      const html = renderer(url, {
+        pageTitle: PageTitle.RESET_SENT,
+        requestState: state,
+      });
```
oliverlloyd, line 439: @coldlink is this right? Be good to get a second pair of eyes on this change

coldlink: In this case, since we've remove the `redirectRenderer` this should become a redirect.

so something like `return res.redirect(303, url)` i think

oliverlloyd: 🤦 This makes a lot more sense as a `res.redirect`!

coldlink (approved):

PR #636, opened 2021-06-14
14th June 2021 | Updating dependencies

coldlink:
## What does this change?

## Notes
- `@types/node` wasn't updated as this caused a clash with `express` types, so we can investigate that separately.
- `@babel/preset-react` wasn't updates as this causes an issue with visual regression in storybook

## Tested?
- [x] Local
- [x] CI
- [ ] CODE


oliverlloyd (approved):
Yup

oliverlloyd:
> Yup

That said, these visual diffs are wild

coldlink:
Yeah, not sure whats up with the visual diffs. I'll investigate.

coldlink:
very weird, caused by an update to `@babel/preset-react`, but it's only affecting storybook and not the main application 🤔

oliverlloyd:
> very weird, caused by an update to `@babel/preset-react`, but it's only affecting storybook and not the main application 🤔

Shall we hold off on that update for a while? Maybe if the code Gods are smiling on us by the time we revisit it there might be an update to SB which fixes it. If not, we can invest some time to investigate more

coldlink:
Yeah, I'll add it to the notes, and merge as is.

PR #637, opened 2021-06-15
Registration routing

oliverlloyd:
## What does this change?
This adds the routing and events for registering a user

### Wha? Why is there a password input there?
Our proposed UX and designs call for this form to only have an email input and for the password to be set later following email validation. I've added one here though because I wanted this form, and this PR, to return a successful result when working against `idapi`. Once we switch over to Okta, we can remove this field and adjust the backend actions

### How do we mark the user as logged in?
I've highlighted this question as I think it is relevant not just for the legacy `idapi` flow but also for the new flow as well. After registering an account (or adding a new password to one) how do we then get and set the required cookies? Do we need to make a second call to auth the user just created?


oliverlloyd:
[Registration routing](https://trello.com/c/wOIJ9Dlp/3018-registration-routing)

src/client/pages/Registration.tsx
```diff
@@ -15,49 +15,71 @@ import { Terms } from '@/client/components/Terms';
...
+    : '';
+
+  return (
+    <>
+      <Header />
+      <Nav
+        tabs={[
+          {
+            displayText: PageTitle.SIGN_IN,
+            linkTo: '/signin',
```
coldlink, line 35:
```suggestion
            linkTo: Routes.SIGN_IN,
```

src/client/pages/Registration.tsx
```diff
@@ -15,49 +15,71 @@ import { Terms } from '@/client/components/Terms';
...
+            isActive: false,
+          },
+          {
+            displayText: PageTitle.REGISTRATION,
+            linkTo: '/register',
```
coldlink, line 40:
```suggestion
            linkTo: Routes.REGISTRATION',
```

src/server/routes/register.ts
```diff
@@ -0,0 +1,76 @@
...
+  Routes.REGISTRATION,
+  handleAsyncErrors(async (req: Request, res: ResponseWithRequestState) => {
+    let state = res.locals;
+
+    const { email = '' } = req.body;
+    const { password = '' } = req.body;
+
+    const { returnUrl } = state.pageData;
+
+    const { defaultReturnUri } = getConfiguration();
```
coldlink, line 38:
```suggestion

```

src/server/routes/register.ts
```diff
@@ -0,0 +1,76 @@
...
+
+router.post(
+  Routes.REGISTRATION,
+  handleAsyncErrors(async (req: Request, res: ResponseWithRequestState) => {
+    let state = res.locals;
+
+    const { email = '' } = req.body;
+    const { password = '' } = req.body;
+
+    const { returnUrl } = state.pageData;
```
coldlink, line 36:
```suggestion
    const { returnUrl } = state.queryParams;
```
`state.queryParams.returnUrl` is typed so that it always exists, so you don't need the call to `getConfiguration` to get the `defaultReturnUri`

src/server/routes/register.ts
```diff
@@ -0,0 +1,76 @@
...
+      const html = renderer(Routes.REGISTRATION, {
+        requestState: state,
+        pageTitle: PageTitle.REGISTRATION,
+      });
+      return res.status(status).type('html').send(html);
+    }
+
+    trackMetric(Metrics.REGISTER_SUCCESS);
+
+    return res.redirect(303, returnUrl || defaultReturnUri);
```
coldlink, line 72:
```suggestion
    return res.redirect(303, returnUrl);
```

coldlink (changes requested):
A couple of small changes, but other than that looking good.

oliverlloyd:
@coldlink thanks for taking the time to suggest commit those changes, I've included them all

coldlink (approved):

PR #638, opened 2021-06-15
Authentication with Okta

coldlink:
## What does this change?

Implement the OKTA Authentication API in Gateway so that users can log in with their credentials against OKTA rather than IDAPI. See the flow diagram below of the flow that has been implemented.

## Prerequisite PRs
- #700 
- #701 
- #702 
- #703 
- #704 
- #723 

## Other project PRs
- https://github.com/guardian/identity-platform/pull/418
- https://github.com/guardian/identity/pull/1910

## Steps
- [x] Implement Okta authentication endpoint
- [x] Implement  certified OIDC Library - [`openid-client`](https://github.com/panva/node-openid-client)
- [x] Implement Authorization Code flow
  - [x] Redirect to Okta `/authorize` after authentication
  - [x] Implement `redirect_uri` endpoint which handles the return from okta `/authorize
  - [x] Exchange auth code for access token
- [x] Identity implementation to retrieve identity cookies
- [x] Redirect to `return_url` once authenticated
- [x] Error propagation
- [x] Fix cypress tests 
- [x] Add feature switch
  - [x] Switches setup PR #723
- [x] Reimplement Identity API sign in

## Pre-merge
- [x] Custom subdomain setup
  - [x] DEV
  - [x] CODE
  - [x] PROD 
- [x] Environment variables setup
  - [x] DEV
  - [x] CODE
  - [x] PROD
  - [x] CI

## Flow Diagram
Guardian Sign In (2)


src/server/lib/okta/oidc.ts
```diff
@@ -0,0 +1,272 @@
...
+// Identity Provider(IdP), Authorization Server(AS) and its metadata.
+const OIDCIssuer = new Issuer(OIDC_METADATA);
+
+// Encapsulates a dynamically registered, discovered or instantiated
+// OpenID Connect Client(Client), Relying Party(RP), and its metadata,
+// its instances hold the methods for getting an authorization URL,
+// consuming callbacks, triggering token endpoint grants, revoking and
+// introspecting tokens.
+// this is set to the "profile" client id and secret
+export const OIDCProfileClient = new OIDCIssuer.Client({
```
oliverlloyd, line 164: It might be simpler if we only export the properties we use rather than this whole object? It seems like we don't need to call it very often?

oliverlloyd (commented):
This is really great work!

I got most of the way through but then got stuck around the OIDC stuff. Maybe we can pair on that?

One general question: Is there a plan to build the sign in flow against idapi? If so, will this PR conflict with that or can we switch back?

src/server/lib/okta/error.ts
```diff
@@ -0,0 +1,14 @@
...
+  errorLink: string;
+  errorId: string;
+  errorCauses: [string?];
+}
+
+export interface FetchOktaError {
+  status: number;
+  statusText: string;
+  oktaError: OktaError;
+}
```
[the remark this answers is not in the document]
coldlink: > This is really great work!
> 
> I got most of the way through but then got stuck around the OIDC stuff. Maybe we can pair on that?
> 
> One general question: Is there a plan to build the sign in flow against idapi? If so, will this PR conflict with that or can we switch back?

Yeah sure, let's pair on this. I might take the engineers through it as a team too, since it would be useful to get multiple eyes on this. Discuss at standup.

src/server/lib/okta/authentication.ts
```diff
@@ -0,0 +1,163 @@
...
+
+  try {
+    const response = await fetch(`${oktaDomain}${API_ROUTE}`, requestOptions);
+    if (!response.ok) {
+      return await handleResponseFailure(response);
+    } else {
+      return await handleResponseSuccess(response);
+    }
+  } catch (error) {
+    throw error;
```
QuarpT, line 161: the try catch is unnecessary if the error is thrown

src/server/lib/okta/oidc.ts
```diff
@@ -0,0 +1,272 @@
...
+custom.setHttpOptionsDefaults({
+  agent: {
+    https: httpsAgent,
+  },
+});
+
+// https://developer.okta.com/docs/reference/api/oidc/#well-known-openid-configuration
+// this is the metadata returned from the /.well-known/openid-configuration
+// used by openid-client to create the "Issuer" and "Client"
+const OIDC_METADATA: IssuerMetadata = {
```
QuarpT, line 41: it would be good to create an oauth2 client which only supports the authorisation mechanisms and scopes we are using.

src/server/routes/oauth.ts
```diff
@@ -0,0 +1,109 @@
...
+      // we have the Authorization State now, so the cookie is
+      // no longer required, so mark cookie for deletion in the response
+      deleteAuthorizationStateCookie(res);
+
+      // exchange the auth code for access token + id token
+      // and check the "state" we got back from the callback
+      // to the "nonce" that was set in the AuthorizationState
+      const tokenSet = await OIDCProfileClient.oauthCallback(
+        // the redirectUri is the callback location (this route)
+        OIDCProfileClient.metadata.redirect_uris?.[0],
```
QuarpT, line 57: it was mentioned in the team review to refer to the constant rather than using `[0]`

src/server/lib/okta/authentication.ts
coldlink: Ah there isn't, it's a hangover from when I implemented it and was testing errors, try catch it's required here. Good spot.

src/server/lib/devHttpsAgent.ts
```diff
@@ -0,0 +1,25 @@
+import { Agent } from 'https';
```
QuarpT, line 1: it should be possible to get the self signed certificate confirmed with local node. This currently works with the Scala apps.

src/server/lib/idapi/auth.ts
```diff
@@ -23,6 +23,8 @@ const handleError = ({ error, status = 500 }: IDAPIError) => {
     const err = error.errors[0];
     const { message } = err;
 
+    logger.error(message);
```
QuarpT, line 26: it would be useful to log the stack trace

src/server/routes/oauth.ts
```diff
@@ -0,0 +1,129 @@
...
+      trackMetric(Metrics.AUTHORIZATION_SUCCESS);
+
+      // return to url from state or default url
+      return res.redirect(authState.returnUrl || defaultReturnUri);
+    } catch (error) {
+      logger.error(error);
+      trackMetric(Metrics.AUTHORIZATION_ERROR);
+      // if there's been an error from the authorization_code flow
+      // then propagate it as a query parameter (error + error_description)
+      if (error.error && error.error_description) {
```
QuarpT, line 114: do these error descriptions come directly from OKTA?
the URLs for identity-frontend are already quite long with propagated component event parameters and return URLs. A bit worried we might go over the maximum of ~2000 bytes.

QuarpT (approved):
lgtm, left some additional minor comments.
It would be good to confirm the nonce is tested against on the oauth callback. You could try modifying the nonce value to confirm it gets rejected.

src/server/lib/devHttpsAgent.ts
coldlink: Ah interesting! I've noticed that the certs for `gulocal.com` work great on my macbook, however on windows/linux (ubuntu) the certs aren't automatically whitelisted using `dev-nginx`, I'll play around further.

src/server/routes/signIn.ts
```diff
@@ -9,68 +8,187 @@ import { trackMetric } from '@/server/lib/trackMetric';
...
-    requestState: res.locals,
+    requestState: state,
     pageTitle: PageTitle.SIGN_IN,
   });
-  res.type('html').send(html);
+  return res.type('html').send(html);
 });
 
 router.get(
   Routes.SIGN_IN_CURRENT,
```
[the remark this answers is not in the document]
coldlink: It's not actually used. Was a placeholder for the current route on identity-frontend `/signin/current` used as part of the two step flow.

We can remove this here, and then redirect anyone on `/signin/current` with vcl to `/signin`

src/server/routes/oauth.ts
coldlink: Yeah they do, Okta will redirect to that endpoint with the `error` and `error_description` parameters if there is a problem with the OAuth flow

Searching for "List of Errors" on https://developer.okta.com/docs/reference/api/oidc/ as well as https://developer.okta.com/docs/reference/error-codes/#example-errors-for-openid-connect-and-social-login will show you which errors are possible

coldlink: We also haven't implemented the component event parameters yet, we may want to rethink how we send these parameters possibly.

src/server/lib/okta/authentication.ts
```diff
@@ -0,0 +1,171 @@
...
+export interface FetchOktaError {
+  status: number;
+  statusText: string;
+  oktaError: OktaError;
+}
+
+const { oktaDomain } = getConfiguration();
+
+const handleResponseFailure = async (response: Response) => {
+  try {
```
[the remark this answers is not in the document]
coldlink: I've refactored it so that it removes the try catch block

src/server/lib/idapi/auth.ts
coldlink: I'm gonna make a separate PR to refactor all the errors, we've been using them wrong the whole time:

From https://github.com/guardian/gateway/blob/496851c636f1d1b17f9ea341fcc78b9dfce973fb/src/server/lib/logger.ts#L14 we can see that the first parameter is a string message, and the second is the error, which we can use to get the stack trace in the line below that.

However we've just been passing the error as the first parameter instead, which means that it doesn't provide the stack.

cypress/integration/sign_in.spec.js
```diff
@@ -86,6 +88,131 @@ describe('Sign in flow', () => {
     });
   });
 
+  context.skip('Signing in - Okta (Skipped while feature disabled)', () => {
```
oliverlloyd, line 91: Is there a way to dynamically set the feature flag so this test can run?

cypress/integration/sign_in.spec.js
```diff
@@ -86,6 +88,131 @@ describe('Sign in flow', () => {
...
+      // so intercept the request and redirect to the guardian homepage
+      cy.intercept(
+        `http://localhost:9000/oauth2/customauthserverid/v1/authorize*`,
+        (req) => {
+          req.redirect('https://theguardian.com/');
+        },
+      );
+
+      cy.get('[data-cy=sign-in-button]').click();
+      cy.contains('UK edition');
```
oliverlloyd, line 212: It might be worth checking for a different string as I've had issues in the past where Github spins up an instance in a different region causing a different edition to load. Or maybe we just take note of this and deal with it if it happens?

src/client/pages/SignIn.stories.tsx
```diff
@@ -13,3 +13,22 @@ export const Default = () => <SignIn />;
 Default.story = {
   name: 'with defaults',
 };
+
+export const withEmail = () => <SignIn email="test@example.com" />;
```
oliverlloyd, line 17: Not sure if this makes much difference but I've always followed the convention of functions that return jsx having a leading capital letter - to mark them as components

src/client/pages/SignIn.tsx
```diff
@@ -31,77 +35,70 @@ const Links = ({ children }: { children: React.ReactNode }) => (
...
+        },
+      ]}
+    />
+    <Main>
+      <PageBox>
+        <PageBody>
+          <form
+            css={form}
+            method="post"
+            action={`${Routes.SIGN_IN}${queryString}`}
```
oliverlloyd, line 61: We could use `joinUrl` here? Although it's kind of overkill because we control both strings, what do you think?

src/client/pages/SignInPage.tsx
```diff
@@ -1,6 +1,23 @@
...
 export const SignInPage = () => {
-  return <SignIn />;
+  const clientState: ClientState = useContext(ClientStateContext);
+  const { pageData = {} } = clientState;
+  const { returnUrl, fieldErrors, email } = pageData;
+  const returnUrlQuery = returnUrl
+    ? `?returnUrl=${encodeURIComponent(returnUrl)}`
+    : '';
+
+  const errorSummary = fieldErrors?.find((e) => e.field === 'summary')?.message;
```
oliverlloyd, line 14: Not for this PR, but it feels a bit wrong to read the first item from the array, ignoring the rest. This obviously works here as there's only one field but I wonder if this api could be made better?

src/client/routes.tsx
```diff
@@ -16,9 +16,9 @@ import { UnexpectedErrorPage } from '@/client/pages/UnexpectedErrorPage';
 import { ConsentsFollowUpPage } from '@/client/pages/ConsentsFollowUpPage';
 import { ClientState } from '@/shared/model/ClientState';
 import { Routes } from '@/shared/model/Routes';
-import { SignInPage } from './pages/SignInPage';
-import { MagicLinkPage } from './pages/MagicLinkPage';
-import { MagicLinkSentPage } from './pages/MagicLinkSentPage';
+import { SignInPage } from '@/client/pages/SignInPage';
+import { MagicLinkPage } from '@/client/pages/MagicLinkPage';
+import { MagicLinkSentPage } from '@/client/pages/MagicLinkSentPage';
```
oliverlloyd, line 21: 👍

src/server/lib/okta/authentication.ts
```diff
@@ -0,0 +1,165 @@
...
+  } catch (error) {
+    throw new Error(`Error decoding JSON response: ${error}`);
+  }
+};
+
+// okta authentication endpoint
+// gateway (profile) is set up as a public application as it is
+// accessible by anyone
+// https://developer.okta.com/docs/reference/api/authn/#request-example-for-primary-authentication-with-public-application
+export const authenticate = async (email: string, password: string) => {
```
oliverlloyd, line 138: I wonder if there's benefit in calling this file `authenticate.ts`? I subscribe to the 'one thing per file' way of dong things - where a file has one main export which has the same name as the file - which could work here? The down side to an open bucket like this is it can be very inviting and can lead to lots of things being added, reducing discoverability.

It's true though that the one file per export thing doesn't always work, especially when there's lots of shared code. I also subscribe to the 'move stuff around until it feels right' way of doing things as well!

src/server/lib/okta/authentication.ts
```diff
@@ -0,0 +1,165 @@
...
+  const requestOptions: RequestInit = {
+    method: 'POST',
+    headers: {
+      Accept: 'application/json',
+      'Content-Type': 'application/json',
+    },
+    body,
+  };
+
+  const response = await fetch(`${oktaDomain}/api/v1/authn`, requestOptions);
```
oliverlloyd, line 159: `joinUrl`?

oliverlloyd (approved):
This is really great stuff!

My comments are all take it or leave it thoughts - feel free to act as you please on them.

My review was fairly light here and didn't dive into the flow or steps but we've reviewed this flow a lot now and I'm happy to move forward with Okta auth behind a flag so that we can do more end to end testing and validation prior to pulling back the curtain later.

src/server/lib/okta/authentication.ts
coldlink: Sounds good to me!

src/client/pages/SignIn.tsx
coldlink: Would joinUrl work in this case? As I'm appending the `queryString` parameters instead rather than a url.

I think there may be a case for a lib method to append query parameter strings to urls though, e.g. checks if a `?` exists in the url or not, and appends the query parameters as appropriate?

cypress/integration/sign_in.spec.js
coldlink: Not at the moment. We could rely on cookies or something in the future? But that's extra work outside this scope.

cypress/integration/sign_in.spec.js
coldlink: Not sure of what other string we could check for? Maybe it's worth checking for an element that always exists instead?

src/client/pages/SignIn.tsx
oliverlloyd: Ah yes, you're right, it won't work here! You could maybe use `URL`? Might be overkill though

cypress/integration/sign_in.spec.js
oliverlloyd: Strangely, I've had this problem a few times. There should probably be an element with a `data-cy-test` attribute we can check for

QuarpT:
I noticed when testing the login with gateway and IDAPI locally; I cannot login if I already have an OKTA session  which does not belong in the required groups. An existing session shouldn't interfere when authenticating as a different user.

cypress/integration/sign_in.spec.js
coldlink: Getting back to this PR after so long.. we changed it a while back to look at the URL after the page changed rather than the content.

src/client/pages/SignIn.tsx
coldlink: Again after so long...

We refactored completely how we do query params in this PR: https://github.com/guardian/gateway/pull/1109

src/client/pages/SignInPage.tsx
coldlink: We've removed this line, the error now comes in the same way as the other errors.

QuarpT (changes requested):
https://github.com/guardian/gateway/pull/638#issuecomment-883263327

dominickendrick:
I agree - we should not merge this until this https://github.com/guardian/gateway/pull/638#issuecomment-883263327 is resolved. Even if its behind a feature switch, we may forget to address it!

coldlink:
> I agree - we should not merge this until this [#638 (comment)](https://github.com/guardian/gateway/pull/638#issuecomment-883263327) is resolved. Even if its behind a feature switch, we may forget to address it!

How would we solve this issue though?
Would be helpful if we can address it sooner rather than later, as this work will help with the proof of concept for the apps stuff.

We technically have the same problem right now in identity too, where you can log in again as a different user despite being logged in as someone else.

So a few potential solutions would be:
a) somehow check for an existing session when authenticating and take action based on that?
b) block sign in pages from rendering if a user is logged in?
c) something else

rebecca-thompson (commented):
Looking good! Just a few comments. I'm also having some issues with errors in the OIDC flow where instead of being redirected to `/signin` I'm getting redirected to `/oauth/authorization-code/profile.thegulocal.com/signin` and getting a page not found error. Maybe it's something in my local set up?

dominickendrick:
Yeah, good suggestions. I checked what google do, and when you visit the sign-in page when your browser already has a session cookie (https://accounts.google.com/signin/v2/identifier) you are redirected to your account management page - https://myaccount.google.com/
This seems to fit with you a) and b) suggestions.

coldlink:
> Yeah, good suggestions. I checked what google do, and when you visit the sign-in page when your browser already has a session cookie (https://accounts.google.com/signin/v2/identifier) you are redirected to your account management page - https://myaccount.google.com/ This seems to fit with you a) and b) suggestions. We should redirect users to manage.theguardian.com if they are already logged in an visit the sign in page

I wonder if we should take a middleware approach then?
Have a middleware on all identity routes which check for an existing okta (or even identity session too?) and redirect then to manage based on the result.

src/server/routes/signIn.ts
```diff
@@ -10,22 +9,32 @@ import { Metrics } from '@/server/models/Metrics';
...
+
+    const { encryptedEmail, error, error_description } = state.queryParams;
+
     const email = encryptedEmail ? await decrypt(encryptedEmail, req.ip) : '';
+
     const errorMessage =
       error === FederationErrors.SOCIAL_SIGNIN_BLOCKED
         ? SignInErrors.ACCOUNT_ALREADY_EXISTS
-        : '';
+        : error_description;
```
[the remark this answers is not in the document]
coldlink: Yeah that might be helpful. The `error_description` itself is what's returned by the Okta OIDC enpoint: https://developer.okta.com/docs/reference/error-codes/#example-errors-for-openid-connect-and-social-login

coldlink:
Looks like we'd have to implement an extra API call to do this: https://developer.okta.com/docs/reference/api/sessions/#get-current-session

QuarpT:
> We technically have the same problem right now in identity too, where you can log in again as a different user despite already being logged in.

It's not the same. In Identity API the existing session will not prevent you signing in as a different user.

coldlink:
> Looking good! Just a few comments. I'm also having some issues with errors in the OIDC flow where instead of being redirected to `/signin` I'm getting redirected to `/oauth/authorization-code/profile.thegulocal.com/signin` and getting a page not found error. Maybe it's something in my local set up?

@rebecca-thompson Hmm.. I just tested the sign in flows again, and I've got it working on my local machine. Let me know when you're free and I can hopefully help diagnose whats happening!

src/server/lib/okta/authenticate.ts
```diff
@@ -0,0 +1,181 @@
...
+// Imports the node-fetch library asynchonously using import(), supported in CJS since node v12.17.
+// This solution avoids the import() being transpiled into a require() by Webpack/Typescript.
+// Subsequent fetch requests will not re-import the module because it's cached after first load.
+// Change necessary because of switch to ESM in version 3.x and our use of CJS.
+// Solution taken from: https://github.com/node-fetch/node-fetch/issues/1279#issuecomment-915063354
+const _importDynamic = new Function('modulePath', 'return import(modulePath)');
+async function fetch(...args: (string | RequestInit | undefined)[]) {
+  const { default: fetch } = await _importDynamic('node-fetch');
+  return fetch(...args);
+}
```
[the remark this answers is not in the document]
coldlink: moved it to `src/server/lib/IDAPIFetch.ts`

src/shared/model/ClientState.ts
```diff
@@ -15,6 +15,11 @@ interface ABTesting {
   forcedTestVariants?: Participations;
 }
 
+export interface FieldError {
+  field: string;
+  message: string;
+}
```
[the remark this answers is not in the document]
coldlink: removed!

src/shared/model/Errors.ts
```diff
@@ -75,6 +75,10 @@ export enum CsrfErrors {
   CSRF_ERROR = 'Sorry, something went wrong. If you made any changes these might have not been saved. Please try again.',
 }
 
+export enum OktaAuthenticateErrors {
+  AUTHENTICATION_FAILED = 'Email and password don’t match',
+}
+
```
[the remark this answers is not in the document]
coldlink: Good spot! I've refactored it to use `SignInErrors`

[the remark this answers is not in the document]
coldlink: There was actually a very slight difference in the apostrophe `Email and password don’t match` and `Email and password don't match`. I've changed it to use the curved one `’`

cypress/integration/mocked/sign_in.spec.ts
```diff
@@ -13,7 +13,9 @@ describe('Sign in flow', () => {
     });
 
     it('Has no detectable a11y violations on sign in page with error', () => {
-      cy.visit('/signin');
+      cy.visit(
+        '/signin?error=error&error_description=oh no somethings gone wrong',
```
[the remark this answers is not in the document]
coldlink: good spot! cypress automatically encodes it, but i've changed it to it's encoded form

coldlink:
Closing in favour of #1410 which has this and additional changes.

PR #660, opened 2021-06-22
`Container`

oliverlloyd:
## What?
This PR adds the latest version of `Container` from Source and then extends it and exports our own version of this component that better suits our needs.

## Why?
`Container` positions content based on the users breakpoint. It also abstracts and handles padding and borders. We want to still get this base styling and logic from Source but we have specific use cases for Gateway that we want to have our own abstraction for.

## Container
```
type Props = {
  children: React.ReactNode; // Children are inserted inside the nested div of the section
  sidePadding?: boolean; // Should side padding be added to the content inside the container (nested div)
  topBorder?: boolean; // Show top border
  sideBorders?: boolean; // Show left and right borders
  borderColor?: string; // Set the colour for borders
  backgroundColor?: string; // Sets the background colour of the section (root) element
};
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -104,7 +104,7 @@
     "@guardian/src-checkbox": "^3.7.0",
     "@guardian/src-foundations": "^3.7.0",
     "@guardian/src-icons": "^3.7.0",
-    "@guardian/src-layout": "3.6.0",
+    "@guardian/src-layout": "3.7.0",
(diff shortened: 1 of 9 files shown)


oliverlloyd:
[Identity Container](https://trello.com/c/hTAZfllj/3077-identity-container)

coldlink (approved):

PR #669, opened 2021-06-23
`yarn tsc`

oliverlloyd:
## What does this change?
Replaces `yarn type-check` with `yarn tsc` (+ shorten the command for prettier)

## Why?
Because I have muscle memory for `yarn tsc` and keep trying to use it!

Most other repos in the dept. (or at least the ones I've worked on anyway) use `yarn tsc` so I think it makes sense to align on that.


coldlink (approved):
Don't really have a strong opinion on this, so it's okay. +1 for the slightly shortened commands though.

PR #674, opened 2021-06-24
Emails

oliverlloyd:
## What does this change?
Adds the framework for an email service.

### Features

- Uses [MJML](https://mjml.io/) for responsive email rendering
- Templates can be previewed in Storybook using `testUtils.tsx`
- Email templates can be composed using Components, each of which is also viewable in Storybook

### Usage
To send an email

```js
import { sendExampleEmail } from '@/email';

try {
  await sendExampleEmail({ to: 'person@example.com' });
} catch(error) {
  // Handle error
}
```

### Pending

- Creating more components to make composing new templates easier
- Consider rate limiting
- Logging


oliverlloyd:
[Gateway Email Service](https://trello.com/c/Uoau1lSa/3092-gateway-email-service)

coldlink (approved):
LGTM 👍 May be worth getting someone else on the team to review it, as I did some of the initial work.

src/server/routes/email.ts
```diff
@@ -0,0 +1,25 @@
+import { Router, Request, Response } from 'express';
+import { sendExampleEmail } from '@/email';
+
+const router = Router();
+
+router.get('/send-example-email?:to', async (req: Request, res: Response) => {
```
QuarpT, line 6: Is this endpoint accessible to the Internet?

We should add protection around this to avoid abuse.

oliverlloyd: It's not public until the Fastly config also exists. But I discussed this with Mahesh and he pointed out that even without that config there's a risk it could be added by mistake later on.

I'm going to remove this code for now. It was useful as an example for how an email could be sent but that now this also exists in the README so this code is redundant

QuarpT: 👍  for removal

> It's not public until the Fastly config also exists

The load balancer address might be accessible to the internet depending on the security group config.

QuarpT (approved):

PR #702, opened 2021-06-30
`ErrorSummary` Component + Guardian fonts in Storybook

coldlink:
## What does this change?
- Adds an `ErrorSummary` component, to be used as an error on the sign in page
  - [Figma Link](https://www.figma.com/file/ZDPonnBXbMhV2smf51Dylc/Sign-in-Journeys?node-id=756%3A14083)
  - This component [hasn't been developed in Source](http://theguardian.design/2a1e5182b/p/108ed3-user-feedback/b/3803b4/t/08c895) yet, so is a candidate to be moved over there too, hence I added in the "context" prop.
    - The `SuccessSummary` component would still need to be developed to if bringing it over to source, but since we're not using it, I haven't developed it
- Fixes missing Guardian fonts in Storybook by adding the `@font-face` to the `preview-head.html`, this was copied from the rendered HTML from a gateway page
  - Hence the lots of UI test changes

#### Default
chrome_6U0BfC5SXg

#### withContext
chrome_ajf7f1g80d


src/client/components/ErrorSummary.tsx
```diff
@@ -0,0 +1,59 @@
...
+import { css, SerializedStyles } from '@emotion/react';
+import { error, space } from '@guardian/src-foundations';
+import { textSans } from '@guardian/src-foundations/typography';
+import { SvgAlertTriangle } from '@guardian/src-icons';
+import { size } from '@guardian/src-foundations/size';
+
+interface ErrorSummaryProps {
+  error: string;
+  context?: string;
+  cssOverrides?: SerializedStyles | SerializedStyles[];
```
oliverlloyd, line 11: I've been removing this pattern when I see it.

Why have I been doing this?
2 reasons:

1. Allowing people to specify their own custom styles that overrides those of the component is a useful pattern for authors of shared UI libraries, like Source. It's useful because you _can't know_ all the possible use cases and don't want to block people from doing what they need. So you create the 'recommended' variations of a component but give users an escape hatch. In many cases though, these escape hatches end up being closed and new props being added to deal with them.

But here, where we control both the code inside the component and the code calling it, there's no need for this pattern. If there is a requirement for a new use case, then we can add a new prop or maybe restructure the component (or create a new one - whatever works best). We don't need a temporary escape hatch.

2. Why not just add it? Because your props go from a tightly controlled interface where anyone can read them and understand the possibilities, to _the entire css api_. Which is really big 😉

oliverlloyd (commented):
This is great. Really useful write up and very readable code!

I actually might have a potential use case I'm working on for the `SuccessSummary` too. For when the user clicks `Resend Email` on a confirmation page we might want to give more context in the message that gets shown

Just one comment around `cssOverrides`. I don't think we should use this prop pattern in Gateway, it's probably only useful in Source but even there it's not encouraged.

src/client/components/ErrorSummary.tsx
coldlink: Ah, I did not realise that! That's useful to know, I'll remove the `cssOverrides` pattern here.

oliverlloyd (approved):
Thanks for adding those fonts to storybook by the way, that's been bugging me for ages! 😄

PR #704, opened 2021-06-30
Move `FieldError` type from `server` to `shared`

coldlink:
This is because that the shared folder should not inherit from either the client or server folder.

## Before
`FieldError` was in `@/server/routes/changePassword` however this was being used in the `shared` file `@/shared/model/ClientState`

## After
`FieldError` defined in `@/shared/model/ClientState`, everywhere else inherits from this.


oliverlloyd (approved):
Nice

PR #706, opened 2021-06-30
28th June 2021 | Updating dependencies

coldlink:
## What does this change?

## Notes
- `mjml-react` was a major version update, but this won't cause us any problems, as it was to set a specific babel runtime version, one that we already use (`7.14.6`). See https://github.com/mjmlio/mjml/issues/2299#issuecomment-864526891
- `wait-on` was a major version update, but this remove node 8, and no other changes, we're on node 14, so this is fine
- `@types/node` wasn't updated as this caused a clash with `express` types
- `@babel/preset-react` wasn't updates as this causes an issue with visual regression in storybook

## Tested?
- [x] Local
- [x] CI
- [x] CODE


oliverlloyd (approved):
I'm happy if TeamCity is happy 👍

coldlink:
> I'm happy if TeamCity is happy 👍

I did a run through on `CODE` and it's working great

PR #707, opened 2021-07-01
`Link` (for emails)

oliverlloyd:
## What does this change?
Adds the `Link` component for emails and uses it in the `Footer`

We now use the `style` attribute to attempt to override the default a tag colour and underline text decoration. How things render in Storybook is not representative of real life. Success is dependant on how different email clients implement things.

I based the `text-decoration` code on [this SO answer](https://stackoverflow.com/questions/8998378/how-do-i-remove-link-underlining-in-my-html-email) but could no implement the full solution because you can't have multiple properties with the same name in JSX

### Before
Screenshot 2021-07-01 at 13 14 14

### After
Screenshot 2021-07-01 at 13 14 02

### DIFF
--- a/src/email/components/Footer.tsx
+++ b/src/email/components/Footer.tsx
@@ -1,7 +1,8 @@
 import React from 'react';
-
 import { MjmlSection, MjmlColumn, MjmlText } from 'mjml-react';
+
 import { background, text } from '@guardian/src-foundations/palette';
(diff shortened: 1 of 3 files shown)


oliverlloyd:
[Style email links](https://trello.com/c/DetjYMfw/3176-style-email-links)

coldlink (approved):

PR #724, opened 2021-07-07
Refactor password breach check

oliverlloyd:
## What?
This PR refactors the way that we check if passwords have been breached.

Previously, we:

1. sent requests to `idapi` and
2. used custom code to prevent spamming the endpoint as users type

We now send the request directly to the HaveIBeenPawned api and use the NPM lib AwesomeDebouncePromise to debounce requests, allowing `ThrottledBreachedPasswordCheck.ts` to be removed.

## Why?
By not using `idapi` we make it easier to develop as requests work locally (`idapi` won’t allow requests from localhost) and we also progress the migration by removing the dependency on this endpoint.

By using `AwesomeDebouncePromise` we simplify our code, offloading that work onto a third party library at the very small cost of 1KB onto the bundle.


QuarpT:
This looks great. We lose metrics around pwned passwords API provided by the IDAPI endpoint - but they are not essential.

oliverlloyd:
[Refactor password validation](https://trello.com/c/Z26jsRkC/3188-refactor-password-validation)

QuarpT (approved):

PR #729, opened 2021-07-09
Use a token in the NoAccount email

oliverlloyd:
## What does this change?
When reader's try to reset the password using an email that doesn't exist then the `NoAccount` email needs a token so that we can set their password when we land them in onboarding

### DIFF
--- a/src/email/templates/NoAccount/NoAccount.tsx
+++ b/src/email/templates/NoAccount/NoAccount.tsx
@@ -24,7 +24,9 @@ export const NoAccount = () => {
         </p>
         <p>Please click below to register.</p>
       </Text>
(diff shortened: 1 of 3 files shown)


oliverlloyd:
[NoAccount token](https://trello.com/c/Lml2nRbl/3211-noaccount-token)

coldlink:
Hmm.. if an account doesn't exist. I don't think we need a token. We should just ask the user to register, and link them to the register page.

The main reason is technical, if the user doesn't have an account, we can't associate a token to them. We don't want want to make an account either to associate a token to as the request may not be from the same person the email belongs to, e.g. spam/bot trying out emails.

oliverlloyd:
Closing as per @coldlink 's comment and our discussions

PR #738, opened 2021-07-14
Refactoring Error Handling

coldlink:
## What does this change?
@QuarpT noted in https://github.com/guardian/gateway/pull/638#discussion_r668628677 that it would be useful to log the stack trace of a specific error.

However I noticed that we were using the `logger.error` method wrong the entire time. In https://github.com/guardian/gateway/blob/496851c636f1d1b17f9ea341fcc78b9dfce973fb/src/server/lib/logger.ts#L14 the first parameter is a message, and the second the error. However we were passing the error as the first argument, which was just sent as a useless string.

This PR refactors and adds errors throughout the project with a useful message, and the error in the correct parameter.

It also removes unnecessary try catch blocks too.

### DIFF
--- a/src/server/lib/idapi/changePassword.ts
+++ b/src/server/lib/idapi/changePassword.ts
@@ -11,6 +11,7 @@ import {
   IdapiErrorMessages,
   ChangePasswordErrors,
 } from '@/shared/model/Errors';
+import { logger } from '../logger';
 
 const handleError = ({ error, status = 500 }: IDAPIError) => {
   if (error.status === 'error' && error.errors?.length) {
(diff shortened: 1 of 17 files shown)


src/server/lib/idapi/changePassword.ts
```diff
@@ -67,6 +68,10 @@ export async function change(password: string, token: string, ip: string) {
     );
     return result.cookies;
   } catch (error) {
+    logger.error(
+      `IDAPI Error changePassword change ${ApiRoutes.CHANGE_PASSWORD}`,
```
oliverlloyd, line 72: Are we not using `url` here because the `qs` string causes issues?

oliverlloyd: If so, maybe we can encode it?

oliverlloyd (approved):
This looks great!

One question: What will happen if someone does `logger.error(e)`? I can see this being something people might expect the api to support

coldlink:
> This looks great!
> 
> One question: What will happen if someone does `logger.error(e)`? I can see this being something people might expect the api to support

this is what's happening at the moment, whats logged out is something like `{ "message": "Error" }` which is super unhelpful

PR #739, opened 2021-07-14
No password confirmation

oliverlloyd:
## What does this change?
This PR removes the second password input field making it so the reader only has to type a new password once.

### What else does this PR do?
I ran into some issues changing the Cypress tests so to make this easier I removed the `page` abstraction and just inlined the commands, like a classic, vanilla script. This helps because you can instantly see what a test is doing and you know that any changes you make will have no side affects in other places

I also renamed the header from 'Reset password' to 'Set password' - this is to be confirmed

Screenshot 2021-07-14 at 17 52 29


oliverlloyd:
[No password confirmation](https://trello.com/c/0QurMNR9/3227-no-password-confirmation)

cypress/integration/change_password.spec.js
```diff
@@ -106,67 +106,63 @@ describe('Password change flow', () => {
...
-      cy.contains(ChangePasswordPage.CONTENT.PASSWORDS_MATCH);
-      page.clickPasswordChange();
+      cy.visit(`/reset-password/fake_token`);
 
-      cy.contains(ChangePasswordPage.CONTENT.PASSWORD_CHANGE_SUCCESS_TITLE);
-      cy.contains(ChangePasswordPage.CONTENT.CONTINUE_BUTTON_TEXT).should(
+      cy.get('input[name="password"]').type('thisisalongandunbreachedpassword');
+      // there seems to be a race condition on the change password page
+      // adding a small wait fixes the issue
+      cy.wait(600);
```
QuarpT, line 162: some waits are longer than others. Is this necessary?

cypress/support/pages/change_password_page.js
```diff
@@ -1,54 +0,0 @@
-class ChangePasswordPage {
```
QuarpT, line 1: Are we moving away from these test helper classes?

QuarpT (approved):
I like the simplification a lot, but it would be good to confirm with the team whether we are happy to ignore the OWASP guideline for this scenario https://cheatsheetseries.owasp.org/cheatsheets/Forgot_Password_Cheat_Sheet.html#user-resets-password

cypress/integration/change_password.spec.js
oliverlloyd: Mayyybe 🤷  - I suspect this wait is needed as there's a race around how long the api call to haveibeenpawnd in which case you're right, it should probably be the same. I'll update that.

Going forward we should keep an eye on this and adjust as required.

QuarpT (approved):

cypress/integration/change_password.spec.js
oliverlloyd: I've fixed this in a later PR by using `intercept()` to make the wait actually _wait_ for the call

PR #740, opened 2021-07-16
Refactor password validation

oliverlloyd:
## What does this change?
This PR refactors `ChangePassword` to simplify the validation logic.

Designs [here](https://www.figma.com/file/z74mPkMGzcATz3ZxrHetrs/Onboarding-UX?node-id=2169%3A2719).

## `Checking...`
I added a small loading state where if the fetch request to haveibeenpwned.com then the validation message reads as `Checking...`

## Supporting text
We now always show the supporting text `Must be between 8 and 72 characters` above the input

## `Valid password`
We now show success text when the input is valid


oliverlloyd:
[Refactor PasswordInput validation](https://trello.com/c/ozepMaJI/3221-refactor-passwordinput-validation)

src/client/pages/ChangePassword.tsx
```diff
@@ -17,203 +17,251 @@ import { Main } from '@/client/layouts/Main';
...
-  password: string,
-): Promise<PasswordValidationResult> => {
-  const lengthResult = validatePasswordLength(password);
-  if (lengthResult !== PasswordValidationResult.VALID_PASSWORD) {
-    return Promise.resolve(lengthResult);
-  }
-
-  return checkIfBreached(password);
-};
+const ValidationSymbol = ({ result }: { result: Validation }) => {
```
QuarpT, line 41: it might be easier to split this into 3 separate components which share baseStyles

oliverlloyd: Yeah, I think you're right. After the changes I've been making this would be the next logic thing to do here.

src/client/pages/ChangePassword.tsx
```diff
@@ -17,203 +17,251 @@ import { Main } from '@/client/layouts/Main';
...
+  isTooLong: boolean;
+  isChecking: boolean;
+}) => {
+  if (isTooShort) {
+    return <TooShort />;
+  } else if (isTooLong) {
+    return <TooLong />;
+  } else if (isWeak) {
+    return <Weak />;
+  } else if (isChecking) {
```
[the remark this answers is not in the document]
oliverlloyd: Yes! I think this is actually a bug - I remember now thinking that it was odd 'checking...' never showed at certain points. Great spot!

QuarpT:
Great work. The code is a lot easier to understand and the new designs look good!

Left a couple of minor comments.

src/client/pages/ChangePassword.tsx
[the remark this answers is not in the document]
oliverlloyd: Fixed now 👍 thanks again

src/client/pages/ChangePassword.tsx
oliverlloyd: Done

QuarpT (approved):
🚢

PR #748, opened 2021-07-20
`EmailSent` - Resend email

oliverlloyd:
## What does this change?

1. Creates a generic email sent page (deleting `MagicLinkSent` & `ResetLinkSent`)
2. Adds the ability to resend the email from this page
3. Adds a link to take the user back to where they initiated the action (so they can enter another email)

#### Session storage
This page is loaded via a form redirect so we can't pass down state the way we might normally do. To get around this we will store the users email in a cookie so that it can be read from there on server render. 

#### Hidden field
To POST the request to resend the email I am using a form with a hidden input field that contains the email.

#### `previousPage`
`EmailSent` can be called from multiple places, this prop the contract for what to do when users click the link or press the Resend button


oliverlloyd:
[EmailSent Page](https://trello.com/c/LTfI9JBD/3179-emailsent-page)

src/client/pages/EmailSent.tsx
```diff
@@ -0,0 +1,81 @@
...
+            ) : (
+              <PageBodyText>We’ve sent you an email.</PageBodyText>
+            )}
+            <PageBodyText>
+              Please follow the instructions in this email. If you can&apos;t
+              find it, it may be in your spam folder.
+            </PageBodyText>
+            <PageBodyText>The link is only valid for 30 minutes.</PageBodyText>
+            {email && previousPage && hasJS && (
+              <form method="post" action={previousPage}>
```
QuarpT, line 45: is previousPage reliable given the user could navigate to the page in an unexpected way?
It might be worth referring to the actual route to POST against.

QuarpT (approved):
looks good!

PR #751, opened 2021-07-23
Gateway sends reset password email

oliverlloyd:
## **Still pending**
We cannot yet use Gateway for sending emails until we implement a solution for rate limiting

## What
This PR moves the responsibility for sending the reset password email away from idapi and into Gateway

The sudo logic for this flow looks a bit like this:

```
// User makes request to reset password
if (isNewUser) {
  sendNoAccountEmail()
} else {
  var token = generateToken(email);
  sendResetPasswordEmail({ token, to: email });
}
```

The actual code is pretty similar:

```js
      const userType = await getType(email, req.ip);

      switch (userType) {
        case 'new': {
          sendNoAccountEmail({ to: email });
          break;
        }
        case 'guest':
        case 'current': {
          const token = await getToken(email, req.ip);
          sendResetPasswordEmail({ token, to: email });
          break;
        }
        default: {
          // This shouldn't happen but these external response types are not typed and
          // could change so we handle this here by triggering a generic error
          throw new Error();
        }
      }
```

This adds 2 new functions:

### `getType`
Calls `/user/type/${email}` asking for the type of user

### `getToken`
Calls `/pwd-reset/token` to get a new reset token

And adds 2 new emails:

### `sendNoAccountEmail`
Sends the existing `NoAccount` email

### `sendResetPasswordEmail`
Sends the existing `ResetPassword` email

## Why
Because idapi won't be around forever

## Why are the tests failing?
The `cypress-ete` tests are not working because we do not have AWS keys set as environment variables to enable the Github Action instance to connect to AWS SES.


oliverlloyd:
Once #787 is merged this PR will become a lot simpler and more focussed on the act of moving sending the reset password email into Gateway

oliverlloyd:
This draft is no longer needed

PR #752, opened 2021-07-26
Cypress: `yarn cypress` cmd

oliverlloyd:
## What does this change?
Move the `open` command that is used to run Cypress is headed mode out of the bash script and into `package.json

## Why?
Because this means we can now use both make and yarn to run Cypress

### DIFF
--- a/cypress-open.sh
+++ b/cypress-open.sh
@@ -13,4 +13,4 @@ yarn mock-server &
 yarn wait-on:mock-server
 yarn start &
 yarn wait-on:server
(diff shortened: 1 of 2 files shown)


coldlink (approved):

oliverlloyd:
Closing this because TeamCity is not happy about this change. I suspect we're using an image which has something baked into it that needs the commands how they were but I'm not sure to be honest. Either way, this was a convenience PR and can be closed.

PR #756, opened 2021-07-26
Fix Cypress race condition

oliverlloyd:
## What does this change?
There was a race condition in the Cypress tests where the breach check took an unknown length of time. We were using `cy.wait(600)` to get around this but it wasn't reliable.

## Now
We're using `cy.intercept()` to spy on the actual call and then `cy.wait`() for this.

### DIFF
--- a/cypress/integration/change_password.spec.js
+++ b/cypress/integration/change_password.spec.js
@@ -53,9 +53,6 @@ describe('Password change flow', () => {
       cy.mockNext(200);
       cy.visit(`/reset-password/fake_token`);
       cy.get('input[name="password"]').type('short');
-      // there seems to be a race condition on the change password page
-      // adding a small wait fixes the issue
-      cy.wait(300);
       cy.get('button[type="submit"]').click();
       injectAndCheckAxe();
     });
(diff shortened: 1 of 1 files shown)


cypress/integration/change_password.spec.js
```diff
@@ -114,37 +113,42 @@ describe('Password change flow', () => {
...
-      // adding a small wait fixes the issue
-      cy.wait(300);
+      cy.wait('@breachCheck');
+      cy.contains('use a password that is hard to guess');
       cy.get('button[type="submit"]').click();
-      cy.contains('common password');
+      cy.contains('is a common password');
+      cy.get('input[name="password"]').type('iamaveryuniqueandlongstring');
+      cy.wait('@breachCheck');
+      cy.contains('At least 8 characters'); // This means success 🤷‍♂
```
QuarpT, line 128: maybe checking the colour would be useful if this message also gets shown for error

QuarpT (approved):

cypress/integration/change_password.spec.js
oliverlloyd: I _think_ that the next PR in this space will change this to `cy.contains('Valid email'); ` (or some similar text) so I deliberately left this vague!

PR #764, opened 2021-07-28
Run Cypress in Github

oliverlloyd:
## What does this change?
Moves the execution of Cypress from Team City to Github

Refactors and renames things to make it clear that we're running Cypress in mocked mode

## Why?
Because using Github Actions is faster, and simpler. 

And also because we want to extend our Cypress suite at a later stage to include tests that don't use the mock server and it's easier to do this by simply adding a new Github action than it would be with the old method

## Too many commits?
I'll squash this one!


package.json
```diff
@@ -19,7 +19,7 @@
     "tsc": "tsc --noEmit --skipLibCheck",
     "watch:server": "nodemon --inspect ./build/server.js",
     "watch": "webpack --watch",
-    "wait-on:server": "wait-on http://localhost:$PORT/healthcheck",
+    "wait-on:server": "wait-on http://localhost:8861/healthcheck",
```
oliverlloyd, line 22: Dotenv is run after this point so we don't have access to env vars here anymore

util/mock-server.js
```diff
@@ -81,25 +81,25 @@ app.all('*', (req, res) => {
   if (permanent.has(req.path)) {
     const { status, payload } = permanent.get(req.path);
     res.status(status).json(payload);
-    console.log(`Mocking: ${req.originalUrl}: ${status} ${inspect(payload)}`);
+    // console.log(`Mocking: ${req.originalUrl}: ${status} ${inspect(payload)}`);
```
oliverlloyd, line 84: I commented out these logs because they're quite verbose and make it hard to read things

webpack.config.js
```diff
@@ -222,10 +227,12 @@ const browser = ({ isLegacy }) => {
...
-    plugins: [new AssetsPlugin({
-      path: path.resolve(__dirname, 'build'),
-      filename: `${isLegacy ? 'legacy.' : ''}webpack-assets.json`
-    })],
+    plugins: [
+      new AssetsPlugin({
+        path: path.resolve(__dirname, 'build'),
+        filename: `${isLegacy ? 'legacy.' : ''}webpack-assets.json`
+      })
+    ],
```
oliverlloyd, line 235: This is just a prettier format diff

webpack.config.js
```diff
@@ -58,7 +59,11 @@ const common = ({ platform }) => ({
       '@': path.join(__dirname, 'src'),
     },
   },
-  watchOptions
+  watchOptions,
+  plugins: [new Dotenv({
+    // Required to ensure Dotenv doesn't override any existing env vars that have been set by the system
+    systemvars: true
+  })],
```
oliverlloyd, line 66: Adds `dotenv-webpack` to read in vars from `.env`

util/mock-server.js
QuarpT: maybe we could put these behind a debug flag

src/server/lib/middleware/login.ts
```diff
@@ -16,8 +15,8 @@ export const loginMiddleware = async (
   res: Response,
   next: NextFunction,
 ) => {
-  const config = getConfiguration();
-  const LOGIN_REDIRECT_URL = config.signInPageUrl;
+  const LOGIN_REDIRECT_URL =
+    process.env.SIGN_IN_PAGE_URL || 'https://profile.theguardian.com/signin';
```
QuarpT, line 19: why do we need this fallback?

oliverlloyd: Because `process.env.THING` could be undefined. In reality it won't be but typescript can't know that

QuarpT: getConfiguration does not return an optional (since it throws an exception), can we use that?

oliverlloyd: Yeah, good call. I moved away from that because this particular env var gets passed in via two methods (which confused me for a while!) but it turns out this is correct so I can revert to the original function

coldlink: Seems you're fixing this. But yeah, in agreement with Peter here, any environment variables required by the server should go through `getConfiguration` first.

util/mock-server.js
coldlink: We might want to add a `logger.debug` method that only runs in dev/test. We already have prior art in `logger.info` `logger.error` and `logger.warn` so could just extend that.

webpack.config.js
coldlink: Might just be misunderstanding this, but how would this work in production?

The project is built in teamcity, and built version is pushed to S3, which is downloaded and used by the server in AWS. We set all the environment variables for production as system environment variables on the server itself which we want to access at runtine. Whereas this plugin seems like it overwrites anything with `process.env` at build time.

oliverlloyd: it should work the other way, any existing system variables should win out where there is a conflict.

Defo one to test on `code` though!

src/server/lib/middleware/login.ts
oliverlloyd: That's now done

webpack.config.js
coldlink: I think from the understanding of the plugin, it replaces anything that is `process.env` at build time, so using the `systemvars` flag it would only pick up system environment variables set on the teamcity server. But yeah do test on `CODE`, and maybe check what's outputted in the bundle when running `yarn build` locally, since we still want `process.env` to be referenced in the bundle.

util/mock-server.js
oliverlloyd: Do we need these console logs? They feel like they were useful for development but I don't think they add much benefit going forwards. If there's a use case sometime in the future where a developer needs to know what might be happening somewhere in the code they can add console statements then, this is pretty normal plus that way they can only log the thing they need to know, instead of 1000s of lines

coldlink: They are super useful in development, especially when setting up the tests or when debugging failed tests since it shows us what's being mocked and where.

oliverlloyd: Honestly, I found them not useful at all - I wasn't able to understand why the tests were failing because there was _so much_ stuff being logged out. Hence why I commented them out. 🤷 

I can always just add them back though if there's a general feeling that they're worth keeping?

webpack.config.js
oliverlloyd: From my testing, without the `systemvars` flag, Dotenv will crash the party and override any system variables already set. This wasn't what we want so I set it to true and now it's more gentle.

I'm assuming it does this on a case by case basis, and doesn't destroy the entire `process.env` object

oliverlloyd:
[End to end test released](https://trello.com/c/txLG5ccY/3252-end-to-end-test-released)

QuarpT:
do the cypress tests against the mocked environment get run in CI?
If not the mocked tests will probably stop working and not get fixed

oliverlloyd:
> do the cypress tests against the mocked environment get run in CI?
> If not the mocked tests will probably stop working and not get fixed

They get run [here](https://github.com/guardian/gateway/pull/764/checks?check_run_id=3191983497) now as a separate Github Action

oliverlloyd:
> > do the cypress tests against the mocked environment get run in CI?
> > If not the mocked tests will probably stop working and not get fixed
> 
> They get run [here](https://github.com/guardian/gateway/pull/764/checks?check_run_id=3191983497) now as a separate Github Action

The idea is to have one action for mocked tests, using localhost:9000, and then, later, to add a new set of tests that run against `code`. That way we:

- get faster results, earlier
- can see the results without being on the vpn (very handy)
- can see at a glance if it was just mocked or live tests that failed (useful for bisection debugging)
- etc.

webpack.config.js
oliverlloyd: I deployed this branch to `code` and this page loads and works as expected. What do you think, does this give us the required level of confidence to move forward?

QuarpT:
> > > do the cypress tests against the mocked environment get run in CI?
> > > If not the mocked tests will probably stop working and not get fixed
> > 
> > 
> > They get run [here](https://github.com/guardian/gateway/pull/764/checks?check_run_id=3191983497) now as a separate Github Action
> 
> The idea is to have one action for mocked tests, using localhost:9000, and then, later, to add a new set of tests that run against `code`. That way we:
> 
> * get faster results, earlier
> * can see the results without being on the vpn (very handy)
> * can see at a glance if it was just mocked or live tests that failed (useful for bisection debugging)
> * etc.

ah great, this makes sense - and I see the config in .github/workflows/cypress.yml 👍

QuarpT (approved):
🚀

PR #765, opened 2021-07-30
Update `Verify` email to new copy

coldlink:
## What does this change?
Screenshot_2021-07-30_at_10 40 40

Change the `Verify` email to use the updated copy.

It also switches to using `$variable` syntax for insertable variables for now, as this email will be being sent from Identity API for the time being, which expects variables in that format.

`$registerAction` will become one of `registering`, `your contribution`, or `subscribing` depending on the action that the user was taking.

`$verificationLink` will become the link to the check token/welcome page. This page has not been set up yet.

### DIFF
--- a/src/email/templates/Verify/Verify.tsx
+++ b/src/email/templates/Verify/Verify.tsx
@@ -1,7 +1,5 @@
 import React from 'react';
 
-import { getProfileUrl } from '@/server/lib/getProfileUrl';
-import { Routes } from '@/shared/model/Routes';
 import { Page } from '@/email/components/Page';
 import { Button } from '@/email/components/Button';
 import { Header } from '@/email/components/Header';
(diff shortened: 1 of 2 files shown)


QuarpT (approved):

src/email/templates/Verify/Verify.tsx
```diff
@@ -1,31 +1,22 @@
...
 import { Footer } from '@/email/components/Footer';
 
-const profileUrl = getProfileUrl();
-
 export const Verify = () => {
   return (
     <Page>
       <Header />
-      <SubHeader>Please verify your email</SubHeader>
+      <SubHeader>Thank you for $registerAction</SubHeader>
```
oliverlloyd, line 14: This seems to be introducing a new convention for how text is replaced. Other areas of the code use upper case text like `TOKEN_PLACEHOLDER`. 

I'm happy to go either way but I think we should be consistent.

src/email/templates/Verify/VerifyText.ts
```diff
@@ -1,11 +1,11 @@
...
 export const VerifyText = () => `
+Thank you for $registerAction
+
 Welcome to the Guardian,
 
-Please click below to verify your emails address and complete your registration.
+Please click below to complete your registration.
 
-https://profile.theguardian.com${Routes.WELCOME}
+$verificationLink
```
oliverlloyd, line 8: Again, happy to go either way, but the other emails construct the url by only inserting the token. Shall we move everything over to this method?

oliverlloyd (approved):
Shall we update the other emails to also use the `$varName` approach for placeholders? If identity works that way that seems like a good reason to go that way.

I'm just very aware that even with the best intentions once a workaround is in place it often ends up just staying that way (priorities...) and then we'd be left with multiple approaches would could be confusing

I also noticed that the story here doesn't have the placeholder being used. Is there a way to mock this somehow?

src/email/templates/Verify/Verify.tsx
coldlink: As mentioned on chat. This is only temporary while we still use IDAPI to send the emails. When we move over to using gateway (or another email service) we can use one way or another on how text is replaced.

PR #783, opened 2021-08-02
Password Show/Hide Accessibility

oliverlloyd:
## What does this change?
Refactors the `PasswordInput` to make showing and hiding the password more accessible.

To achieve this I:

- Removed the feature where the icon only appears when the input has focus, this hides this feature from the user and was the source of a bug
- Changed from a `div` to a `button` for the toggle
- Put the icon inside the button, as a child, not alongside it to remove the need to control tabbing
- Put the icon after the text input so that we tab onto it afterwards - this seems to be the standard in other implementations so becomes expected behaviour
- Made sure the toggle action is fired on both `<space>` and `<enter>`


oliverlloyd:
[Refactor Change Password Page (S)](https://trello.com/c/pMa0bnYg/3275-refactor-change-password-page-s)

oliverlloyd:
[Refactor PasswordInput validation](https://trello.com/c/ozepMaJI/3221-refactor-passwordinput-validation)

coldlink (approved):

PR #784, opened 2021-08-03
Improve error logging and remove removeNoCache

QuarpT:
## What does this change and why?

- Improve error logging format by wrapping the objects passed in with `util.format`. This will pretty print objects which are currently logged as `[object]` in the logs. `util.format` is wrapped on individual logging parameters to avoid string placeholders from being accidentally parsed.
- Also allow error objects to be passed to warn and info since this is useful.
- remove removeNoCache for safety. No responses should be cached for Gateway in case engineers accidentally introduce personal information into state, and forget to remove `removeNoCache`. Right now `removeNoCache` doesn't seem to cache, since no cache headers are set.

### DIFF
--- a/src/server/lib/logger.ts
+++ b/src/server/lib/logger.ts
@@ -1,34 +1,47 @@
 import { Logger, LogLevel } from '@/server/models/Logger';
 import { createLogger, transports } from 'winston';
+import { format } from 'util';
 
 const winstonLogger = createLogger({
   transports: [new transports.Console()],
 });
 
 export const logger: Logger = {
-  log(level: LogLevel, message: string) {
-    winstonLogger.log(level, message);
-  },
-  // errors can be anything
   // eslint-disable-next-line
-  error(message: string, error?: any) {
-    if (error && error.stack && typeof error.message === 'string') {
-      return logger.log(
-        LogLevel.ERROR,
-        `${message} | ${error.message} | ${error.stack}`,
+  log(level: LogLevel, message: string, error?: any) {
+    if (
+      error &&
+      typeof error === 'object' &&
+      error.stack &&
+      typeof error.message === 'string'
+    ) {
+      return winstonLogger.log(
+        level,
+        `${format(message)} - ${format(error.message)} - ${format(
+          error.stack,
+        )}`,
       );
     }
 
     if (error) {
-      return logger.log(LogLevel.ERROR, `${message} | ${error}`);
+      return winstonLogger.log(level, `${format(message)} - ${format(error)}`);
     }
 
-    return logger.log(LogLevel.ERROR, message);
+    return winstonLogger.log(level, `${format(message)}`);
   },
-  warn(message: string) {
-    logger.log(LogLevel.INFO, message);
+
+  // eslint-disable-next-line
+  error(message: string, error?: any) {
+    return logger.log(LogLevel.ERROR, message, error);
   },
-  info(message: string) {
-    logger.log(LogLevel.INFO, message);
+
+  // eslint-disable-next-line
+  warn(message: string, error?: any) {
+    return logger.log(LogLevel.WARN, message, error);
+  },
+
+  // eslint-disable-next-line
+  info(message: string, error?: any) {
+    logger.log(LogLevel.INFO, message, error);
   },
 };
--- a/src/server/lib/middleware/cache.ts
+++ b/src/server/lib/middleware/cache.ts
@@ -8,14 +8,3 @@ export const noCache = (_: Request, res: Response, next: NextFunction) => {
 
   next();
 };
-
-export const removeNoCache = (
-  _: Request,
-  res: Response,
-  next: NextFunction,
-) => {
-  res.removeHeader('Cache-Control');
-  res.removeHeader('Pragma');
-
-  next();
-};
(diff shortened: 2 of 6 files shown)


src/server/lib/logger.ts
```diff
@@ -1,34 +1,45 @@
 import { Logger, LogLevel } from '@/server/models/Logger';
 import { createLogger, transports } from 'winston';
+import * as util from 'util';
```
oliverlloyd, line 3:
```suggestion
import { format } from 'util';
```

Just for a mini win on performance?

src/server/lib/logger.ts
```diff
@@ -1,34 +1,45 @@
...
   // eslint-disable-next-line
-  error(message: string, error?: any) {
+  log(level: LogLevel, message: string, error?: any) {
     if (error && error.stack && typeof error.message === 'string') {
-      return logger.log(
-        LogLevel.ERROR,
-        `${message} | ${error.message} | ${error.stack}`,
+      return winstonLogger.log(
+        level,
+        `${util.format(message)} - ${util.format(
```
oliverlloyd, line 15:
```suggestion
        `${format(message)} - ${format(
```

src/server/lib/logger.ts
```diff
@@ -1,34 +1,45 @@
...
+          error.message,
+        )} - ${util.format(error.stack)}`,
```
oliverlloyd, line 17:
```suggestion
        )} - ${format(error.stack)}`,
```

src/server/lib/logger.ts
```diff
@@ -1,34 +1,45 @@
...
       );
     }
 
     if (error) {
-      return logger.log(LogLevel.ERROR, `${message} | ${error}`);
+      return winstonLogger.log(
+        level,
+        `${util.format(message)} - ${util.format(error)}`,
```
oliverlloyd, line 24:
```suggestion
        `${format(message)} - ${format(error)}`,
```

src/server/lib/logger.ts
```diff
@@ -1,34 +1,45 @@
...
+      );
     }
 
-    return logger.log(LogLevel.ERROR, message);
+    return winstonLogger.log(level, `${util.format(message)}`);
```
oliverlloyd, line 28:
```suggestion
    return winstonLogger.log(level, `${format(message)}`);
```

src/server/lib/logger.ts
```diff
@@ -1,34 +1,45 @@
...
 });
 
 export const logger: Logger = {
-  log(level: LogLevel, message: string) {
-    winstonLogger.log(level, message);
-  },
-  // errors can be anything
   // eslint-disable-next-line
-  error(message: string, error?: any) {
+  log(level: LogLevel, message: string, error?: any) {
```
oliverlloyd, line 11: We seem to be assuming `message` exists on `error`. I know this is outside this PR but maybe we can type it as `unknown` and see what that throws up?

src/server/lib/middleware/cache.ts
```diff
@@ -8,14 +8,3 @@ export const noCache = (_: Request, res: Response, next: NextFunction) => {
...
-export const removeNoCache = (
-  _: Request,
-  res: Response,
-  next: NextFunction,
-) => {
-  res.removeHeader('Cache-Control');
-  res.removeHeader('Pragma');
-
-  next();
-};
```
oliverlloyd, line 21: Now that the `removeNoCache` function is removed we can inline the remaining `noCache` function to the place where it's used (index.ts)?

src/server/lib/playSessionCookie.ts
```diff
@@ -18,7 +18,7 @@ const getCookieFromPlaySession: (req: Request) => any = (req) => {
       );
       return session.data;
     } catch (error) {
-      logger.warn(error);
+      logger.warn('getCookieFromPlaySession failed', error);
```
oliverlloyd, line 21: 👍

oliverlloyd (approved):
Nice!

Just a few comments but this is really great work, thanks for doing this!

coldlink:
Would it be useful to add `noCahce` to all routes, instead of individually?

This could be done in `express.ts` where global middleware is added.

src/server/lib/middleware/cache.ts
QuarpT: I've made a slightly different improvement, adding `server.use` to src/server/lib/middleware/index.ts and removing it from the individual routes src/server/routes/index.ts, following the middleware pattern in the project https://github.com/guardian/gateway/pull/784/commits/637936a44bddab138d7f1c0d8c88151db5c62d03

QuarpT:
> Would it be useful to add `noCahce` to all routes, instead of individually?
> 
> This could be done in `express.ts` where global middleware is added.

I've added it as middleware here https://github.com/guardian/gateway/pull/784/commits/637936a44bddab138d7f1c0d8c88151db5c62d03

src/server/lib/logger.ts
QuarpT: I've made an improvement here to ensure error is a non null object before looking at its properties. Using `any` is necessary because JavaScript can throw `any`, and we want to be able to log all possible unexpected errors thrown by libraries which we don't control. https://github.com/guardian/gateway/pull/784/commits/381d8730143f13171dcf2b5011ef20c8ddde40c2

QuarpT:
I've removed some of the global noCache refactoring changes since they were caused unexpected behaviour in staging

PR #787, opened 2021-08-04
Run Cypress End to End

oliverlloyd:
## What does this change?
This PR adds an end to end integration test using Cypress and MailSlurp.

### No mocks
When Cypress is running in `ete` mode it does not mock idapi; all requests are sent to the `code` environment.

### MailSlurp
This is a third party testing service that we are trialling. It allows the following functions:

`emptyInbox` - we use this to clear the inbox we are using at the start of any test so that we don't read stale emails.
`createInbox` - useful for new testing registrations
`waitForLatestEmail` - this is the function we use to read the email from the inbox that we hope was sent

## Reset password
We now test this end to end flow using live endpoints. We still have the existing mocked tests under `cypress-mocked` but this new, live test gives us more confidence.

## Why are the tests failing?
These tests will fail until we have AWS keys set as Github Secrets


oliverlloyd:
[End to end test released](https://trello.com/c/txLG5ccY/3252-end-to-end-test-released)

cypress/integration/ete/reset_password.spec.js
```diff
@@ -0,0 +1,36 @@
...
+    const existing = {
+      email: '0298a96c-1028-4e3d-b943-a2b478c84dbd@mailslurp.com',
+      inbox: '0298a96c-1028-4e3d-b943-a2b478c84dbd',
+    };
+
+    it("changes the reader's password", () => {
+      cy.emptyInbox(existing.inbox).then(() => {
+        cy.intercept({
+          method: 'GET',
+          url: 'https://api.pwnedpasswords.com/range/*',
```
ob6160, line 15: general q about behaviour of this api endpoint in the app for my benefit -- is it used to stop people from changing their password to a compromised one?

ob6160: and-if it is-do we want a test case for when this happens?

ob6160 (approved):

cypress/integration/ete/reset_password.spec.js
oliverlloyd: > general q about behaviour of this api endpoint in the app for my benefit -- is it used to stop people from changing their password to a compromised one?

Yes, exactly that. You can play around with this feature (and all of our pages and components) here: https://www.chromatic.com/component?appId=60929d168594f80039336501&name=Pages%2FChangePassword&buildNumber=631

oliverlloyd: > and-if it is-do we want a test case for when this happens?

Excellent point. This is defo something we want to be testing for, but actually [we already do](https://github.com/guardian/gateway/blob/main/cypress/integration/mocked/change_password.spec.js#L103)

.github/workflows/cypress-ete.yml
```diff
@@ -0,0 +1,35 @@
...
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+      - name: Cypress run
+        uses: cypress-io/github-action@v2
+        env:
+          PORT: 8861
+          NODE_ENV: development
+          ENVIRONMENT: production
+          STAGE: production
```
coldlink, line 16: stage in this case would be what the guardian use

so either `DEV`, `CODE`, or `PROD`

think in this case you want `CODE`

oliverlloyd: You have hawk eyesight!

coldlink: Yeah that's right, `CODE` is pretty much what we call our staging/manual testing environments. In the `yml` file you can see what CODE urls look like, generally in the form `*.code.dev-theguardian.com`

cypress/integration/ete/reset_password.spec.js
oliverlloyd: Yeah, Chromatic is fantastic. It's caught a _lot_ of issues, especially over on DCR.

Whenever you add a story it is automatically captured by Chromatic and then visual regression gets run on each push

.github/workflows/cypress-ete.yml
oliverlloyd: `CODE` is the name for the test/stage/pre prod/qa environment. It is all of those rolled into one. Not sure why it's called 'CODE' - I think there's a story there

See: https://m.code.dev-theguardian.com/uk

cypress/integration/ete/reset_password.spec.js
ob6160: visual regression checking is super neat, does a regression fail the build or is it mainly there for a quick check?

.github/workflows/cypress-ete.yml
oliverlloyd: > I'm going to be v annoying and ask loads of questions while I'm still new

ALWAYS ask questions - never annoying

cypress/integration/ete/reset_password.spec.js
oliverlloyd: It fails the build if it finds diffs. Someone has to click through to Chromatic and manually approve them (or fix the bug which caused them)

ob6160: nice – sounds like a good setup :)

.github/workflows/cypress-ete.yml
ob6160: lovely, thanks guys

PR #791, opened 2021-08-09
Set password/welcome page

coldlink:
## What does this change?
This PR adds the new "Welcome" page which will be used to allow a user to set a password after registering through the guest user flow without a password.

A companion PR in Identity is also related, which adds in the new welcome email template, and changes to routes in Identity API: https://github.com/guardian/identity/pull/1932

And a companion PR in Identity-Platform is used to set the new routes in the VCL configuration: https://github.com/guardian/identity-platform/pull/429

A user will access this welcome page by going through the new account verification email. This email links them to a page on Gateway in the format `profile.theguardian.com/welcome/some-token-here`.

The token that is used is the same as the one used in the reset/change password flow, since this token matches our criteria - a) a user can only access this token via email, and b) this token can be used to set a password.

## Other changes
- Updates the copy on the Link Expired page to make it more generic, as per UX request
- Sets up link expired page to use with this new welcome page too, if the token is invalid/expired
- Sets up a "resend welcome email" page, which is displayed if the users choses to resend the verification email after the link had expired/invalidated
- Move the business login/controller for the password token set and check into a single controller, so that the logic can be shared between the welcome page, and the existing reset/change password page.
- Update the step counter on the onboarding flow to include mention of "set password"
- Minor refactorings around route names to make them more generic
- Moves email cookie logic into it's own lib file, and exposes a method which can be used to read the email from both GU_email and PLAY_SESSION_2 cookies
- Update email `Footer` to take a `mistakeParagraphComponent` prop, to customise the text or show and hide the "received this email by mistake" paragraph

#### Link Expired
link_expired (Small)

#### Email Sent
email_sent (Small)

#### Welcome/Set Password Page
welcome (Small)

## Tested
- [x] CI
- [x] CI Cypress
- [x] CODE


src/client/components/ConsentsSubHeader.tsx
```diff
@@ -66,15 +66,29 @@ const circle = `
...
+  padding-top: ${space[6] + space[2]}px;
   &.active {
-    ${textSans.small({ fontWeight: 'bold' })}
+    ${textSans.xxsmall({ fontWeight: 'bold' })}
+    ${from.mobileMedium} {
+      ${textSans.xsmall({ fontWeight: 'bold' })}
+    }
+    ${from.phablet} {
+      ${textSans.small({ fontWeight: 'bold' })}
+    }
```
coldlink, line 89: Fixes issue with overlapping text on smaller screen sizes with the onboarding flow step counter

src/client/components/PasswordForm.tsx
```diff
@@ -228,7 +237,7 @@ export const PasswordForm = ({ submitUrl, fieldErrors }: Props) => {
 
   return (
     <form
-      css={form}
+      css={[form, gridAutoRow && gridAutoRow(passwordFormSpanDef)]}
```
coldlink, line 240: if this password form is on a page which uses a css grid layout, like the consents flow, calling `gridAutoRow` sets up the css grid correctly for this form element

src/client/layouts/ConsentsLayout.tsx
```diff
@@ -117,7 +119,7 @@ export const ConsentsLayout: FunctionComponent<ConsentsLayoutProps> = ({
             </div>
             <ConsentsBlueBackground>
               <div css={[gridItem(gridItemColumnConsents), controls]}>
-                {!error && (
+                {!error && !hideContinueButton && (
```
coldlink, line 122: Used to hide the `Save and continue` button on the `ConsentsLayout`, as we don't want this button to show on the "Welcome"/"Set password" page

src/client/pages/ResendPasswordPage.tsx
```diff
@@ -10,9 +10,9 @@ export const ResendPasswordPage = () => {
   return (
     <ResetPassword
       email={email}
-      headerText="Forgotten password link expired"
-      bodyText="Oh no! The link has expired. Enter your email address and we'll send you another link to change or set your password."
-      buttonText="Send me another"
+      headerText="Link expired"
+      bodyText="The link you are using has expired. Please enter your email address and we'll send you a new one."
+      buttonText="Send me a link"
```
coldlink, line 15: New Link expired page copy from UX

src/email/components/Footer.tsx
```diff
@@ -14,11 +16,13 @@ export const Footer = () => (
...
-          won&apos;t be subscribed if you don&apos;t click the confirmation
-          button above.
-        </p>
+        {mistakeText && (
+          <p>
+            If you received this email by mistake, simply delete it. You
+            won&apos;t be {mistakeText} if you don&apos;t click the confirmation
+            button above.
+          </p>
+        )}
```
oliverlloyd, line 25: There's a bit of a grammar coupling here (as in the two pieces of text are strongly coupled because the grammar has to flow). Maybe we could make `mistakeText` the whole paragraph? Or perhaps have a `source` or `context` prop to enable this dynamic text?

oliverlloyd (commented):
There are a lot of changes here so I wasn't able to follow the logic all the way through to have confidence I understood everything. That said, it looked great! I've just made some comments about some small things and then one big rambling comment on abstraction!

oliverlloyd:
Just looking at those diffs, is that shade of blue in the margins right for [this page](https://www.chromatic.com/test?appId=60929d168594f80039336501&id=611a7bbcb15a87003a97ed4f)?

src/client/components/ConsentsSubHeader.tsx
```diff
@@ -66,15 +66,29 @@ const circle = `
...
+  }
+  ${from.phablet} {
+    ${textSans.small()}
+  }
   position: relative;
   width: ${100 / numPages}%;
   display: flex;
   flex-direction: column;
-  justify-content: flex-end;
+  justify-content: flex-flex-start;
```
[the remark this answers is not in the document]
coldlink: nice spot!

src/email/components/Footer.tsx
coldlink: Actually, I think passing in the whole paragraph may be a better suggestion, as there's some cases where the text will not read correctly, so being able to customise the whole text is a better suggestion. 👍

cypress/integration/mocked/change_password.spec.js
```diff
@@ -60,10 +60,7 @@ describe('Password change flow', () => {
     it('Has no detectable a11y violations on change password complete page', () => {
       cy.mockNext(200);
       cy.mockNext(200, fakeSuccessResponse);
-      cy.intercept({
-        method: 'GET',
-        url: 'https://api.pwnedpasswords.com/range/*',
-      }).as('breachCheck');
+      cy.interceptPasswordBreachCheck();
```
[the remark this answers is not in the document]
coldlink: Ah interesting! I actually found things a bit harder to read, as I have to do a mental check (in this case) as to what `cy.intercept` does each time, so I found putting into a command/method easier to understand `interceptPasswordBreachCheck`!

But I'm happy to revert my changes to this.

cypress/integration/mocked/welcome.spec.js
```diff
@@ -0,0 +1,215 @@
...
+        errors: [
+          {
+            message: 'Breached password',
+          },
+        ],
+      });
+      cy.visit(`/welcome/fake_token`);
+      cy.get('input[name="password"]').type('thisisalongandunbreachedpassword');
+      cy.get('button[type="submit"]').click();
+      cy.contains('This is a common password.');
```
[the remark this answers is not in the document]
coldlink: Ah good spot! I'll change this around too.

coldlink:
> Just looking at those diffs, is that shade of blue in the margins right for [this page](https://www.chromatic.com/test?appId=60929d168594f80039336501&id=611a7bbcb15a87003a97ed4f)?

Oh interesting, that definitely should not be the case! Weird as all I've done is use the same consents layout. I'll take a look.

src/email/components/Footer.tsx
coldlink: Renamed it and turned it into an optional component that can be passed in.

webpack.development.js
```diff
@@ -10,6 +10,7 @@ const { mergeWithRules } = require('webpack-merge');
  * ForkTsCheckerWebpackPlugin provides type checking on top of webpack watch.
  */
 const common = {
+  devtool: false,
```
coldlink, line 13: fixes `unsafe-eval` errors in development, causing js to break

src/client/components/ConsentsSubHeader.tsx
```diff
@@ -66,15 +66,29 @@ const circle = `
...
+  ${from.phablet} {
+    ${textSans.small()}
+  }
   position: relative;
   width: ${100 / numPages}%;
   display: flex;
   flex-direction: column;
-  justify-content: flex-end;
+  justify-content: flex-start;
+  padding-top: ${space[6] + space[2]}px;
```
ob6160, line 81: just wondering if there's single space value for this instead of having to add them together?

src/client/components/PasswordForm.stories.tsx
```diff
@@ -8,5 +8,11 @@ export default {
   component: PasswordForm,
 } as Meta;
 
-export const Default = () => <PasswordForm submitUrl="" fieldErrors={[]} />;
+export const Default = () => (
+  <PasswordForm
```
ob6160, line 12: just thinking having an extra export with an example of a field error might be nice

src/client/components/PasswordForm.tsx
```diff
@@ -190,7 +194,12 @@ const isBreached = AwesomeDebouncePromise(
   300,
 );
 
-export const PasswordForm = ({ submitUrl, fieldErrors }: Props) => {
+export const PasswordForm = ({
+  submitUrl,
+  fieldErrors,
+  submitButtonText,
+  gridAutoRow,
+}: Props) => {
```
ob6160, line 202: small question on the naming of this Props type - I've seen a pattern where interfaces for functional components are called `IComponentNameProps`, e.g: (`IPasswordFormProps`), might be a bit more descriptive, especially if it's exported at some point. I realise that wasn't added as part of this, but maybe worth considering for our components?

src/client/layouts/ConsentsLayout.tsx
ob6160, line 122: just a thought - maybe is this was called `showContinueButton` it might help with readability as we wouldn't have the cognitive load of the not operator

src/server/controllers/changePassword.ts
```diff
@@ -0,0 +1,131 @@
...
+  validate as validateToken,
+  change as changePassword,
+} from '@/server/lib/idapi/changePassword';
+import { FieldError } from '@/shared/model/ClientState';
+import { ChangePasswordErrors } from '@/shared/model/Errors';
+import { setIDAPICookies } from '@/server/lib/cookies';
+import { trackMetric } from '@/server/lib/trackMetric';
+import { Metrics } from '@/server/models/Metrics';
+
+const validatePasswordField = (password: string): Array<FieldError> => {
```
ob6160, line 19: do we also have any disallowed characters in our passwords?

src/server/controllers/changePassword.ts
```diff
@@ -0,0 +1,131 @@
+import { Request } from 'express';
```
ob6160, line 1: general q - are there unit tests for the controllers in the express app or do we rely on the cypress suite for these?

ob6160 (commented):
really good work 🙂 difficult to follow in places without having full context. few small comments mainly on the client side. was thinking it might be nice to have a quick PR review call where we can run through the change on a running instance of the application too?

src/client/components/ConsentsSubHeader.tsx
coldlink: Unfortunately not: http://theguardian.design/2a1e5182b/p/449bd5-space

src/client/components/PasswordForm.tsx
coldlink: There isn't a hard rule in gateway as to how we name prop interfaces, in some components you'll notice `interface ComponentProps` in others it's just `interface Props`.

My general opinion is that because props are self-contained within the component/component file, we don't need to explicitly name them, as it's easy to tell which component those props belong too.

As for prefixing interfaces with `I` there is a big debate around these, and the general consensus is that the prefix isn't necessary: https://github.com/microsoft/TypeScript-Handbook/issues/121#issuecomment-412810455

src/client/layouts/ConsentsLayout.tsx
coldlink: Good shout, I went with this as by default we want it to show rather than hide. But I've refactored it after this suggestion to `showContinueButton` and setting the default value to true, so you have to explicitly pass `showContinueButton={false}` to hide it.

src/server/controllers/changePassword.ts
coldlink: We rely on Cypress to integration test these, as these controllers/routes are only accessed through a web browser.

src/server/controllers/changePassword.ts
coldlink: The short answer is no.
The long answer is all the validation is done in IDAPI for passwords on submit, however IDAPI only checks the length too. As long as we detect the password hasn't been breached, and is standard unicode we allow it. The checks we do clientside are really only sanity checks.

src/client/layouts/ConsentsLayout.tsx
ob6160: nice :)

src/client/components/PasswordForm.tsx
ob6160: makes sense, thanks for explaining rationale. not seen that thread before will take a look

src/client/components/ConsentsSubHeader.tsx
ob6160: hm, if we have to make a space value that doesn't exist on the scale is that circumventing the design system? not sure how strict we want to be at sticking to their scale. I realise this is a tiny issue but thought worth bringing up :)

src/server/controllers/changePassword.ts
ob6160: nice one 👍

webpack.development.js
ob6160: haha I didn't notice this. just shipped an alternative fix: https://github.com/guardian/gateway/pull/837/commits/13599eac741b2be59e807fb10e1940be2dda17d6, letting the script eval only on dev envs. What do you think about that method? That way we can keep the source maps that I think devtool gives us.

src/client/components/PasswordForm.tsx
ob6160: nice 👍

src/client/components/ConsentsSubHeader.tsx
coldlink: Yeah, in general we try to stick to the guidelines, in this case it was a bit different as rather than padding between elements, we had to pad in the element to push the text down from the step counter line:

Inkedwelcome_LI

webpack.development.js
coldlink: ah nice! thats a good approach, i'll also comment on that PR, but a slightly safer way of reading from `process.env` in gateway is to read from the `getConfiguration` method provided in `server/lib/configuration.ts`

oliverlloyd:
With this new page, I presume the Set/Change Password is now just Change Password? 

https://www.chromatic.com/test?appId=60929d168594f80039336501&id=611cf4cd93672d003a6cf520

In which case, maybe the copy could be updated there to be more changy?

oliverlloyd:
Is there a use case for someone entering the onboarding flow without the set password step? Do we need to capture this?

coldlink:
> Is there a use case for someone entering the onboarding flow without the set password step? Do we need to capture this?

Depending on when they set the password there is a case where they can land on the onboarding flow without the set password step, but I don't think we need to capture this, as the step counter already reflects this, when they land on the onboarding flow "Stay in touch" step, the "Create password" will already be checked.

coldlink:
> With this new page, I presume the Set/Change Password is now just Change Password?
> 
> https://www.chromatic.com/test?appId=60929d168594f80039336501&id=611cf4cd93672d003a6cf520
> 
> In which case, maybe the copy could be updated there to be more changy?

Can just change it to "Change password" rather than "Set password" on that page?

oliverlloyd:
> Can just change it to "Change password" rather than "Set password" on that page? This is probably something we can do in a separate PR

Yeah, that copy change might be all we need. And possibly a url edit - not sure what that page url is at the moment.

👍 for doing this in a new PR

oliverlloyd:
> > Is there a use case for someone entering the onboarding flow without the set password step? Do we need to capture this?
> 
> Depending on when they set the password there is a case where they can land on the onboarding flow without the set password step, but I don't think we need to capture this, as the step counter already reflects this, when they land on the onboarding flow "Stay in touch" step, the "Create password" will already be checked, as they'd have already set a password previously.

I agree, it's much nicer to keep things simpler here and just show the password step as complete. Arguably that's more informative and perhaps even encourages more completion of onboarding! (Greater sense of accomplishment)

cypress/integration/mocked/welcome.spec.js
```diff
@@ -0,0 +1,221 @@
...
+        checkTokenSuccessResponse.user.primaryEmailAddress,
+      );
+      cy.get('button[type="submit"]').click();
+
+      cy.contains('Change email address').click();
+
+      cy.contains('Link expired');
+    });
+  });
+});
```
oliverlloyd, line 221: Really readable tests!

💯

src/client/pages/Welcome.tsx
```diff
@@ -16,21 +17,25 @@ type Props = {
...
-    </>
+    <ConsentsLayout
+      title="Welcome to the Guardian"
+      current={CONSENTS_PAGES.PASSWORD}
+      bgColor={CONSENTS_MAIN_COLOR}
+      showContinueButton={false}
+    >
+      <ConsentsContent>
+        <p css={[text, autoRow()]}>
+          Please create a password for {email || 'your new account'}
```
oliverlloyd, line 30: This reads really nice but isn't email required so should always be present? (Not a blocker!)

src/server/controllers/changePassword.ts
```diff
@@ -0,0 +1,131 @@
...
+      trackMetric(Metrics.ACCOUNT_VERIFICATION_SUCCESS);
+      trackMetric(Metrics.CHANGE_PASSWORD_SUCCESS);
+
+      return successCallback(res);
+    } catch (error) {
+      const { message, status } = error;
+      logger.error(`${req.method} ${req.originalUrl}  Error`, error);
+
+      trackMetric(Metrics.ACCOUNT_VERIFICATION_FAILURE);
+      trackMetric(Metrics.CHANGE_PASSWORD_FAILURE);
```
oliverlloyd, line 119: It feels a bit iffy to have two success and failure metrics for one event. I get that there are two things happening here, but still, writing two events to the log might cause confusion later on?

oliverlloyd: Is this perhaps a naming problem instead?

src/server/controllers/changePassword.ts
```diff
@@ -0,0 +1,131 @@
...
+      return res.type('html').send(
+        renderer(`${resendEmailPagePath}${Routes.RESEND}`, {
+          requestState: state,
+          pageTitle: resendEmailPageTitle,
+        }),
+      );
+    }
+  });
+
+export const setPasswordTokenController = (
```
oliverlloyd, line 72: Would it make sense to have two controllers, one for setting, one for changing? Or maybe we should change the name of this one from `changePassword` to...? Just a thought

src/server/lib/cookies.ts
```diff
@@ -0,0 +1,65 @@
...
+}
+
+interface IdapiCookies {
+  values: Array<IdapiCookie>;
+  expiresAt: string;
+}
+
+const { baseUri } = getConfiguration();
+
+export const setIDAPICookies = (res: Response, cookies: IdapiCookies) => {
```
oliverlloyd, line 18: Personally, I generally prefer one export per file - I find it easier to search and read code like this (mainly because it works better with VSCode, I can CMD+P and go straight to the definition) but also because having the scope of that one function in one file makes it easier to reason about.

I totally get that this is subjective though and that we're straying close to tab vs. spaces territory here!

src/server/routes/index.ts
```diff
@@ -41,6 +42,9 @@ uncachedRoutes.use(verifyEmail);
 // magic link routes
 uncachedRoutes.use(magicLink);
```
oliverlloyd, line 43: Unrelated, but should this route also use `noCache`?

oliverlloyd (approved):
Really great work!

I added some comments on things but nothing blocking

src/client/pages/Welcome.tsx
coldlink: Actually I'll rethink the logic around this! I think email should be optional, as there may be cases where a user accidently lands on the page directly and the email is no longer in the cookie!

src/server/controllers/changePassword.ts
coldlink: Yeah, it's because the change password endpoint in IDAPI is now doing two things, a) ACCOUNT_VERIFICATION and b) CHANGE_PASSWORD which we'd want to track separately.
I think I'll add a comment to the code to explain why there's two things going in.

src/server/controllers/changePassword.ts
coldlink: Yeah this is a naming problem again, we use "set" and "change" interchangeably as in our case they have exactly the same logic.
Personally I feel that "set" makes more sense than "change" as whether your creating (set) a password for the first time or resetting your password where you "set" a new password.
So maybe the codebase should be renamed throughout?
It's difficult though. Naming things is hard.

src/server/lib/cookies.ts
coldlink: Ah interesting.
I'm a big fan of CTRL/CMD+T and searching method names over using CMD+P to find filenames, or CTRL/CMD+Click on a method to go to it's definition.

src/server/routes/index.ts
coldlink: it does already! `uncachedRoutes` is a wrapper which Peter created in #788 to automatically apply `noCache`

src/client/pages/Welcome.tsx
oliverlloyd: Good point!

src/server/lib/cookies.ts
oliverlloyd: Fairs

src/server/controllers/changePassword.ts
oliverlloyd: 'set' defo seems more generic, which might be beneficial here. Defo one for a later PR - I don't want to hold this one up

src/server/lib/cookies.ts
coldlink: I've actually just refactored this into it's own file in this commit: https://github.com/guardian/gateway/pull/791/commits/d6f7e7859fbfa9b0716f41db6db9a40e3f6e37f3

After a discussion just now with @ob6160 who's going to be sending an email to gateway using the `PLAY_SESSION_2` cookie, it made sense to have a single method in that new file which first checked the `GU_email` cookie for the email, followed by the `PLAY_SESSION_2` cookie in a case where the user was coming from identity-frontend, e.g. from the registration page to the email sent page.

PR #811, opened 2021-08-10
Fix isLegacy not defined error in webpack.development.js

coldlink:
## What does this change?
I got a `ReferenceError: isLegacy is not defined` when running `make dev`/`yarn dev` because `isLegacy` was not in scope in `webpack.development.js`.

This PR wraps the object in a function which takes `isLegacy` as a parameter to solve this.

In some cases it appeared to run the spread operator on `isLegacy` itself rather than the outcome of the ternary operator, wrapping this in brackets solves this.

### DIFF
--- a/webpack.development.js
+++ b/webpack.development.js
@@ -36,7 +36,7 @@ const serverConfig = {
   }
 }
 
-const browserConfig = {
+const browserConfig = (isLegacy = false) => ({
   module: {
     rules: [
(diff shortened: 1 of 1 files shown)


ob6160:
thanks for fixing - missed this when I removed the variable in error during my tidyup of this file

ob6160 (approved):

PR #812, opened 2021-08-10
Ollie/tweak stmux

ob6160:
Some tweaks to the dev commands in the makefile so that we default to running as before and make the tiled view an optional extra.


package.json
```diff
@@ -19,7 +19,6 @@
     "tsc": "tsc --noEmit --skipLibCheck",
     "watch:server": "nodemon --inspect ./build/server.js",
     "watch": "webpack --watch --config webpack.development.js",
-    "dev": "stmux -M -e ERROR -m beep -- [ [ \"yarn watch\" .. \"yarn watch:server\" ] ]",
```
coldlink, line 22: removing `yarn dev` would mean the `make dev` command wouldn't work

either `yarn dev`  or `make dev` should become `yarn watch:server & yarn watch & wait`

coldlink (approved):

package.json
ob6160: oh hm, I thought I had changed that - thanks for catching

ob6160: fixed 👍

makefile
```diff
@@ -25,7 +25,17 @@ build: clear clean-build install
 dev: clean-build clear install
 	$(call banner, "gateway is starting")
 	$(call log, "starting development server")
-	@yarn dev
+	@(set -a && source .env && yarn watch:server & yarn watch & wait)
```
coldlink, line 28: Ah! Another small change, we don't need the `set -a && source .env` anymore, as we have a webpack plugin which does that for us now. 

https://github.com/guardian/gateway/blob/60459676fd596288e48dad583a9487b6b4f754a1/webpack.config.js#L63

```suggestion
	@(yarn watch:server & yarn watch & wait)
```

ob6160: ah just removed this :)

PR #834, opened 2021-08-16
16th August 2021 | Yet more dependency updates

coldlink:
## What does this change?

## Notes
- `@babel/preset-react` wasn't updated as this causes an issue with visual regression in storybook where it removes all styles in storybook

## Tested?
- [x] Local
- [x] CI
- [ ] CODE


QuarpT (approved):

ob6160:
weird about the storybook thing - might look into it at some point

oliverlloyd (approved):

coldlink:
> weird about the storybook thing - might look into it at some point

Yeah it's really weird, I haven't dug into it deeply yet, there's nothing in the `@babel/preset-react` changelogs which would cause this to happen, assuming it's a peer dependency somewhere. Another very good candidate for 10% time

PR #835, opened 2021-08-17
Header refactor

oliverlloyd:
## What does this change?
This PR refactors `Header` to reduce the size of the logo.

I took the chance here to consider the [larger use case for a Header at the Guardian](https://github.com/guardian/source/discussions/816) and changed things with this in mind. Perhaps a follow on PR from this could be to hoist the `Logo` component into Source? I looked at how Header and Logo are implemented in DCR and made sure to consider their use cases as well so hopefully any Source version will be useful in other places.

## `logoType`
Allows the version of the logo to be specified
```type LogoType = 'standard' | 'anniversary' | 'bestWebsite';```

See: https://github.com/guardian/source/pull/935 for more details

https://user-images.githubusercontent.com/1336821/130223316-d660384f-c9a4-49c0-beb9-8829c6a1f962.mov

### CODE REVIEW

File: src/client/components/Logo.tsx [hunk 1]
```diff
@@ -0,0 +1,87 @@
...
+import { Link } from '@guardian/src-link';
+import { from } from '@guardian/src-foundations/mq';
+import { visuallyHidden } from '@guardian/src-foundations/accessibility';
+
+const svgSize = css`
+  svg {
+    width: 149px;
+    height: 50px;
+    ${from.tablet} {
+      width: 200px;
```
ob6160, line 15: how are these sizes at each breakpoint level decided? are they done by eye or set out in DCR?

src/client/components/Logo.tsx
```diff
@@ -0,0 +1,87 @@
...
+
+export const Logo = ({ isAnniversary }: { isAnniversary?: boolean }) => (
+  <Link
+    href="https://www.theguardian.com"
+    title="The Guardian Homepage"
+    subdued={true}
+    cssOverrides={css`
+      color: ${brandText.primary};
+      :hover {
+        color: ${brandText.primary};
```
ob6160, line 74: just wondering why have this hover selector if it doesn't change the colour

ob6160 (commented):
looks good :) just a few comments

src/client/components/Logo.tsx
oliverlloyd: Yeah, that's a really good point. We should align with DCR, let me look into doing that.

src/client/components/Logo.tsx
oliverlloyd: This is there to override the hover value that you get when using `Link`

ob6160: ah yes, missed that :)

src/client/components/Logo.tsx
ob6160: cool cool

oliverlloyd: This is now done. We should be using the same exact sizes as DCR

ob6160 (approved):
🥇 nice one

src/client/components/Logo.tsx
ob6160: brill :)

oliverlloyd:
[Forgotten password: Reduce logo size and margins to match dotcom](https://trello.com/c/IGr9x5Td/3289-forgotten-password-reduce-logo-size-and-margins-to-match-dotcom)

PR #837, opened 2021-08-18
Compile the server using esbuild.

ob6160:
## What does this change?

* Builds the server using [esbuild](https://esbuild.github.io/) instead of webpack. This results in an extremely fast build time whilst retaining feature parity with the webpack server build. Should also fix a memory leak problem mentioned by @oliverlloyd when running `make dev`
* Splits out build and watch commands to be more granular so it is clearer what each does:
  * `build:client`, builds the client side for production.
  * `build:client:dev`, dev build of the client side (using dev webpack config to take advantage of caching for speed)
  * `watch:client`, runs webpack in watch mode
  * `build:server`, builds the server for production
  * `watch:server`, rebuilds the server on file change and restarts the server when rebuilt.
    * The `WATCH` env variable will start esbuild in watch mode. 
* `yarn watch` and `yarn build` then compose the above commands and functionally remain the same as before.
* Add some detail to the error thrown by `getAssets.ts` asking if the client has been built.
* Fixes hydration on dev build. It seems like it was already broken on dev before this PR, so I've let the eval through only on dev environments in `helmet.ts`

### New server build time:
Screenshot 2021-08-18 at 18 41 55

### Old server build time:
Screenshot 2021-08-18 at 18 52 16


esbuild.server.js
```diff
@@ -0,0 +1,36 @@
...
+
+// eslint-disable-next-line @typescript-eslint/no-var-requires
+require('esbuild').build({
+  entryPoints: ['./src/server/index.ts'],
+  bundle: true,
+  minify: true,
+  platform: 'node',
+  external: ['fsevents', 'esbuild'],
+  outfile: 'build/server.js',
+  target: 'node10.4',
```
coldlink, line 23: This should be node 14 lts, as that's what we're running gateway on

src/server/lib/middleware/helmet.ts
```diff
@@ -35,6 +36,9 @@ const helmetConfig = {
         gaUID.hash, // google analytics id
         CSP_VALID_URI.GOOGLE_ANALYTICS,
         CSP_VALID_URI.CMP,
+        ...(process.env.NODE_ENV === 'production'
```
coldlink, line 39: we generally use `getConfiguration()` and check for the guardian `stage` value and one of `DEV`, `CODE`, `PROD`, in this case we should not apply `UNSAFE_EVAL` in `DEV`

coldlink (commented):
Great work! A couple of small comments.

esbuild.server.js
oliverlloyd: Is there a dept. wide supported version of Node? It might be `14.16.1`?

It's ever so slightly annoying having to `nvm use` when switching projects - not a big deal though

oliverlloyd:
This is great stuff!

One thing, would it be possible to tidy up the commit history?

This code will live for years and years and we want out git history to be the best it can be to help future version of ourselves. I very much recommend watching this video on this subject: https://tekin.co.uk/2019/02/a-talk-about-revision-histories which explains _why_ this is a good idea. It also gives on a lot of detail on how to edit git history but keep in mind you can also use Github's Squash and Merge feature (assuming you're happy with a single commit). Personally I prefer smaller atomic messages but sometimes squashing is better.

ob6160:
> sh and Merge feature (assuming you're happy with a single commit). Personally I prefer smaller atomic messages but sometimes squashing is

Thanks! I'll give the video a watch for sure. I think using the squash feature is a great rule to follow - happy to use going forward. Maybe we could set is as the default merge method in the github config?

esbuild.server.js
ob6160: good catch

oliverlloyd:
> > sh and Merge feature (assuming you're happy with a single commit). Personally I prefer smaller atomic messages but sometimes squashing is
> 
> Thanks! I'll give the video a watch for sure. I think using the squash feature is a great rule to follow - happy to use going forward. Maybe we could set is as the default merge method in the github config?

I would caution against  making it the default. It's a bit like a kind of nuclear option that you can fallback on if you fail to create a more meaningful series of commit messages. I tend to use it when things have gone wrong!

In some situations, if I think the area I'm working on is important, then I will actually start from scratch so that I can commit the changes properly in order to create the best history. That combination of good commit messages, descriptions and great PR write ups is hugely useful!

esbuild.server.js
ob6160: @oliverlloyd it is frustrating, there is a shell script that nvm provide which looks for an `nvmrc` and auto-`nvm use`s but I've found it can slow down my terminal a little

ob6160:
> Thanks



> > > sh and Merge feature (assuming you're happy with a single commit). Personally I prefer smaller atomic messages but sometimes squashing is
> > 
> > 
> > Thanks! I'll give the video a watch for sure. I think using the squash feature is a great rule to follow - happy to use going forward. Maybe we could set is as the default merge method in the github config?
> 
> I would caution against making it the default. It's a bit like a kind of nuclear option that you can fallback on if you fail to create a more meaningful series of commit messages. I tend to use it when things have gone wrong!
> 
> In some situations, if I think the area I'm working on is important, then I will actually start from scratch so that I can commit the changes properly in order to create the best history. That combination of good commit messages, descriptions and great PR write ups is hugely useful!

Makes sense, I have seen a git precommit hook tool before that checks to see if commits are prefixed with `refactor:` `feat:` `fix:` etc, maybe we could look at getting that in place as a way to add a little linting to our commit messages? Am aware it is a bit restrictive though, thought I'd suggest anyhow.

oliverlloyd:
> I have seen a git precommit hook tool before that checks to see if commits are prefixed with `refactor:` `feat:` `fix:` etc

I guess semantic commit messages are a personal choice on Gateway. This isn't a _lib_ so we're not doing releases (which is the main benefit for semantic commits). I still like to use them as a I find it makes me think about what I'm doing but I'm not sure that pattern is worth enforcing with linting?

coldlink:
> > I have seen a git precommit hook tool before that checks to see if commits are prefixed with `refactor:` `feat:` `fix:` etc
> 
> I guess semantic commit messages are a personal choice on Gateway. This isn't a _lib_ so we're not doing releases (which is the main benefit for semantic commits). I still like to use them as a I find it makes me think about what I'm doing but I'm not sure that pattern is worth enforcing with linting?

Yeah I'm in agreement with @oliverlloyd here. Really helps me think through commits, and actually helps me break commits down into smaller components. Not everything needs to be standardised. Maybe something worth in the readme to say that we prefer one way over another but it's not enforced, for example what we say about merging in the readme already: https://github.com/guardian/gateway/blob/main/README.md#contributing

esbuild.server.js
coldlink: honestly hard coding it is probably fine

coldlink: there isn't really a standard version, the general advice is to use the latest LTS  version if possible

ob6160: I'll go ahead and set the target to `node14` if we're all in agreement? :)

src/server/lib/middleware/helmet.ts
ob6160: sorry I think I'm being a bit thick - do we want to include `UNSAFE_EVAL` if `process.env.NODE_ENV !== 'production' || getConfiguration().stage === 'DEV'

ob6160:
> > > I have seen a git precommit hook tool before that checks to see if commits are prefixed with `refactor:` `feat:` `fix:` etc
> > 
> > 
> > I guess semantic commit messages are a personal choice on Gateway. This isn't a _lib_ so we're not doing releases (which is the main benefit for semantic commits). I still like to use them as a I find it makes me think about what I'm doing but I'm not sure that pattern is worth enforcing with linting?
> 
> Yeah I'm in agreement with @oliverlloyd here. Really helps me think through commits, and actually helps me break commits down into smaller components. Not everything needs to be standardised though, people have different ways of working. Maybe something worth in the readme to say that we prefer one way over another but it's not enforced, for example what we say about merging in the readme already: https://github.com/guardian/gateway/blob/main/README.md#contributing

agree with you both - sounds good! Let's chat in tomorrow's engineering standup about maybe changing that contributing section? :)

esbuild.server.js
oliverlloyd: Or maybe we can use 14.16.1 just so we're aligned with DCR and friends?

ob6160: good suggestion, pushed that up

src/server/lib/middleware/helmet.ts
ob6160: ah - I did some experimentation and I've pushed what I think you meant :)

esbuild.server.js
```diff
@@ -0,0 +1,36 @@
+// eslint-disable-next-line @typescript-eslint/no-var-requires
```
oliverlloyd, line 1: Aside: Why is `typescript-eslint` interfering in `.js` files?

esbuild.server.js
```diff
@@ -0,0 +1,36 @@
...
+const dotenv = require('dotenv');
+
+// Pulls in .env and populates the define object.
+// This is used to populate environment variables at built time.
+dotenv.config();
+const define = {};
+for (const k in process.env) {
```
oliverlloyd, line 8: I'm of the school of thought that says variable names should be meaningful. I know you often see variables given one letter names to shorten the code and the argument is that's it's obvious here, but I think it's better to have `key` here. It just reads nicer and saves my brain at least the one tick it takes to work out what k stands for.

esbuild.server.js
```diff
@@ -0,0 +1,36 @@
...
+  // eslint-disable-next-line functional/immutable-data
+  define[`process.env.${k}`] = JSON.stringify(process.env[k]);
+}
+
+const watchMode = process.env['WATCH'];
```
oliverlloyd, line 13: If `watchMode` is an environment variable it should be in upper case, no? Not sure where this convention comes from - it might just be in my head! - but it feels like a thing

oliverlloyd: You could also just inline `process.env['WATCH']`. If you need to assign it to a variable to describe what it does that can be useful but maybe we could look at renaming the variable?

esbuild.server.js
```diff
@@ -0,0 +1,36 @@
...
+  target: 'node14.16.1',
+  loader: { '.png': 'dataurl', '.jpg': 'dataurl', '.js': 'jsx' },
+  jsxFactory: 'jsx',
+  inject: ['./react-shim.js'],
+  define,
+  watch: watchMode
+    ? {
+        onRebuild(error, result) {
+          if (error) console.error('watch build failed:', error);
+          else console.log('watch build succeeded:', result);
```
oliverlloyd, line 32: Do we need the success log?

package.json
```diff
@@ -5,7 +5,13 @@
...
   "private": true,
   "scripts": {
-    "build": "webpack --mode production && node scripts/copyNodeModules.js",
+    "build": "yarn build:client && yarn build:server",
+    "watch": "yarn watch:client & yarn watch:server",
+    "build:client": "webpack --mode production && node scripts/copyNodeModules.js",
+    "build:client:dev": "webpack --config webpack.development.js && node scripts/copyNodeModules.js",
+    "watch:client": "webpack --watch --config webpack.development.js",
+    "build:server": "node esbuild.server.js",
+    "watch:server": "WATCH=1 yarn build:server & nodemon --inspect build/server.js",
```
oliverlloyd, line 14: Is there a use case for these yarn scripts? (It's possible there is and I'm not seeing it) I find it's better to keep the list of scripts in the package file as small as possible so that new devs have a clearer idea of what commands they _should_ run.

My thinking is, if I shouldn't run this command as part of my dev routine, then it shouldn't be here.

react-shim.js
```diff
@@ -0,0 +1,2 @@
+import { jsx } from '@emotion/react';
+export { jsx };
```
oliverlloyd, line 2: This is a bit quirky. Is this file only needed for the esbuild? If so, is there a way we can indicate this - maybe with colocation, maybe with naming, maybe with a comment, etc. so that it's not misunderstood

src/server/lib/getAssets.ts
```diff
@@ -26,7 +26,7 @@ export const getAssets = (isLegacy = false): Assets => {
       isLegacy ? 'legacy.' : ''
     }webpack-assets.json`;
     if (!fs.existsSync(assetsFilePath)) {
-      throw new Error('Assets file does not exist');
+      throw new Error('Assets file does not exist. Have you built the client?');
```
oliverlloyd, line 29: Very nice touch!

src/server/lib/middleware/helmet.ts
```diff
@@ -1,12 +1,13 @@
...
 import { getConfiguration } from '@/server/lib/getConfiguration';
 
-const { baseUri, gaUID, apiDomain, idapiBaseUrl } = getConfiguration();
+const { baseUri, gaUID, apiDomain, idapiBaseUrl, stage } = getConfiguration();
 
 enum HELMET_OPTIONS {
   SELF = "'self'",
   NONE = "'none'",
   UNSAFE_INLINE = "'unsafe-inline'",
+  UNSAFE_EVAL = "'unsafe-eval'",
```
oliverlloyd, line 10:
```suggestion
  UNSAFE_EVAL = "'unsafe-eval'", // !!! ONLY USE FOR DEV !!!
```

src/server/lib/middleware/helmet.ts
```diff
@@ -35,6 +36,7 @@ const helmetConfig = {
         gaUID.hash, // google analytics id
         CSP_VALID_URI.GOOGLE_ANALYTICS,
         CSP_VALID_URI.CMP,
+        ...(stage === 'DEV' ? [HELMET_OPTIONS.UNSAFE_EVAL] : []),
```
oliverlloyd, line 39: This ternary spread operator combo is a bit narly and it's in an area that is arguably critically important. Shall we break this logic out so that we define the `scriptSrc` array above the `helmetConfig` object? That was we can use much simpler javascript for conditionally adding to the array Eg `if(stage === 'DEV') scriptSrc.push(HELMET_OPTIONS.UNSAFE_EVAL)` which is a lot easier to read

webpack.development.js
```diff
@@ -64,9 +45,8 @@ const browserConfig = (isLegacy = false) => ({
   ]
 });
 
-const baseServerConfig = baseConfig[0];
-const baseBrowserLegacyConfig = baseConfig[1];
-const baseBrowserConfig = baseConfig[2];
+const baseBrowserLegacyConfig = baseConfig[0];
+const baseBrowserConfig = baseConfig[1];
```
oliverlloyd, line 49: Maybe now this file is so much simpler we can rename these?

oliverlloyd (commented):
This is such a cool PR, really looking forward to the developing in light speed mode!

package.json
coldlink: Yeah, I agree. I think these can be just in-lined into the `dev` and `build` commands. I can't think of a reason as to when we'd have to run one of these commands separately

esbuild.server.js
ob6160: Good catch! I've done some digging and I think we'll need to make some tweaks to our eslintrc to only use `typescript-eslint` for ts/tsx files. I'm thinking it'd make a good followup PR? https://github.com/typescript-eslint/typescript-eslint/issues/109#issuecomment-536160947

esbuild.server.js
ob6160: definitely, thanks for picking up on that, will change

esbuild.server.js
ob6160: I think inlining it is a good idea, it's quite self explanatory as-is

esbuild.server.js
ob6160: yeah, not really needed, I'll cut it

package.json
ob6160: got a bit carried away with the scripts there, I'll narrow these down to dev and build as suggested

react-shim.js
ob6160: this is doing the job of `@emotion/babel-reset-css-prop` (docs: https://emotion.sh/docs/@emotion/babel-preset-css-prop#usage), it does the job of injecting an import from `@emotion/react` for `jsx` which is needed to enable support for the `css` prop. Here's a good thread where it's discussed: https://github.com/evanw/esbuild/issues/832#issuecomment-781497513

I think a clarifying comment here is an excellent suggestion

ob6160 (commented):
@oliverlloyd @coldlink Finally gotten around to reply to your review comments here. I got a bit distracted by MVP0 work :)

ob6160:
This was an interesting POC to work on, however I think best to close this PR for now because Esbuild isn't widely used by the community yet. The docs on production readiness: https://esbuild.github.io/faq/#production-readiness

PR #839, opened 2021-08-20
Email sent page for new registrations

ob6160:
## What does this change?
This adds an endpoint (`/register/email-sent`). The user will be redirected to it upon registering as a guest user in **identity-frontend** after this companion PR is merged: https://github.com/guardian/identity-frontend/pull/675

The endpoint renders the `EmailSentPage` component and passes it the users' email when rendering. Their email is read from the encrypted email stored in the _PLAY_SESSION_2_ cookie.

This also includes some cypress tests, checking for the correct behaviour of the email sent page in Gateway.

## Screenshot
Screenshot 2021-08-23 at 23 52 11


src/client/routes.tsx
```diff
@@ -37,6 +37,9 @@ export const GatewayRoutes = () => (
     <Route exact path={Routes.REGISTRATION}>
       <RegistrationPage />
     </Route>
+    <Route exact path={Routes.REGISTRATION_EMAIL_SENT}>
+      <EmailSentPage />
```
oliverlloyd, line 41: We could just pass in `subTitle` here? Unless I'm missing a use case (possible!)

oliverlloyd: Or just have `RegistrationSentPage` and then we can hardcode these dynamic values? (`subTitle`, `previousPage` and `pageTitle` - non of them really need to be set in state)

oliverlloyd (commented):
This looks great!

Two things:

1. I wonder if - now that we have more dynamic values for `EmailSent` - it's worth creating separate page components, like `RegistrationSentPage`?

2. Can we add a test for this flow? We've not been doing this up to now (for good reasons) but now that we have the option it'd be great to start including tests to validate new logic like this.

src/client/routes.tsx
ob6160: yes I like that idea to have a `RegistrationSentPage` :) I'm going to make a small change anyway to POST the resend email to a different endpoint to the "change email" link so I'll do that as part of this change

ob6160:
> This looks great!
> 
> Two things:
> 
> 1. I wonder if - now that we have more dynamic values for `EmailSent` - it's worth creating separate page components, like `RegistrationSentPage`?
> 2. Can we add a test for this flow? We've not been doing this up to now (for good reasons) but now that we have the option it'd be great to start including tests to validate new logic like this.

1. Agreed will create a separate page component
2. Yes! I'll add a test which loads this page with an encrypted email and checks for correct behaviour

oliverlloyd:
> 2\. Yes! I'll add a test which loads this page with an encrypted email and checks for correct behaviour

I'd be happy to pair on this!

cypress/integration/ete/register_email_sent.spec.js
```diff
@@ -0,0 +1,79 @@
...
+      log: true,
+    });
+    cy.visit(`/register/email-sent`);
+    cy.contains('Change email address').click();
+    cy.contains('Sign in');
+    cy.title().should('eq', 'Sign in | The Guardian');
+  });
+
+  it('should render properly if the encrypted email cookie is not set', () => {
+    cy.visit(`/register/email-sent`);
```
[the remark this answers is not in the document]
ob6160: nice spot, will add

cypress/integration/ete/register_email_sent.spec.js
```diff
@@ -0,0 +1,79 @@
...
+    cy.contains('Change email address');
+  });
+
+  // This test depends on this Mailslurp account already being sent to the /guest endpoint.
+  const resendTestEmail = {
+    email: '7b4f97c5-79cd-4a3c-9212-cfb6db959dfc@mailslurp.com',
+    inbox: '7b4f97c5-79cd-4a3c-9212-cfb6db959dfc',
+  };
+  // Set encrypted email using test email defined above.
+  const testMailSlurpEncryptedEmailCookie = Buffer.from(
```
oliverlloyd, line 40: It might be nice to name this something more descriptive, like `encryptedEmail`?

cypress/integration/ete/register_email_sent.spec.js
```diff
@@ -0,0 +1,79 @@
...
+      });
+      cy.visit(`/register/email-sent`);
+      injectAndCheckAxe();
+    });
+  });
+
+  const testUserEmail = 'test+13411@guardian.com';
+  const testUserEncryptedEmailCookie = Buffer.from(
+    JSON.stringify(testUserEmail),
+  ).toString('base64');
```
oliverlloyd, line 20: This is only used in the context of the text below so should live inside that `it` scope. This helps communicate intent and also prevents unwanted naming impacts

cypress/integration/ete/register_email_sent.spec.js
```diff
@@ -0,0 +1,79 @@
...
+  it('should resend the email when the resend button is clicked', () => {
+    cy.emptyInbox(resendTestEmail.inbox).then(() => {
+      cy.setCookie('PLAY_SESSION_2', testMailSlurpEncryptedEmailCookie, {
+        log: true,
+      });
+      cy.visit(`/register/email-sent`);
+      cy.contains(resendTestEmail.email);
+      cy.contains('Resend email').click();
+      cy.waitForLatestEmail(resendTestEmail.inbox).then((email) => {
+        expect(email.body).to.contain('Complete registration');
```
oliverlloyd, line 53: 👍 nice

cypress/integration/ete/register_email_sent.spec.js
```diff
@@ -0,0 +1,79 @@
...
+        const match = email.body.match(/theguardian.com\/welcome\/([^"]*)/);
+        const token = match[1];
+        cy.visit(`/welcome/${token}`);
+        cy.contains('7b4f97c5-79cd-4a3c-9212-cfb6db959dfc@mailslurp.com');
+      });
+    });
+  });
+
+  it('should navigate back to the correct page when change email is clicked', () => {
+    // set encrypted email
```
oliverlloyd, line 64: Is this comment needed? Comments come with a cost in that they need to be maintained to make sure that they still say the right thing. But in reality this rarely happens so they end up being wrong, or worse, misleading. Because of this it's good to use them sparingly and only when the code itself is not self explanatory

src/server/routes/register.ts
```diff
@@ -25,6 +29,52 @@ router.get(
...
+          email: getEmailFromPlaySessionCookie(req),
+        },
+      }),
+      pageTitle: PageTitle.REGISTRATION_EMAIL_SENT,
+    });
+    res.type('html').send(html);
+  },
+);
+
+// POST form handler to resend account verification email
```
oliverlloyd, line 48: Ditto regarding the comment - is it needed?

cypress/integration/ete/register_email_sent.spec.js
```diff
@@ -0,0 +1,79 @@
...
+    cy.emptyInbox(resendTestEmail.inbox).then(() => {
+      cy.setCookie('PLAY_SESSION_2', testMailSlurpEncryptedEmailCookie, {
+        log: true,
+      });
+      cy.visit(`/register/email-sent`);
+      cy.contains(resendTestEmail.email);
+      cy.contains('Resend email').click();
+      cy.waitForLatestEmail(resendTestEmail.inbox).then((email) => {
+        expect(email.body).to.contain('Complete registration');
+        // extract the reset token (so we can reset this reader's password)
```
oliverlloyd, line 54: This comment is not true!

cypress/integration/ete/register_email_sent.spec.js
ob6160: how about `mailSlurpEncryptedEmail` and `userEncryptedEmail` for the two emails we have in the file?

cypress/integration/ete/register_email_sent.spec.js
ob6160: will change

cypress/integration/ete/register_email_sent.spec.js
ob6160: agreed, will remove

cypress/integration/ete/register_email_sent.spec.js
oliverlloyd: Ha, it's so true that naming things is hard It is very much worth putting time in on it though - our future selves will thank us!

So in this file we have

```
testUserEmail
testUserEncryptedEmailCookie
resendTestEmail
testMailSlurpEncryptedEmailCookie
```

But these names don't help with the understanding of what the tests are doing. I think mainly because they have such long names but also because I'm not sure if I need to think differently about them based on their names

Breaking this down, there are only two tests using these variables. 

1. `should load the page with a success banner given a valid encrypted email cookie`
and
2. `should resend the email when the resend button is clicked`

1 uses `testUserEmail` & `testUserEncryptedEmailCookie`
2 uses `resendTestEmail` & `testMailSlurpEncryptedEmailCookie`

So I can now see that we have 2 email addresses and two encrypted strings used when setting a cookie.

But maybe we can use the same email address in both tests? I don't see any data being mutated and I think the resend logic works the same in both situations?

Like that we could simply have `email` and `encryptedEmail` which represents what the strings are and all the other context can drop away? Of course, that only works if these string really do work fine when you use them in both tests

ob6160: Yeah I think that's a really good perspective on the problem. I think you're 100% right, we don't need two email addresses here. I will change so that we only use the mailslurp email. No data is mutated and it should work fine when we use it for both tests (the other tests just want to find a valid email on the page)

oliverlloyd (approved):
🚀

cypress/integration/ete/register_email_sent.spec.js
```diff
@@ -0,0 +1,74 @@
...
+  // Depends on a Guest account already created using this email.
+  it('should resend the email when the resend button is clicked', () => {
+    cy.emptyInbox(email.inbox).then(() => {
+      cy.setCookie('PLAY_SESSION_2', encryptedEmail, {
+        log: true,
+      });
+      cy.visit(`/register/email-sent`);
+      cy.contains(email.value);
+      cy.contains('Resend email').click();
+      cy.waitForLatestEmail(email.inbox).then((emailResponse) => {
```
oliverlloyd, line 46:
```suggestion
      cy.waitForLatestEmail(email.inbox).then((response) => {
```

cypress/integration/ete/register_email_sent.spec.js
```diff
@@ -0,0 +1,74 @@
...
+        expect(emailResponse.body).to.have.string('Complete registration');
```
oliverlloyd, line 47:
```suggestion
        expect(response.body).to.have.string('Complete registration');
```

cypress/integration/ete/register_email_sent.spec.js
```diff
@@ -0,0 +1,74 @@
...
+        // Extract the welcome token, so we can redirect to the welcome flow.
+        const match = emailResponse.body.match(
```
oliverlloyd, line 49:
```suggestion
        const match = response.body.match(
```

oliverlloyd (approved):
Just a persnickety suggested naming change but otherwise 👍 !

ob6160:
> Just a persnickety suggested naming change but otherwise 👍 !

persnickety is my new word of the day!

PR #840, opened 2021-08-23
Always show 👁️ 

oliverlloyd:
## What does this change?
Replaces the plain text input used on the `SignIn` page for passwords with out abstracted `PasswordInput`.

## Why
So that we can offer the reader the ability to show/hide their password

### Before
Screenshot 2021-08-23 at 11 01 09

### After
Screenshot 2021-08-23 at 11 00 52


oliverlloyd:
[Show and hide the password on the sign in page!](https://trello.com/c/ACAl9gpq/3319-note-from-ux-ideally-we-would-have-the-ability-to-show-and-hide-the-password-on-the-sign-in-page-as-well-is-this-possible-for-mv)

src/client/components/PasswordInput.tsx
```diff
@@ -10,8 +10,10 @@ import { ClientState } from '@/shared/model/ClientState';
 import { ClientStateContext } from '@/client/components/ClientState';
 
 type Props = {
+  label: string;
   error?: string;
-  onChange: (e: React.ChangeEvent<HTMLInputElement>) => void;
+  supporting?: string;
```
ob6160, line 15: wonder if `supportText` or something along those lines might be a clearer name for this?

ob6160 (approved):
nice work, one small comment about a prop name

src/client/components/PasswordInput.tsx
oliverlloyd: I agree that this would be a clearer name however this property is set by Source originally so we're just proxying it through 🤷

PR #851, opened 2021-08-24
No div inside p

oliverlloyd:
## What does this change?
Removes some html that was setting a div inside a p tag, breaking html semantics

## Why?
It caused things to look broken when server rendered. This seemed to get fixed on the client after hydration but the brokwn view was still being shown, causing jank

## Before
Screenshot 2021-08-24 at 12 25 16

## After
Screenshot 2021-08-24 at 12 26 38


ob6160 (approved):
nice : )

PR #852, opened 2021-08-24
Don't hide success

oliverlloyd:
## What does this change?
Shows the success message for longer

Previously, we removed the success message from the page after n seconds. Now we keep it on there all the time, only hiding it when Resend Email is clicked

### Before
b

### After
a

### DIFF
--- a/src/client/layouts/Main.tsx
+++ b/src/client/layouts/Main.tsx
@@ -1,4 +1,4 @@
-import React, { useContext, useState, useEffect } from 'react';
+import React, { useContext } from 'react';
 import { SubHeader } from '@/client/components/SubHeader';
 import { ClientState } from '@/shared/model/ClientState';
 import { ClientStateContext } from '@/client/components/ClientState';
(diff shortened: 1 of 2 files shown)


oliverlloyd:
[Show/Hide Success based on `loading`](https://trello.com/c/aCUbzmUm/3322-show-hide-success-based-on-loading)

ob6160 (approved):
lgtm :)

ob6160:
> lgtm :)

One quick question. If the resend fails for some reason do we want to display an error message instead of never setting loading state back to false?

oliverlloyd:
> > lgtm :)
> 
> One quick question. If the resend fails for some reason do we want to display an error message instead of never setting loading state back to false?

In this case the form submission causes the page to reload so there's no need to change state. Indeed, it's not even possible because there's no callback or onResponse event. I agree though it's a bit confusing having the state seemingly just hang, but because we're using forms for making POST requests instead of, say, using `fetch`, this is how it works

If there's an error on the server then we render the page that the form was POSTing to (`previousPage`) with an error message.

PR #855, opened 2021-08-25
feat: Run with `unsafe-eval` in our CSP, but only locally

oliverlloyd:
## What does this change?
Without `unsafe-eval` in our Helmet config we see CSP errors when running locally. This resolves this by conditionally setting this setting based on `stage`


ob6160 (approved):
:shipit:

coldlink (approved):

PR #857, opened 2021-08-26
Space footer items equally

oliverlloyd:
## What does this change?
Replaces

```css
flex-basis: auto; /* this was the default previously so wasn't explicitly present */
flex-grow: 1;
```

 with

```css
flex: 1;
```

which is shorthand for

```css
flex-basis: 0;
flex-grow: 1;
```

## Why?
To ensure that all footer groups are evenly spaced.

Some [interesting reading](https://stackoverflow.com/questions/25066214/flexbox-not-giving-equal-width-to-elements/25066844#25066844) on why this change works

### Before
Screenshot 2021-08-26 at 10 58 03

### After
Screenshot 2021-08-26 at 10 57 36

### DIFF
--- a/src/client/components/Footer.stories.tsx
+++ b/src/client/components/Footer.stories.tsx
@@ -32,3 +32,11 @@ Tablet.parameters = {
     defaultViewport: 'TABLET',
   },
 };
+
+export const Mobile320 = () => <Footer />;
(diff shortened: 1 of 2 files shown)


ob6160 (approved):
this pull request has left me feeling well centred

PR #877, opened 2021-08-31
Update Mailslurp Test

ob6160:
## What does this change?
Sets the Mailslurp inbox to a new one created through an account registered under my name.

This is a temporary solution necessitated by the mail limit enforced by Mailslurp. This should be resolved once we have a paid team account in place.

### DIFF
--- a/cypress/integration/ete/reset_password.spec.js
+++ b/cypress/integration/ete/reset_password.spec.js
@@ -4,8 +4,8 @@ describe('Password reset flow', () => {
   context('Account exists', () => {
     // This test depends on this Mailslurp account already being registered
     const existing = {
-      email: '0298a96c-1028-4e3d-b943-a2b478c84dbd@mailslurp.com',
-      inbox: '0298a96c-1028-4e3d-b943-a2b478c84dbd',
+      email: 'a1ae9048-2823-45a0-a0e1-89e8f7a22420@mailslurp.com',
+      inbox: 'a1ae9048-2823-45a0-a0e1-89e8f7a22420',
     };
 
     it("changes the reader's password", () => {


cypress/integration/ete/reset_password.spec.js
```diff
@@ -4,8 +4,8 @@ describe('Password reset flow', () => {
   context('Account exists', () => {
     // This test depends on this Mailslurp account already being registered
     const existing = {
-      email: '0298a96c-1028-4e3d-b943-a2b478c84dbd@mailslurp.com',
```
dominickendrick, line 7: Should this key not be private ? Can we put it in config instead ?

ob6160: good point! this is how the emails were set up by @oliverlloyd so not sure if he considered this already. I guess we could have a test email config file that we populate?

ob6160: seems important because someone could come along and use up our mailbox limit

oliverlloyd: This isn't the key. It's just a random uuid to get a unique email address

oliverlloyd: The key (the api key) is set as a Github Secret

oliverlloyd (approved):

cypress/integration/ete/reset_password.spec.js
oliverlloyd: > seems important because someone could come along and use up our mailbox limit

This is technically true so we could abstract it... That said, I suspect this vector is wide open is many other places (do we keep test email accounts secret?) and any attacker would lose more than they gain, so it would purely be an attack of spite. But if we really to be sure we could put these in env vars.

The downside to this is it adds to the work needed to onboard new devs, adding new tests and changing values. Plus the _whole team_ would need to action every change/new test so that they also have the secret values set locally. Personally I don't think this is worth it but it's a judgement call.

PR #878, opened 2021-08-31
Upgrade project dependencies to latest

ob6160:
## What does this change?
Brings dependencies up to date

Skipped upgrades with reasons:
* preset-react (https://github.com/guardian/gateway/pull/878#discussion_r702791500)
* node-fetch (https://github.com/node-fetch/node-fetch/issues/1266)

Typescript 4.4 upgrade split out here: #881 
Source upgrade split out here: #882


package.json
```diff
@@ -54,45 +54,45 @@
...
     "cypress-axe": "^0.13.0",
     "dotenv-webpack": "^7.0.3",
     "eslint": "^7.32.0",
     "eslint-config-prettier": "^8.3.0",
-    "eslint-plugin-functional": "^3.6.0",
+    "eslint-plugin-functional": "^3.7.0",
     "eslint-plugin-jsx-a11y": "^6.4.1",
-    "eslint-plugin-prettier": "^3.4.1",
-    "eslint-plugin-react": "^7.24.0",
+    "eslint-plugin-prettier": "^4.0.0",
```
oliverlloyd, line 81: Ref: https://github.com/prettier/eslint-plugin-prettier/blob/master/CHANGELOG.md

oliverlloyd (commented):
Thanks for doing this!

One thing, can we break the typescript changes out into a separate PR? This will help keep the git history useful for future us.

Also, there's some work triggered around the logo by the Source upgrade so it would be good to keep that separate as well (for similar reasons).

package.json
ob6160: > Thanks for doing this!
> 
> One thing, can we break the typescript changes out into a separate PR? This will help keep the git history useful for future us.
> 
> Also, there's some work triggered around the logo by the Source upgrade so it would be good to keep that separate as well (for similar reasons).

that split sounds good to me, will let you know once I've done that

package.json
```diff
@@ -35,12 +35,12 @@
     "@babel/plugin-proposal-optional-chaining": "^7.14.5",
     "@babel/plugin-transform-runtime": "^7.15.0",
     "@babel/preset-env": "^7.15.0",
-    "@babel/preset-react": "^7.13.13",
+    "@babel/preset-react": "^7.14.5",
```
coldlink, line 38: This is the buggy version of preset-react, we can't update this. Still need to look into what's causing this. So best to leave it at `"^7.13.13"` for now.

coldlink (commented):
Small comment.

Also best to run another `make upgrade`/`yarn upgrade-interactive --latest` this morning to pull in new dependencies from the last few days.

package.json
ob6160: Aha, good spot! I'll roll it back

coldlink: Looks like there are still problems. I have a feeling that the lockfile is still pulling in dependencies from 7.14.5.

My suggestion would actually just be to close this PR, and open up a new one with the rest of the updates, minus this dependency.

ob6160: ah I missed something with my fix for node-fetch. I think all fixed with latest push

ob6160:
closed in favour of: https://github.com/guardian/gateway/pull/903

package.json
ob6160: https://github.com/guardian/gateway/pull/903

PR #879, opened 2021-08-31
Stack social buttons

oliverlloyd:
## What does this change?
We now respect the [rules laid out in the design system](https://theguardian.design/2a1e5182b/p/435225-button/b/86f344/t/053f03) and stack groups of buttons on mobile

### Before
Screenshot 2021-08-31 at 13 58 29

### After
Screenshot 2021-08-31 at 13 56 44

### DIFF
--- a/src/client/components/SocialButtons.tsx
+++ b/src/client/components/SocialButtons.tsx
@@ -5,6 +5,7 @@ import { LinkButton } from '@guardian/src-button';
 import { SvgGoogleBrand } from '@guardian/src-icons';
 import { SvgAppleBrand } from '@guardian/src-icons';
 import { SvgFacebookBrand } from '@guardian/src-icons';
+import { from } from '@guardian/src-foundations/mq';
 
 type Props = {
   returnUrl: string;
@@ -12,16 +13,22 @@ type Props = {
 
 const containerStyles = css`
   display: flex;
-  flex-direction: row;
+  flex-direction: column;
+  ${from.mobileLandscape} {
+    flex-direction: row;
+  }
   justify-content: center;
+  margin-top: ${space[12]}px;
+  margin-bottom: ${space[12]}px;
   width: 100%;
-  margin: ${space[9]}px 0;
 `;
 
 const buttonOverrides = css`
   border-color: ${brand[400]};
   justify-content: flex-end;
-  min-width: 145px;
+  ${from.mobileLandscape} {
+    min-width: 145px;
+  }
 `;
 
 // TODO: If the issue below is fixed and a new version of Source published with that fix in it, then
@@ -36,7 +43,12 @@ const iconOverrides = css`
 const Gap = () => (
   <span
     css={css`
-      width: ${space[2]}px;
+      width: 0;
+      height: ${space[3]}px;
+      ${from.mobileLandscape} {
+        width: ${space[2]}px;
+        height: 0;
+      }
     `}
   ></span>
 );
--- a/src/client/components/SocialButtons.stories.tsx
+++ b/src/client/components/SocialButtons.stories.tsx
@@ -16,6 +16,7 @@ Desktop.parameters = {
   viewport: {
     defaultViewport: 'DESKTOP',
   },
+  chromatic: { viewports: [1300] },
 };
 
 export const Mobile = () => (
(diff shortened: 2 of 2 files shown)


oliverlloyd:
[Social buttons should stack on mobile](https://trello.com/c/023oZHr4/3320-social-buttons-should-stack-on-mobile)

src/client/components/SocialButtons.tsx
```diff
@@ -5,23 +5,30 @@ import { LinkButton } from '@guardian/src-button';
...
   width: 100%;
-  margin: ${space[9]}px 0;
 `;
 
 const buttonOverrides = css`
   border-color: ${brand[400]};
   justify-content: flex-end;
-  min-width: 145px;
+  ${from.mobileLandscape} {
+    min-width: 145px;
```
ob6160, line 30: wondering if we can use a space value for this?

ob6160 (commented):
nice one. I wonder, instead of using a div with `containerStyles` could we use the design system stack component to stack them? https://guardian.github.io/source/?path=/story/source-src-layout-stack--playground

src/client/components/SocialButtons.tsx
oliverlloyd: I think `space` is more about distance between things, so more aimed at padding and margins. I think arbitrary numbers in the code is fine in cases like this where we want to prevent the buttons stretching too far

oliverlloyd:
> nice one. I wonder, instead of using a div with `containerStyles` could we use the design system stack component to stack them? https://guardian.github.io/source/?path=/story/source-src-layout-stack--playground

I initially got excited as this component looks great. However, it doesn't have a way to limit stacking based on a breakpoint so by using it we get stacked buttons all the time, even above `mobileLandscape`.

Ideally there would be a couple of props like `above: Breakpoint` or `below: Breakpoint` to set conditions. @SiAdcock ?

ob6160:
> > nice one. I wonder, instead of using a div with `containerStyles` could we use the design system stack component to stack them? https://guardian.github.io/source/?path=/story/source-src-layout-stack--playground
> 
> I initially got excited as this component looks great. However, it doesn't have a way to limit stacking based on a breakpoint so by using it we get stacked buttons all the time, even above `mobileLandscape`.
> 
> Ideally there would be a couple of props like `above: Breakpoint` or `below: Breakpoint` to set conditions. @SiAdcock ?

I've worked on a codebase that uses https://chakra-ui.com/docs/layout/stack#responsive-direction - they have a really nice way of handling responsive direction which we could maybe take cues from @oliverlloyd @SiAdcock

src/client/components/SocialButtons.tsx
ob6160: seems like a fair approach. tangential question - what's our policy on using `rem`s, I've seen some people take a pretty hardline approch to using them everywhere

oliverlloyd: That's a big topic!

Might be a good question to ask in the Client Side channel? I'm sure there are lots of great ideas out there.

SiAdcock:
Maybe you've already seen it, but would the [`Columns`](
  <Column></Column>) component work, with the `collapseUntil` prop?

```
<Columns collapseUntil="tablet">
  <Column></Column>
  <Column></Column>
  <Column></Column>
</Columns>
```

oliverlloyd:
> we should define a `<ButtonGroup>` component in Source that enforces these layout rules.

Very much agree on this. We have a clear distinction for the design of buttons when in a group so encapsulating that in a component makes a lot of sense

ob6160 (approved):
⭐  

re: button groups, what's the standard procedure for moving forward with something new like this? @oliverlloyd @SiAdcock  I'm interested in landing a first contribution to Source :)

oliverlloyd:
> ⭐
> 
> re: button groups, what's the standard procedure for moving forward with something new like this? @oliverlloyd @SiAdcock I'm interested in landing a first contribution to Source :)

Do you want to pair on this? This looks like a great issue for a contribution!

SiAdcock:
This would be greatly appreciated! If you can hold off until next week, we are going to launch the [Source Development Kitchen](https://github.com/guardian/source/pull/963) that will give you a fast track to getting components into Source. Does that sound okay?

ob6160:
Brilliant, yes @oliverlloyd let's put some time in next week! Looks great  @SiAdcock, in the meantime I'll have a play around with the kitchen and familiarise myself with what's going on there

PR #880, opened 2021-09-02
Small CSS tweak to fix formatting on short pages

rebecca-thompson:
## What does this change?
On short pages (such as the new set password page on the onboarding flow), the vertical line does not go right to the bottom of the page. See screenshots below for the before/after

## Images
Before
Screenshot 2021-09-02 at 09 45 28

After
Screenshot 2021-09-02 at 09 45 56

### DIFF
--- a/src/client/app.tsx
+++ b/src/client/app.tsx
@@ -36,7 +36,7 @@ export const App = (props: ClientState) => {
             height: 100%;
           }
           #app {
-            min-height: 100%;
+            height: 100%;
             display: flex;
             flex-direction: column;
           }


ob6160 (approved):
looks way better!

oliverlloyd:
I wonder if there's a way to add a story for this? It'd be great to capture this use case and make sure we don't have issues with it in the future.

Perhaps there's a way to force the viewport height to be larger in Storybook / Chromatic?

rebecca-thompson:
[MVP0 design tweaks](https://trello.com/c/cKH45aRk/3361-mvp0-design-tweaks)

PR #881, opened 2021-09-03
Chore/upgrade typescript

ob6160:
## What does this change?
Upgrades Typescript to 4.4

Typescript 4.4 has turned on `unknown` typing of thrown errors by default instead of using `any`. This has meant that we need to use stricter types in the places where we destructure caught errors. Ref: https://devblogs.microsoft.com/typescript/announcing-typescript-4-4-rc/#use-unknown-catch-variables

_Broken off from original PR #878_


src/server/routes/consents.ts
```diff
@@ -558,12 +560,13 @@ router.post(
         url = addReturnUrlToPath(url, state.queryParams.returnUrl);
       }
       return res.redirect(303, url);
-    } catch (e) {
-      logger.error(`${req.method} ${req.originalUrl}  Error`, e);
-      status = e.status;
+    } catch (error) {
+      logger.error(`${req.method} ${req.originalUrl}  Error`, error);
+      const { status: errorStatus, message } = error as RequestError;
+      status = errorStatus;
```
oliverlloyd, line 566: Do we do anything with `status` here?

Also, do we need to rename it to `errorStatus` if we are just going to reassign it back to `status` straight away? Sorry if I'm missing some context on that one - possible!

oliverlloyd (approved):
Nice 👍 

Be good to squashy merge this one to squish those fix commits

src/server/routes/consents.ts
ob6160: yep, so it was engineered so that there's a single `status` variable for the whole request. above this, status is set to `200` and if an error is caught here it is only then set to the `errorStatus` from the error.

below this we then return the status from the controller

PR #882, opened 2021-09-03
Upgrade Guardian source library and add missing resend email test back

ob6160:
## What does this change?
Updates Source separately because of Chromatic diff. Also includes upgrade to guardian/libs 
Adds back test which was mistakenly removed in #881 (whoops)


package.json
```diff
@@ -105,17 +105,17 @@
...
-    "@guardian/src-brand": "^3.8.0",
-    "@guardian/src-button": "^3.8.0",
-    "@guardian/src-checkbox": "^3.8.0",
-    "@guardian/src-foundations": "^3.8.0",
-    "@guardian/src-icons": "^3.8.0",
-    "@guardian/src-layout": "3.8.0",
-    "@guardian/src-link": "^3.8.0",
-    "@guardian/src-radio": "^3.8.0",
-    "@guardian/src-text-input": "^3.8.0",
+    "@guardian/libs": "^3.1.0",
```
oliverlloyd, line 108: Nitpick but this lib isn't aprt of source

cypress/integration/ete/register_email_sent.spec.js
```diff
@@ -0,0 +1,72 @@
+/* eslint-disable @typescript-eslint/no-var-requires */
```
oliverlloyd, line 1: Is this file meant to be part of this PR?

ob6160: yep, mistake where I deleted it in the last upgrade Friday slipup 😳

package.json
ob6160: all good if I add it to the description of the PR?

coldlink (approved):

coldlink:
Looking good, the only thing I would say is to update the title of the PR before merging in, so that it's more meaningful!

PR #905, opened 2021-09-07
feat: track referral event in Ophan

ob6160:
## What changed?
We use the `viewId` passed in from the account verification email to execute the Ophan referral event once on first render of the `/welcome` flow.

### DIFF
--- a/src/client/pages/Welcome.tsx
+++ b/src/client/pages/Welcome.tsx
@@ -9,26 +9,23 @@ import {
 import { getAutoRow, gridItemColumnConsents } from '@/client/styles/Grid';
 import { CONSENTS_PAGES } from '@/client/models/ConsentsPages';
 import { sendOphanReferrerEvent } from '../lib/ophan';
-import { useLocation } from 'react-router-dom';
+
 type Props = {
   submitUrl: string;
   fieldErrors: FieldError[];
+  viewId?: string;
 };
 
-export const Welcome = ({ submitUrl, fieldErrors }: Props) => {
+export const Welcome = ({ submitUrl, fieldErrors, viewId }: Props) => {
   const autoRow = getAutoRow(1, gridItemColumnConsents);
-  const { search } = useLocation();
 
   useEffect(() => {
-    const params = new URLSearchParams(search);
-    const viewId = params.get('viewId');
-
     if (viewId) {
       sendOphanReferrerEvent({
         viewId,
       });
     }
-  }, []);
+  }, [viewId]);
 
   return (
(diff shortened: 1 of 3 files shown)


ob6160:
[Analytics for the new pages : Onboarding flow](https://trello.com/c/6QGlplge/3265-analytics-for-the-new-pages-onboarding-flow)

src/client/pages/Welcome.stories.tsx
```diff
@@ -2,11 +2,19 @@ import React from 'react';
...
 
 export default {
   title: 'Pages/Welcome',
   component: Welcome,
   parameters: { layout: 'fullscreen' },
+  decorators: [
+    (Story) => (
+      <MemoryRouter>
+        <Story />
+      </MemoryRouter>
```
coldlink, line 15: I'm not sure I understand what's going on here, why do we need the `MemoryRouter`?

coldlink: Ah wait, you need it in the test/story to render the component, as `location` would not exist otherwise.

coldlink (approved):

src/client/pages/Welcome.stories.tsx
coldlink: I wonder if it's worth moving the `useLocation` from the `Welcome.tsx` to `WelcomePage.tsx` and passing in search as a `prop`, this would mean you wouldn't need the dependency on `react-router-dom` here, and the `Welcome` component could be rendered independently.

coldlink: you could either pass in `search` as the prop, or go a level lower and pass in `viewId` as a prop to the welcome component as an optional prop.

ob6160: I like the idea of passing down `viewId`. To me passing `viewId` as a prop shows intent more clearly than passing  `search` which could be used for anything. keeps the URL param fetching logic all in one place too

src/client/pages/Welcome.tsx
```diff
@@ -8,14 +8,25 @@ import {
 } from '@/client/layouts/shared/Consents';
 import { getAutoRow, gridItemColumnConsents } from '@/client/styles/Grid';
 import { CONSENTS_PAGES } from '@/client/models/ConsentsPages';
+import { sendOphanReferrerEvent } from '../lib/ophan';
```
coldlink, line 11:
```suggestion
import { sendOphanReferrerEvent } from '@/client/lib/ophan';
```

small style change 😆

PR #906, opened 2021-09-13
Fix Ophan referral on welcome/set password page

coldlink:
## What does this change?

When investigating why the `referrer_page_view_id` was not being populated in the data late, we found that the way we were sending this parameter to ophan was incorrect. Instead of sending just the `viewId` as a parameter using ophan's `record` method, this should actually be called `refViewId` according to the Ophan source files:

https://github.com/guardian/ophan/blob/70b658e785c490c411670bbd3c7fde62ae0224fc/the-slab/app/extractors/ReferrerExtractor.scala#L129

We also realised that we needed to pass a second parameter to Ophan too. This is the `ref` parameter, which is the url of the referrer page.

https://github.com/guardian/ophan/blob/70b658e785c490c411670bbd3c7fde62ae0224fc/the-slab/app/extractors/ReferrerExtractor.scala#L171

This PR renames `viewId` to `refViewId` and adds the `ref` parameter. We then make sure that we're sending these parameters as expected by Ophan.

This PR also adds a method to validate the `ref` url parameter to make sure that it's from a Guardian domain. Which is why `validateReturnUrl.ts` was renamed to `validateUrl.ts` as this now exports multiple methods.

### DIFF
--- a/src/client/pages/Welcome.tsx
+++ b/src/client/pages/Welcome.tsx
@@ -13,19 +13,26 @@ import { sendOphanReferrerEvent } from '@/client/lib/ophan';
 type Props = {
   submitUrl: string;
   fieldErrors: FieldError[];
-  viewId?: string;
+  refViewId?: string;
+  refUrl?: string;
 };
 
-export const Welcome = ({ submitUrl, fieldErrors, viewId }: Props) => {
+export const Welcome = ({
+  submitUrl,
+  fieldErrors,
+  refViewId,
+  refUrl,
+}: Props) => {
   const autoRow = getAutoRow(1, gridItemColumnConsents);
 
   useEffect(() => {
-    if (viewId) {
+    if (refViewId && refUrl) {
       sendOphanReferrerEvent({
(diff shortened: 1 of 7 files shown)


src/client/pages/Welcome.tsx
```diff
@@ -13,19 +13,26 @@ import { sendOphanReferrerEvent } from '@/client/lib/ophan';
 type Props = {
   submitUrl: string;
   fieldErrors: FieldError[];
-  viewId?: string;
+  refViewId?: string;
+  refPage?: string;
```
coldlink, line 17: We have to call the prop `refPage` rather than `ref` as `ref` is a React specific prop: https://reactjs.org/docs/refs-and-the-dom.html

rebecca-thompson (approved):
Looks good 👍

src/client/pages/Welcome.tsx
coldlink: changed to `refUrl` as that makes more semantic sense

PR #930, opened 2021-09-20
Fix Ophan referral tracking using ophan init method

coldlink:
Despite the efforts in https://github.com/guardian/gateway/pull/906 we noticed that the referral information was still not pulling through in Ophan and the data lake.

After some research we discovered that the `ref` parameter (the url of the referrer page), is only able to be recorded on the init method of the `ophan-tracker-js` library.

https://github.com/guardian/ophan/blob/fb36f3dab919abb482b453bdf6b6fbd5e60e7a44/tracker-js/assets/coffee/ophan/core.coffee#L18

However there was no way of modifying the values directly of what was sent in this init method.

@ob6160 then discovered that using `localStorage` it is possible to override the parameters sent to ophan in the init method.

https://github.com/guardian/ophan/blob/fb36f3dab919abb482b453bdf6b6fbd5e60e7a44/tracker-js/assets/coffee/ophan/core.coffee#L66

This means that this commit removes the previous way that we were sending the `ref` abd `refViewId` parameter. Now before ophan is imported and initialised, we get the `ref` and the `refViewId` parameter and add them to localStorage key `ophan_follow` if they exist, which will be picked up by ophan, and then sent across in the when ophan
is imported.

This is done in the same way that the `support` platform sends this same information.

https://github.com/guardian/support-frontend/blob/3a1ff437be200f2e37013b519ed037566be67b9c/support-frontend/assets/helpers/tracking/ophan.js#L116-L126


ob6160 (approved):
great work :)

src/client/static/analytics/ophan.ts
```diff
@@ -1,4 +1,26 @@
...
+const addRefToOphanFollow = () => {
+  if (storage.local.isAvailable()) {
+    const { search } = window.location;
+    const params = new URLSearchParams(search);
+    const ref = params.get('ref');
+    const refViewId = params.get('refViewId');
+
+    if (ref && refViewId) {
+      storage.local.setRaw(
+        'ophan_follow',
```
ob6160, line 13: one small q, could we be overwriting a value already set against `ophan_follow` in local storage here?

coldlink: I dont think so, as it seems it's hardly used: https://github.com/search?q=org%3Aguardian+ophan_follow&type=code

oliverlloyd:
I know this is merged and working (thanks to the great spot by @ob6160 !) but I thought I'd check that there wasn't any relation to [this code](https://github.com/guardian/dotcom-rendering/blob/0286680bb32e8256988003d8de8f1bf2e4ebb30b/dotcom-rendering/src/web/server/htmlTemplate.ts#L232) in DCR?

To be honest, I'm not really following along but I noticed talk about Ophan and also things that sounded like race conditions so my brain surfaced this piece of code and I thought I'd drop it here in case it's useful.

PR #933, opened 2021-09-21
chore: upgrade node-fetch to v3

ob6160:
Upgrades node-fetch to v3 and uses an asynchronous import to avoid converting the import to a CJS require which is no longer supported by the library because it is ESM-only. Comment included to explain the solution which was sourced from: https://github.com/node-fetch/node-fetch/issues/1279#issuecomment-915063354


src/server/lib/IDAPIFetch.ts
```diff
@@ -1,9 +1,20 @@
 /* eslint-disable @typescript-eslint/no-explicit-any */
 
-import fetch, { RequestInit, Response } from 'node-fetch';
+import type { RequestInit, Response } from 'node-fetch';
```
coldlink, line 3: woah, honestly surprised that this is the only file where we're using node-fetch

coldlink (approved):

PR #935, opened 2021-09-23
Upgrade dependencies

ob6160:
## What does this change?
Brings deps up to date to solve some dependabot vulns


coldlink:
Looks like the build has broken. Worth looking into.

package.json
```diff
@@ -130,7 +130,7 @@
     "mjml": "^4.10.3",
     "mjml-browser": "^4.10.3",
     "mjml-react": "^2.0.0",
-    "ms": "^2.1.3",
+    "ms": "^3.0.0-canary.1",
```
coldlink, line 133: this is causing the build to break, canary is pre-release, not sure we want to use it

coldlink: i think 3.0.0 also uses ESM, so same issue as node-fetch

coldlink:
As a general suggestion for your PRs, I also suggest updating the title to be more relevant rather than just using the commit messages.

package.json
ob6160: rolling back

coldlink:
> ## What does this change?
> Brings deps up to date to solve some dependabot vulns
> 
> Screenshot 2021-09-23 at 16 11 15
Sorry another small nitpick:
I wouldn't suggest showing screenshots of possible vulnerabilities on the PR, as this is public, and that info could be used as a possible attack vector.

coldlink (approved):

ob6160:
> > ## What does this change?
> > Brings deps up to date to solve some dependabot vulns
> > Screenshot 2021-09-23 at 16 11 15
> 
> Sorry another small nitpick:
> I wouldn't suggest showing screenshots of possible vulnerabilities on the PR, as this is public, and that info could be used as a possible attack vector. Whereas the alerts themselves are not public.

thanks really good to know - didn't realise dependabot updates weren't public!

ob6160:
Looking into IE11 problem running Storybook which causes the Chromatic build to fail

ob6160:
> Looking into IE11 problem running Storybook which causes the Chromatic build to fail

Update: I don't think this is caused by the PR, seems as though our Storybook build isn't running in IE11 when deployed to Chromatic. I've tested a rerun of the Chromatic Github Action from main and that is failing too: https://github.com/guardian/gateway/runs/3690743799?check_suite_focus=true

Looks like something is up with Chromatic. I've also tested locally and Storybook won't run in IE11 locally either.

ob6160:
> > Looking into IE11 problem running Storybook which causes the Chromatic build to fail
> 
> Update: I don't think this is caused by the PR, seems as though our Storybook build isn't running in IE11 when deployed to Chromatic. I've tested a rerun of the Chromatic Github Action from main and that is failing too: https://github.com/guardian/gateway/runs/3690743799?check_suite_focus=true
> 
> Although something is likely up with Chromatic, I have also tested locally and Storybook won't run in IE11 on my machine either.

I've fixed IE11 support in the Storybook build here: https://github.com/guardian/gateway/pull/936. Once that's merged we can merge this.

Presuming that Chromatic made their service break if IE11 couldn't run Storybook as it seems to fix the issue.

PR #936, opened 2021-09-23
Fix Storybook in IE11

ob6160:
## What does this change?
Storybook wasn't working in IE11 because the `@guardian/*` and `strip-ansi` modules were not being transpiled into ES5.

This PR adds a rule to the Storybook webpack configuration which transpiles these files using the same rule that we use for our production build.

This change was made because the Chromatic IE11 build started to fail, presumably because of a change made on their side which failed the build if a render in IE11 couldn't be made (**this is a guess, needs a source to be sure**)
### Before
Screenshot 2021-09-23 at 19 58 41

### After
Screenshot 2021-09-23 at 19 48 30

### DIFF
--- a/.storybook/main.js
+++ b/.storybook/main.js
@@ -1,4 +1,5 @@
 const path = require('path');
+const babelConfig = require('../babel.config');
 
 module.exports = {
   stories: ['../src/**/*.stories.mdx', '../src/**/*.stories.@(js|jsx|ts|tsx)'],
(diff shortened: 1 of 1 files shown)


coldlink (approved):

coldlink:
As a note, this broke as I enabled IE11 (and Firefox) visual regression testing in the chromatic dashboard for gateway, as I thought it would be good to do visual regression testing for all major rendering engines.

PR #939, opened 2021-09-27
Fixing X-Forwarded-For header in IDAPI requests

rebecca-thompson:
## What does this change?
We are currently sending the IP address of the EC2 instance in the `X-Forwarded-For` header to IDAPI, which is then being used for rate limiting. This configures the Express server to be able to pass in the client's IP address. This is required when the server is sitting behind a reverse proxy (https://expressjs.com/en/guide/behind-proxies.html).

## How to test
This was tested by running gateway `CODE` with and without changing the setting and seeing the different IP addresses being logged and passed through to IDAPI.

### DIFF
--- a/src/server/index.ts
+++ b/src/server/index.ts
@@ -10,4 +10,5 @@ const server: Express = express();
 applyMiddleware(server);
 
 server.listen(port);
+server.set('trust proxy', true);
 logger.info(`server running on port ${port}`);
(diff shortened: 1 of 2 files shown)


rebecca-thompson:
[gateway should forward client IP addresses to IDAPI endpoints to enable IP rate limiting (T)](https://trello.com/c/YQ4Q6GoE/3311-gateway-should-forward-client-ip-addresses-to-idapi-endpoints-to-enable-ip-rate-limiting-t)

QuarpT (approved):

PR #951, opened 2021-09-27
Update mailosaur to use new trial account

coldlink:
## What does this change?
- Updates the `encryptedEmail` in the test
- Adds the mailosaur variables to `.env.example`
- Fixes flaky tests by checking url instead of content


ob6160 (approved):
nice work 😄

PR #953, opened 2021-09-28
Update the Grid styles

coldlink:
## What does this change?
Updating the grid styles currently used in Gateway to that defined in our Figma file. The template of which is shown here:

Screenshot_2021-09-27_at_11 55 09

I've also added some storybook files with examples on how to use the grid styles.

Following on from this PR, we will be updating the pages not currently using the grid to use it.

### DIFF
--- a/src/client/styles/Grid.ts
+++ b/src/client/styles/Grid.ts
@@ -5,60 +5,94 @@ import { from } from '@guardian/src-foundations/mq';
 
 export enum COLUMNS {
   MOBILE = 4,
+  MOBILE_MEDIUM = 4,
+  MOBILE_LANDSCAPE = 4,
   TABLET = 12,
   DESKTOP = 12,
+  LEFT_COL = 14,
   WIDE = 16,
 }
 
 enum COLUMN_WIDTH {
   MOBILE = 'minmax(0, 1fr)',
+  MOBILE_MEDIUM = 'minmax(0, 1fr)',
+  MOBILE_LANDSCAPE = 'minmax(0, 1fr)',
   TABLET = '40px',
   DESKTOP = '60px',
+  LEFT_COL = '60px',
   WIDE = '60px',
 }
 
 enum SPACING {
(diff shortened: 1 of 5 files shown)


src/client/styles/Grid.ts
```diff
@@ -80,15 +114,28 @@ export const gridRow = css`
   display: -ms-grid;
   display: grid;
   width: 100%;
-  column-gap: ${px(space[5])};
 
-  -ms-grid-columns: \(${COLUMN_WIDTH.MOBILE}\)\[${COLUMNS.MOBILE}\];
-  grid-template-columns: repeat(${COLUMNS.MOBILE}, ${COLUMN_WIDTH.MOBILE});
-  padding-left: ${px(SPACING.MOBILE)};
-  padding-right: ${px(SPACING.MOBILE)};
```
coldlink, line 88: These particular css lines were generated already using `generateGridRowCss` and `MOBILE`

src/client/styles/Grid.ts
```diff
@@ -5,60 +5,94 @@ import { from } from '@guardian/src-foundations/mq';
...
   TABLET = '40px',
   DESKTOP = '60px',
+  LEFT_COL = '60px',
   WIDE = '60px',
 }
 
 enum SPACING {
-  MOBILE = space[3],
+  MOBILE = 10,
+  MOBILE_MEDIUM = 10,
```
rebecca-thompson, line 28: weird that there's no value in the source library for this. Is it worth adding it?

rebecca-thompson (approved):
🎉

src/client/styles/Grid.ts
coldlink: It's because 10px isn't a standard source unit. Source units are normally in multiples of 4(px).

After speaking to the source team yesterday, they're going to be reviewing the spacing with the designer to allow for non standard units like this, as this isn't the only place where non-standard units are used.

PR #954, opened 2021-10-01
Add source kitchen and use where possible

coldlink:
The source development kitchen has recently been released with the aim of speeding up component development.

https://github.com/guardian/source/tree/main/packages/%40guardian/source-react-components-development-kitchen

This PR adds the library as a dependency, and replaces existing components to those that have been moved to the kitchen where possible.

`ErrorSummary` wasn't being used anywhere, so I removed the component, as this can also be provided by the kitchen.


oliverlloyd (approved):
🥇 Drop in replacement with no diffs! 🥇 

(Well, okay, there's a 1 pixel diff on Divider but I think this is okay)

PR #957, opened 2021-10-04
Update header component to use the grid

coldlink:
## Accessibility


src/client/components/Header.tsx
```diff
@@ -2,64 +2,56 @@ import React from 'react';
...
-    margin-bottom: ${space[4]}px;
-  }
+  margin-bottom: ${space[4]}px;
 `;
 
 const topMarginStyles = css`
-  margin-top: ${space[2]}px;
+  margin-top: ${space[5]}px;
+  ${from.mobileMedium} {
+    margin-top: 14px;
```
ob6160, line 15: shame there's no space value for this

src/client/components/Header.tsx
```diff
@@ -2,64 +2,56 @@ import React from 'react';
...
-    margin-right: ${space[24]}px;
-  }
 `;
 
 const backgroundColor = css`
   background-color: ${brand[400]};
 `;
 
+const headerGridRightToLeft = css`
+  direction: rtl;
```
ob6160, line 27: hadn't seen that before, neat!

ob6160 (approved):
nice work, looking forward to a well aligned header 👍

src/client/components/Header.tsx
coldlink: It's actually used for displaying languages that write from right to left, but we can repurpose it here, for the header element only as there is no text displayed.

src/client/components/Header.tsx
coldlink: Yeah source was initially designed for multiples of 4. but it's fine to break away from it where necessary. In this case it was to match the same margin used on the dotcom header.

ob6160 (approved):
fine tuning to match DCR looks good, nice work!

PR #958, opened 2021-10-04
style changes to Terms component

rebecca-thompson:
## What does this change?
This implements style updates requested by design that are shared between the register and the sign-in page. Page specific design updates will be made in a subsequent PR

## Accessibility

### DIFF
--- a/src/client/components/Terms.tsx
+++ b/src/client/components/Terms.tsx
@@ -1,15 +1,22 @@
 import React from 'react';
 import { css } from '@emotion/react';
-import { space } from '@guardian/src-foundations';
 import { Link } from '@guardian/src-link';
 import { textSans } from '@guardian/src-foundations/typography';
+import { from } from '@guardian/src-foundations/mq';
+import { space } from '@guardian/src-foundations';
 
 const Text = ({ children }: { children: React.ReactNode }) => (
   <p
     css={css`
-      ${textSans.small()}
+      ${textSans.xxsmall()}
       margin-top: 0;
-      margin-bottom: ${space[4]}px;
+      margin-bottom: 0px;
+      ${from.mobileMedium} {
+        margin-bottom: 6px;
+      }
+      ${from.tablet} {
+        margin-bottom: ${space[2]}px;
+      }
     `}
   >
     {children}
@@ -26,7 +33,7 @@ const TermsLink = ({
   <Link
     subdued={true}
     cssOverrides={css`
-      ${textSans.small()}
+      ${textSans.xxsmall()}
     `}
     href={href}
   >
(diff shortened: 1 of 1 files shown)


src/client/components/PasswordInput.tsx
```diff
@@ -36,7 +35,6 @@ const textInputBorderStyle = (error?: string) => css`
   bottom: 0px;
   width: 100%;
   height: ${height.inputMedium}px;
-  margin-bottom: ${space[3]}px;
```
rebecca-thompson, line 39: The password eye on the password input field has an absolute position which takes this margin into account. By removing the bottom margin of this element we have more control over the spacing around this form field without having to worry about putting the eye out of  alignment

src/client/components/PasswordInput.tsx
```diff
@@ -100,7 +98,7 @@ const EyeSymbol = ({
     height: 30px;
     position: absolute;
     right: 5px;
-    bottom: 19px;
+    bottom: 7px;
```
rebecca-thompson, line 101: the absolute position minus the removed bottom margin (12px)

src/client/components/PasswordForm.tsx
```diff
@@ -64,6 +64,10 @@ const form = css`
   padding-top: ${space[3]}px;
 `;
 
+const passwordInput = css`
+  margin-bottom: ${space[3]}px;
+`;
```
rebecca-thompson, line 69: Adding the bottom margin to the password form, as it has been removed from the password input field (see below for details)

src/client/components/PasswordInput.tsx
```diff
@@ -144,7 +142,6 @@ export const PasswordInput = ({
         label={label}
         name="password"
         supporting={supporting}
-        css={textInput}
```
rebecca-thompson, line 147: This css also just adding a bottom margin

src/client/components/Divider.tsx
```diff
@@ -0,0 +1,141 @@
...
+          `;
+  }
+};
+
+const decideSize = (size: Sizes) => {
+  switch (size) {
+    case 'fit':
+      return css`
+        :before {
+          margin-left: 0;
```
ob6160, line 58: What do you think about this being a copy of the version that we have ready to go into source (https://github.com/guardian/source/pull/1012) instead so we just have to drop in the css overrides for the behaviour and remove this once it's been added to the package

src/client/components/Divider.tsx
```diff
@@ -0,0 +1,141 @@
...
+            margin-top: ${space[12]}px;
+          `;
+  }
+};
+
+const decideSize = (size: Sizes) => {
+  switch (size) {
+    case 'fit':
+      return css`
+        :before {
```
ob6160, line 57: What do you think about this being a copy of the version that we have ready to go into source (https://github.com/guardian/source/pull/1012) instead so we just have to drop in the css overrides for the behaviour and remove this once it's been added to the package

src/client/components/Terms.tsx
```diff
@@ -26,36 +33,47 @@ const TermsLink = ({
...
+      ${textSans.xxsmall()}
     `}
     href={href}
   >
     {children}
   </Link>
 );
 
+const terms = css`
+  margin-top: 12px;
```
ob6160, line 45: can these be `space[3]` and `space[4]`?

src/client/pages/SignIn.tsx
```diff
@@ -16,20 +14,42 @@ import { PageTitle } from '@/shared/model/PageTitle';
...
+  }
+`;
 
 const Links = ({ children }: { children: React.ReactNode }) => (
-  <p
+  <div
     css={css`
-      ${textSans.medium()}
-      margin-top: 0;
+      margin-top: 8px;
```
ob6160, line 48: can it be `space[2]`?

src/client/components/SocialButtons.tsx
```diff
@@ -18,16 +18,23 @@ const containerStyles = css`
...
   justify-content: center;
-  margin-top: ${space[12]}px;
-  margin-bottom: ${space[12]}px;
+  margin-top: ${space[5]}px;
+  ${from.mobileMedium} {
+    margin-top: ${space[6]}px;
+  }
+  margin-bottom: 60px;
+  ${from.desktop} {
+    margin-bottom: ${space[24]}px;
```
ob6160, line 27: Just thinking, if we come to use the social buttons in the future outside of this context, would the large margin here be desirable.

Thoughts on adding the margins to a wrapper instead?

ob6160 (commented):
nice work!

src/client/components/Terms.tsx
rebecca-thompson: 👍

src/client/components/SocialButtons.tsx
rebecca-thompson: I can't see a future scenario where these buttons would be used outside the context of sign-in/registration. If there is a future use case for these buttons perhaps it can be refactored at the time

src/client/components/Divider.tsx
rebecca-thompson: It might be simpler to keep this PR to the style updates and give the Divider updates its own PR

ob6160: okay, that makes sense to me

rebecca-thompson: I'm thinking leave it as is, in this PR and remove it and replace it with the source kitchen component in a subsequent PR once your changes have gone in. That way Helene can review the designs and the alignment will be correct

ob6160: alright, sounds good to me 🙂

PR #960, opened 2021-10-04
Connect registration flow to IDAPI.

ob6160:
## What does this change?
Connects the registration flow to the /guest endpoint in IDAPI, forwarding ref, refViewId and returnUrl as query params.

## How was this tested
* Mocked tests for the registration flow. (cypress/integration/mocked/register.spec.js)
* End to end tests for registration flow. (cypress/integration/ete/register.spec.js)


ob6160:
[Connect Gateway registration page with Identity for basic registration (m)](https://trello.com/c/6dVj3GyL/3328-connect-gateway-registration-page-with-identity-for-basic-registration-m)

src/client/pages/EmailSent.tsx
```diff
@@ -47,17 +71,26 @@ export const EmailSent = ({
           <PageHeader>Check your email inbox</PageHeader>
           <PageBody>
             {email ? (
-              <PageBodyText>We’ve sent an email to {email}.</PageBodyText>
+              <PageBodyText>
+                We’ve sent an email to <b>{email}</b>.
```
ob6160, line 75: Text bolded and subtitle changed to match Figma design 
Screenshot 2021-10-05 at 11 26 48

ob6160: Also removed some `<br/>`s to match the design spec margins

QuarpT (commented):
looks very good. Just to check, will you be releasing guest registration behind recaptcha? I don't see any recaptcha verification here. It would be good to prevent users from spamming guest creation on IDAPI which could trigger global rate limits

ob6160:
> looks very good. Just to check, will you be releasing guest registration behind recaptcha? I don't see any recaptcha verification here. It would be good to prevent users from spamming guest creation on IDAPI which could trigger global rate limits

Thanks for the feedback 🙂, when we merge this it shouldn't be reachable until we configure fastly to point at it. I'm going to open a followup PR to this which will integrate it with #961 after both are merged

QuarpT:
> Thanks for the feedback 🙂, when we merge this it shouldn't be reachable until we configure fastly to point at it. I'm going to open a followup PR to this which will integrate it with #961 after both are merged

I think the load balancer is open to the internet, so might be worth disabling to begin with https://github.com/guardian/gateway/blob/main/cloudformation.yaml#L337

ob6160:
> > Thanks for the feedback 🙂, when we merge this it shouldn't be reachable until we configure fastly to point at it. I'm going to open a followup PR to this which will integrate it with #961 after both are merged
> 
> I think the load balancer is open to the internet, so might be worth disabling to begin with https://github.com/guardian/gateway/blob/main/cloudformation.yaml#L337

Aha, I didn't know about this - let's catch up about it tomorrow?

QuarpT:
> > > Thanks for the feedback 🙂, when we merge this it shouldn't be reachable until we configure fastly to point at it. I'm going to open a followup PR to this which will integrate it with #961 after both are merged
> > 
> > 
> > I think the load balancer is open to the internet, so might be worth disabling to begin with https://github.com/guardian/gateway/blob/main/cloudformation.yaml#L337
> 
> Aha, I didn't know about this - let's catch up about it tomorrow?

anytime

PR #961, opened 2021-10-04
feat: implement a useRecaptcha hook to integrate with the Google Reca…

ob6160:
`useRecaptcha` is a custom hook implemented to handle adding Google Recaptcha to Gateway.

It attaches the Google recaptcha script to the document if it does not already exist when mounted.

The hooks' function signature is defined as follows:
```typescript
export type CaptchaSize = 'invisible' | 'compact' | 'normal';
type UseRecaptcha = (
  siteKey: string,
  renderElement: HTMLDivElement | string,
  size?: CaptchaSize,
) => {
  executeCaptcha: () => void;
  token: string;
  error: boolean;
  expired: boolean;
  widgetId: number;
};
```

## How is it tested?
This is unit tested using react-testing-library and react-hooks-testing-library.
These two libraries will allow us to render our components in isolation and make assertions, opening up the possibility to unit test more complicated hooks and components in the future where we feel it's necessary.

## Accessibility


src/client/app.tsx
```diff
@@ -43,6 +43,9 @@ export const App = (props: ClientState) => {
           * {
             box-sizing: border-box;
           }
+          .grecaptcha-badge {
```
ob6160, line 46: Badge is hidden for Gateway, because we're using the legal text to cover this

ob6160:
[Add reCaptcha to the new registration page](https://trello.com/c/Y4ZFlHbH/3400-add-recaptcha-to-the-new-registration-page)

src/client/lib/hooks/useRecaptcha.tsx
```diff
@@ -0,0 +1,167 @@
...
+  url = 'https://www.google.com/recaptcha/api.js?render=explicit',
+) => {
+  const [loaded, setLoaded] = React.useState(false);
+
+  React.useEffect(() => {
+    const scriptExists =
+      typeof document !== undefined &&
+      document.getElementById('g-captcha-script');
+
+    if (scriptExists || recaptchaReady()) {
```
QuarpT, line 26: is it possible for the script to exist before recaptchaReady ?
This might cause problems if the hook is used more than once on a page 🤔

ob6160: good point, I'll have a think about that and get back to you

src/client/lib/hooks/useRecaptcha.tsx
```diff
@@ -0,0 +1,167 @@
...
+      setLoaded(true);
+      return;
+    }
+    const script = document.createElement('script');
```
sndrs, line 30: would be good to use `loadScript` from https://github.com/guardian/libs here if possible?

QuarpT:
this looks very good. Are you going to raise another PR to test the recaptcha code on the server?

src/client/lib/hooks/useRecaptcha.tsx
ob6160: interesting - I wasn't aware that this existed. I'll look into it :)

ob6160:
> this looks very good. Are you going to raise another PR to test the recaptcha code on the server?

thanks! Yep, in my PR to integrate the Gateway registration flow (coming up soon) I'll write an end to end test that covers this. Could also do an individual unit test for the controller that's handling it too

PR #963, opened 2021-10-05
feat(nav): update nav to match grid using flexbox

coldlink:
## Accessibility


src/client/components/Nav.stories.tsx
```diff
@@ -108,3 +110,43 @@ NoBreakpoint.parameters = {
   },
   chromatic: { disable: true },
 };
+
+export const WithGrid = () => {
```
coldlink, line 114: Added this in so that it is easier to compare alignment to the grid

PR #964, opened 2021-10-06
Fix scroll issues by setting min-height on the app

coldlink:
## What does this change?
Change the height to a min-height on the `#app` so that on smaller screens there is a scroll.

#### Before
IMG_3386

#### After 
Screenshot 2021-10-06 at 14 46 06


ob6160 (approved):
🏆

PR #1060, opened 2021-10-27
Set password page (with changes to reset/change password page)

coldlink:
## What does this change?
- Add the `/set-password` page routes, allowing a user to create a password
  - This is similar to the reset password flow, but for users without an existing password
  - We do this by allowing the `ChangePassword` and `ChangePasswordComplete` components to pass in any text as props
  - This flow is used in registration, where an existing user, without a password set, attempts to register
- Update the copy on the reset and set password pages, and confirmation pages
- Add `rel="noreferrer"` to external links
- Redirect to set/reset password confirmation page, rather than rendering the page to avoid token leaking
- Update cypress local scripts to only show tests for that environment (mocked/ete)

## Flow
chrome_Fso0r2Bkl1

## TODO
- [x] IDAPI PR to redirect users to set password page for guest users, and from create password email
  - https://github.com/guardian/identity/pull/2005
- [x] Update the VCL configuration to make set password page available to gateway
  - https://github.com/guardian/identity-platform/pull/462 
- [x] Add cypress tests after IDAPI PR has gone in

#### Set Password Page
localhost_6006_iframe html_id=pages-changepassword--set-password args= viewMode=story(Pixel 2)

#### Set Password Complete Page
localhost_6006_iframe html_id=pages-changepasswordcomplete--set-password args= viewMode=story(Pixel 2)


src/client/pages/ChangePasswordPage.tsx
```diff
@@ -20,12 +20,14 @@ export const ChangePasswordPage = () => {
   // if the token expires while the user is on the current page
   if (typeof window !== 'undefined' && tokenExpiryTimestamp) {
```
QuarpT, line 21: it's probably only executing this on the first render `useEffect(..., [])`

src/server/routes/setPassword.ts
```diff
@@ -0,0 +1,150 @@
...
+      await sendCreatePasswordEmail(email, req.ip, returnUrl);
+
+      setEncryptedStateCookie(res, {
+        email,
+        emailType: EmailType.CREATE_PASSWORD,
+      });
+
+      return res.redirect(303, `${Routes.SET_PASSWORD_EMAIL_SENT}`);
+    } catch (error) {
+      const { message, status } = error as RequestError;
```
QuarpT, line 81: what if the error is not a RequestError?
If a standard JS error is caught here, then it could have a message with status undefined

QuarpT:
looks very good!

src/client/pages/ChangePasswordPage.tsx
coldlink: good spot! I'll wrap these in a `useEffect`

src/server/routes/setPassword.ts
coldlink: added some defaults in to return to the user, logs will show true error if needed

QuarpT: We should try to avoid surfacing technical error messages to users. Perhaps the handled api errors should use a JavaScript class instead of a `type`, so we can check the error class name at runtime.

QuarpT (approved):

src/server/routes/setPassword.ts
coldlink: We're not surfacing any technical error messages to users, just a generic "We had trouble resetting your password". The `logger `method will log the actual error to kibana if we need to read it.

I do agree though that it would be useful to refactor the errors in the long term, especially to differentiate between javascript runtime errors, and api errors from idapi/okta in the future.

coldlink: Note: Errors were refactored in https://github.com/guardian/gateway/pull/1104

src/client/components/Footer.tsx
```diff
@@ -146,24 +146,40 @@ export const Footer = () => (
         <div css={ulWrapperStyles}>
           <ul css={[ulStyles, columnStyles]}>
             <li>
-              <Link cssOverrides={linkStyles} href={locations.PRIVACY}>
+              <Link
```
dominickendrick, line 149: What do you think about a `<ExternalLink>` component that adds `rel=noreferrer` 🤔  Would this make it more explicit that you were linking off site ? Could we also update the namespace in the routes, ie `locations.external.CONTACT_US` ?

coldlink: I like that idea! I'll make a separate PR for that suggestion.

coldlink: https://github.com/guardian/gateway/pull/1107

PR #1061, opened 2021-10-27
Email Input component

ob6160:
_(This is part of a series of PRs updating the error messages across the Sign In and Registration pages)_

## Motivation
We'd like the ability to display errors when the email field is empty or an invalid email has been entered on the **Sign In** and **Registration** pages.

## What does this change?
This introduces an `EmailInput` component with the intention that it can be reused across pages where an email input is required that carries out validation checks on blur (when it loses focus).

Currently our requirements state that we should handle cases where the input is **empty** or has an **invalid email**. The invalid email check is carried out using the HTML5 Constraint validation process, using the `invalid` event [^1]. 

The component is rendered and its behaviour is unit tested using react-testing-library and the jest-dom extension [^2]. It was also added to Storybook.

IE11 support for the HTML Constraint Validation API is fine for our use case. I've tested it in my IE11 vm and all seems fine. Note: _`badInput` and `tooShort` are not supported in IE11, but these flags are defaulted to `false` so this means the field won't be marked invalid when it is valid_ [^3].

### Empty email input
image

### Invalid email (not a real email)
image

### Empty email input
Screenshot 2021-10-27 at 17 14 48

### Invalid email (not a real email)
Screenshot 2021-10-27 at 17 15 15

[^1]: https://developer.mozilla.org/en-US/docs/Web/Guide/HTML/Constraint_validation#constraint_validation_process
[^2]: https://testing-library.com/docs/react-testing-library/example-intro
[^3]: https://caniuse.com/constraint-validation

### DIFF
--- a/src/client/__tests__/EmailInput.test.tsx
+++ b/src/client/__tests__/EmailInput.test.tsx
@@ -5,7 +5,7 @@
 import React from 'react';
 import { render, fireEvent } from '@testing-library/react';
 import '@testing-library/jest-dom/extend-expect';
-import EmailInput from '../EmailInput';
+import { EmailInput } from '../components/EmailInput';
 
 const setup = () => {
   const utils = render(<EmailInput defaultValue="test@email.com" />);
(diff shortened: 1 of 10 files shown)


src/client/__tests__/EmailInput.test.tsx
```diff
@@ -0,0 +1,81 @@
+/**
```
coldlink, line 1: loving these tests, i wonder if we need to spend some time adding them in for existing components too?

coldlink (approved):

src/client/__tests__/EmailInput.test.tsx
ob6160: they're fun to write :) yes! Components like `PasswordInput` with a bit of state might be good candidates. I could put together a list

ob6160: even without state, they're useful in a lot of cases tbh

ob6160: Maybe a good dev standup topic!

PR #1070, opened 2021-10-29
Cypress typescript & tidyup (10% time)

ob6160:
## What does this change?
This 10% time I looked at improving our  Cypress development experience by converting the sub-project to typescript.

I've done my best to condense my changes into a set of bullet points below:
* Renamed all `.js` files to `.ts` in the `cypress/` directory.
  * Changed our Github action config to reflect the file ext change.
* Turned on strict type checking for the cypress project and fixed all type errors.
* Split out our custom cypress commands into individual files for an improved experience adding new ones. 
  * Each command extends the `Cypress` namespace to add their own typings.
  * I've documented each command to the best of my knowledge
* Defined a custom `tsconfig.json` for our cypress test suite as it has different requirements to our core tsconfig.
  * One key reason was to avoid types for jest clashing with those that cypress includes [^1]
* Added new `yarn tsc:cypress` and `yarn watch:cypress` commands, so that we can develop our cypress tests.
  * Added `yarn tsc:cypress` as another husky push requirement so that we make sure type checking happens.
* Added a new command to help with email testing `getEmailDetails`: https://github.com/guardian/gateway/pull/1070/files#diff-f07a49b8a432bd055c3eb8928bd6014232513e85953308603d9da7d21849d9c5R10
* Use the `uuid4` package instead of relying on the in-browser `window.crypto` object which has mixed browser support and no types
 

[^1]: https://docs.cypress.io/guides/tooling/typescript-support#Clashing-types-with-Jest


cypress/integration/ete/registration/register.spec.ts
```diff
@@ -111,17 +111,18 @@ describe('Registration flow', () => {
         receivedAfter: timeRequestWasMade,
       },
     ).then((email) => {
-      const body = email.html.body;
+      const id = email.id ?? '';
```
QuarpT, line 114: it would be worth failing the test early if email.id is undefined.
You could do the same with body, rather than rely on the assertions failing against empty string

ob6160: a lot more explicit, I like that idea. Will change and resolve the conflicts :)

ob6160: I've refactored the email data extraction into its own dedicated command, everything is a lot cleaner now! (see cypress/support/commands/getEmailDetails.ts)

cypress/integration/ete/registration/register.spec.ts
```diff
@@ -216,35 +214,29 @@ describe('Registration flow', () => {
         receivedAfter: timeRequestWasMade,
       },
     ).then((email) => {
-      const body = email.html.body;
-
-      const links = email.html.links ?? [];
+      cy.getEmailDetails(email, /\/reset-password\/([^"]*)/).then(
```
QuarpT, line 217: do we always do this after mailosaurGetMessage?
it might be worth combining the two 🤔

QuarpT: also, de we always call delete at the end?
could be worth combining that too 😄

ob6160: both good ideas, I feel like combining them would need a bit more thought and would be a nice followup PR I could open after!

QuarpT (approved):
looks great!

PR #1071, opened 2021-11-01
Fix the reset password test

ob6160:
## What does this change?
Changes reset password test after the path was changed in the reset password email from `c` to `reset-password` so that our e2e tests pass.


ob6160:
closed in favour of #1072 because after a rebase the e2e test disappeared from the status icons

PR #1100, opened 2021-11-01
Update dependencies

ob6160:
## What does this change?
Brings dependencies up to date
* Adds web vitals peer dep (needed since https://github.com/guardian/libs/pull/280)
* Not upgrading babel to `7.16.0` due to multiple issues with Storybook and module transpilation. Will be investigated separately


coldlink (approved):

package.json
```diff
@@ -107,20 +108,20 @@
...
-    "@guardian/src-text-input": "^3.10.1",
+    "@guardian/libs": "^3.5.0",
+    "@guardian/source-react-components-development-kitchen": "^0.0.20",
+    "@guardian/src-accordion": "^3.10.2",
+    "@guardian/src-brand": "^3.10.2-alpha.1",
+    "@guardian/src-button": "^3.10.2",
+    "@guardian/src-checkbox": "^3.10.2",
+    "@guardian/src-foundations": "^3.10.2-alpha.1",
+    "@guardian/src-icons": "^3.10.2",
+    "@guardian/src-layout": "3.10.2-alpha.1",
```
coldlink, line 119: we shouldn't be bringing alpha packages into production, i suggest leaving this at `3.10.1` for now

coldlink (changes requested):

package.json
```diff
@@ -91,12 +91,13 @@
...
     "prettier": "^2.4.1",
     "react-test-renderer": "^17.0.2",
-    "snyk": "^1.745.0",
+    "snyk": "^1.749.0",
     "stmux": "^1.8.1",
     "ts-jest": "^27.0.7",
     "typescript": "^4.4.4",
     "wait-on": "^6.0.0",
-    "webpack": "^5.59.1",
+    "web-vitals": "^2.1.2",
```
coldlink, line 99: instead of adding this as a dependency, could we add it as a peerDependency instead?

coldlink (approved):

PR #1104, opened 2021-11-02
Differentiate between API/IDAPI errors and JS runtime errors

coldlink:
We don't want to propagate runtime JavaScript error messages to the user, as was potentially occurring when destructing the `error` object in many of our routes, and instead provide a generic message and status back to the user.

To differentiate Api/Idapi errors from runtime JavaScript error, we create a new `ApiError` class and `IdapiError` (which extends `ApiError`) class, and replaces the existing `RequestError` interface.

Then when we want to check the `message` and `status` from the error we can do `error instanceOf ApiError` to determine if it's an `ApiError` or `IdapiError`. If it's neither then we instantiate an empty `ApiError` class, which returns a generic message.

We'll still be able to view runtime errors as we log out the whole error object when an error is caught.


src/server/models/Error.ts
```diff
@@ -0,0 +1,38 @@
...
+    status = 500,
+    field,
+  }: {
+    message?: string;
+    name?: string;
+    status?: number;
+    field?: string;
+  } = {}) {
+    super();
+    this.message = message;
```
QuarpT, line 19: will this be identical to `super(message);` ?

QuarpT (approved):
looks great!

src/server/models/Error.ts
coldlink: ah yes! good point, let me refactor slightly

PR #1105, opened 2021-11-03
Adds an eslint plugin for rules of hooks

ob6160:
## What does this change?
Adds an eslint plugin [^1] that ensures we are following the official React "rules of hooks" as defined here: https://reactjs.org/docs/hooks-rules.html 

It adds a helpful set of intellisense suggestions to vscode for when we make hooks that go against these rules. For example, we need to add a dependency here it shows us in a hint:

Screenshot 2021-11-18 at 17 17 01

Fixes all warnings/errors that the plugin throws up. Includes a small change to the test for going online/offline during the reCAPTCHA test.

[^1]: https://www.npmjs.com/package/eslint-plugin-react-hooks

### DIFF
--- a/src/client/lib/hooks/useRecaptcha.tsx
+++ b/src/client/lib/hooks/useRecaptcha.tsx
@@ -50,7 +50,7 @@ const useRecaptchaScript = (src: string) => {
     recaptchaScriptLoadPromise
       .then(initialiseRecaptcha)
       .catch(() => setError(true));
-  }, []);
+  }, [src]);
 
   return {
     loaded,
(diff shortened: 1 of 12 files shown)


src/client/lib/hooks/useRecaptcha.tsx
```diff
@@ -118,14 +118,7 @@ const useRecaptcha: UseRecaptcha = (
...
   if (siteKey === '' || renderElement === '') {
-    return {
-      token,
-      error: true,
-      expired,
-      widgetId,
-      executeCaptcha: () => null,
-      requestCount,
-    };
+    throw new Error('No site key or render element passed');
```
ob6160, line 121: I've changed this to throw an error now, because by returning early we were running the `useRecaptchaScript` conditionally on re-render, which goes against the rules of using hooks

ob6160: also updated the test to check for the exception occuring

coldlink (approved):

PR #1107, opened 2021-11-03
Add `ExternalLink` and `ExternalLinkButton` components

coldlink:
We want to apply `noreferrer` and `noopener` on external links for security reasons.

We could easily apply these to the `rel` attribute on the source `Link`/`LinkButton` component, but to be more explicit that the link is pointing externally, we create the `ExternalLink` and `ExternalLinkButton` components that extends the link component with the attribute already applied.


ob6160 (approved):
nice 🐻‍❄️

PR #1109, opened 2021-11-04
Refactor how we do `QueryParams`

coldlink:
## What does this change?

We seem to have trouble when reading query parameters on the client side of gateway. Often passing them in as extra properties on page data, or attempting to read them using `location.search`/`useLocation` etc.

We're also don't differentiate between parameters that can persist from request to request, and those that should only be available for a single request.

By splitting out the `QueryParams` type into `SafeQueryParams` and `UnsafeQueryParams` it is possible to differentiate between these two cases, safe meaning they can persist, and unsafe meaning they'll only be available for a specific route. The `QueryParams` type is then a joining of the two.

The `getSafeQueryParams` method then can then be used to extract only the `SafeQueryParams` from a `QueryParams` object, which is useful in cases where we want to turn the query params object into a query string and not include anything that has been deemed unsafe.

Since we do this "strigification" often the `addQueryParamsToPath` method helps to facilitate this. By default it takes a `path`/string to append query parameters onto, determines the correct divider (`?` or `&`), and takes the QueryParams, extracts the `SafeQueryParams` using `getSafeQueryParams` and appends this with the divider to the string/path. An optional 3rd parameter allows the setting of any `UnsafeQueryParams` if needed.

This method is useful for redirects on the server, and appended to the form action url on the client.

In other parts, by adding `queryParams` as a property on the `ClientState` it would make it easier to read and use the query parameter properties on the client similar to how we read `pageData`, and saves us from using disparate methods on the client to read any modify the query parameters, for example the mess that existed when using `returnUrl`, `ref` and `refViewId`, allowing us to remove the `buildRegistrationUrlQueryParamString` and other instances where we pass `ref` and `refViewId` on the client.

The readme has also been updated to improve the section on query parameters.

## Other changes
- Fixes ts-jest memory leak issue: https://github.com/guardian/gateway/pull/1109#pullrequestreview-798234477
- Updates and fixes cypress tests and configuration


jest.config.js
```diff
@@ -6,4 +7,9 @@ module.exports = {
     '@/([^\\.]*)$': '<rootDir>/src/$1',
   },
   testPathIgnorePatterns: ['<rootDir>/cypress/', '<rootDir>/node_modules/'],
+  globals: {
+    'ts-jest': {
+      isolatedModules: true,
+    },
+  },
```
coldlink, line 14: This fixes a jest memory leak when using typescript.

Before:
Code_w1BKfr265f

After:
Code_FXU7EN3REx

dominickendrick: 🥇

src/server/lib/queryParams.ts
```diff
@@ -71,12 +76,28 @@ export const addReturnUrlToPath = (path: string, returnUrl: string): string => {
...
-  return `${path}${divider}ref=${encodeURIComponent(ref)}`;
+  const safeQueryString = stringify(getSafeQueryParams(params), {
+    skipNull: true,
+    skipEmptyString: true,
+  });
+  const unsafeQueryString = stringify(unsafeParams || {}, {
+    skipNull: true,
+    skipEmptyString: true,
+  });
+  return `${path}${divider}${safeQueryString}${
```
dominickendrick, line 100: is it worth using [`UrlSearchParams`](https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams) here

src/server/lib/middleware/login.ts
```diff
@@ -8,40 +8,46 @@ import { trackMetric } from '@/server/lib/trackMetric';
...
+        returnUrl: encodeURIComponent(joinUrl(profileUrl, req.path)),
+      });
       return res.redirect(redirect);
     }
 
-    return res.redirect(generateRedirectUrl(LOGIN_REDIRECT_URL));
+    return res.redirect(
+      addQueryParamsToPath(LOGIN_REDIRECT_URL, {
+        ...res.locals.queryParams,
+        returnUrl: encodeURIComponent(joinUrl(profileUrl, req.path)),
```
dominickendrick, line 36: Is it worth allowing the `addQueryParamsToPath` function to deal with the `encodeURIComponent`, This happens by default with [`qs.stringify` function](https://github.com/ljharb/qs#stringifying)

src/server/lib/queryParams.ts
dominickendrick: Ignore me ;)

jest.config.js
coldlink: Found the fix in this thread: https://github.com/kulshekhar/ts-jest/issues/1967

src/server/lib/middleware/login.ts
coldlink: ah ha! good spot, the `encodeURIComponent` does not need to happen here!

src/client/pages/SignInPage.tsx
```diff
@@ -3,13 +3,21 @@ import { SignIn } from '@/client/pages/SignIn';
...
+    pageData = {},
+    globalMessage = {},
+    clientHosts,
+    queryParams,
+  } = clientState;
   const { returnUrl, email } = pageData;
   const { error } = globalMessage;
   const { oauthBaseUrl } = clientHosts;
+  const queryString = addQueryParamsToPath('', queryParams);
+  console.log(queryString);
```
coldlink, line 20:
```suggestion
```

src/client/components/ClientState.tsx
```diff
@@ -1,9 +1,12 @@
 import React, { createContext, FunctionComponent } from 'react';
 import { ClientState } from '@/shared/model/ClientState';
 
-export const ClientStateContext = createContext({
+export const ClientStateContext = createContext<ClientState>({
   clientHosts: { idapiBaseUrl: '', oauthBaseUrl: '' },
   recaptchaConfig: { recaptchaSiteKey: '' },
+  queryParams: {
+    returnUrl: 'https://theguardian.com',
```
QuarpT, line 8: can we set this based on the environment (code/prod)

src/client/pages/Registration.tsx
```diff
@@ -140,7 +120,7 @@ export const Registration = ({
       >
         <form
           method="post"
-          action={`${Routes.REGISTRATION}?${registrationUrlQueryParamString}`}
+          action={`${Routes.REGISTRATION}?${queryString}`}
```
QuarpT, line 123: It looks like queryString already includes the `?` separator 🤔

coldlink: Good spot!

```suggestion
          action={`${Routes.REGISTRATION}${queryString}`}
```

src/client/components/ClientState.tsx
coldlink: Unfortunately not, as this is on the client, it doesn't have access to the environment. However in most (if not all) cases this default value will not be reached as `returnUrl` is a required query parameter on every request.

QuarpT:
looks very good, makes it much easier to propagate url parameters on forms

an optional naming suggestion: 
SafeQueryParams/UnsafeQueryParams doesn't necessarily reflect the usage. Sometimes query parameters are 'safe', but we don't want them to be propagated. It might be more accurate to use the naming `PropagatedQueryParams` just to separate the params that always get propagated from the 'normal' ones. e.g.:

```
export interface PropagatedQueryParams extends StringifiableRecord {
  returnUrl: string;
  clientId?: string;
  ref?: string;
  refViewId?: string;
}

export interface QueryParams extends PropagatedQueryParams, StringifiableRecord {
    emailVerified?: boolean;
    csrfError?: boolean;
    encryptedEmail?: string;
    error?: string;
}
```

coldlink:
> looks very good, makes it much easier to propagate url parameters on forms
> 
> an optional naming suggestion: SafeQueryParams/UnsafeQueryParams doesn't necessarily reflect the usage. Sometimes query parameters are 'safe', but we don't want them to be propagated. It might be more accurate to use the naming `PropagatedQueryParams` just to separate the params that always get propagated from the 'normal' ones. e.g.:
> 
> ```
> export interface PropagatedQueryParams extends StringifiableRecord {
>   returnUrl: string;
>   clientId?: string;
>   ref?: string;
>   refViewId?: string;
> }
> 
> export interface QueryParams extends PropagatedQueryParams, StringifiableRecord {
>     emailVerified?: boolean;
>     csrfError?: boolean;
>     encryptedEmail?: string;
>     error?: string;
> }
> ```

I like that suggestion! I'll implement that.

QuarpT (approved):

PR #1110, opened 2021-11-05
Refactoring PR to clean up routing

rebecca-thompson:
## What does this change?
This is a refactoring PR which aims to create more consistency in the definition of our url paths. It:

- refactors the `Routes` enum so that each route is only one `/` deep and paths are then defined by combinations of these routes, eg `${Routes.REGISTRATION_EMAIL_SENT}` becomes `${Routes.REGISTRATION}${Routes.EMAIL_SENT}`. This was already happening in parts of the code, but it wasn't applied consistently in all situations.
- replaces IDAPI routes defined as strings in methods, with routes defined in the `ApiRoutes` enum.
- removes `/signin/current` route. This is a legacy from the old system and should be handled in VCL config
- changes the `POST /register/email-sent/resend` to `POST /register/resend` to make it consistent with other resend endpoints. The registration code is not released yet, so this is a safe change.


rebecca-thompson:
[Refactoring gateway routes](https://trello.com/c/hCrRDbvX/3547-refactoring-gateway-routes)

src/client/pages/RegistrationEmailSentPage.tsx
```diff
@@ -12,8 +13,8 @@ export const RegistrationEmailSentPage = () => {
     <EmailSent
       email={email}
       subTitle={"We've sent you an email"}
-      previousPage={'/signin'}
-      resendEmailAction={'/register/email-sent/resend'}
+      previousPage={`${Routes.SIGN_IN}`}
+      resendEmailAction={`${Routes.REGISTRATION}${Routes.RESEND}`}
```
dominickendrick, line 17: Shouldn't this be `${Routes.REGISTRATION}${EMAIL_SENT}${Routes.RESEND}` ?

rebecca-thompson: there's a comment about this in the pr description

dominickendrick: 👍

ob6160 (approved):
looks good. as discussed I think strongly typing routes using template literal types would be a nice improvement to make in the future

coldlink (approved):

PR #1111, opened 2021-11-06
Use test user endpoint in Cypress

ob6160:
## What does this change?
Uses an improved `/user/test` endpoint to generate test users at runtime during our e2e tests.

To do this, a new `createTestUser` command was implemented [^1]. This command accepts the following options:

```
type IDAPITestUserOptions = {
  primaryEmailAddress?: `${string}@${string}.mailosaur.net`;
  isUserEmailValidated?: boolean;
  socialLinks?: SocialLink[];
  password?: string;
  deleteAfterMinute?: boolean;
};
```

__Note: These tests will not pass until a companion PR in IDAPI (https://github.com/guardian/identity/pull/2011) which extends the `user/test` endpoint is released.__

The e2e test suite was then transitioned to use `createTestUser` where appropriate at the beginning of each test. By default, if nothing is passed for an email, a random email is created.

In order to clean up test users following each test run, an option was added to the IDAPI endpoint: `deleteAfterMinute`, if this is set to `true`, the IDAPI server will schedule a task for 60 seconds time which will delete the user.

It was implemented this way because it seemed more resilient than calling a delete endpoint from the cypress test itself, because our Cypress tests are often stopped midway through (during development for example), or if a runtime error happens.

A new helper command: `checkForEmailAndGetDetails` [^2] was also implemented. This wraps the `cypress.mailosaurGetMessage` method. This reduces some code repetition that developed after a while, as our use case was largely the same across each test.

[^1]: https://github.com/guardian/gateway/pull/1111/files#diff-ee6b852a30c06f7bab0b2ba2e9ba15fab814c5642b99ae33ddbb41d8904fe616R47
[^2]: https://github.com/guardian/gateway/pull/1111/files#diff-f07a49b8a432bd055c3eb8928bd6014232513e85953308603d9da7d21849d9c5R12

### DIFF
--- a/cypress/integration/ete/registration/register.spec.ts
+++ b/cypress/integration/ete/registration/register.spec.ts
@@ -119,7 +119,6 @@ describe('Registration flow', () => {
 
   it('sends user an account exists email for user with existing account with password trying to register, clicks sign in, taken to /signin', () => {
     cy.createTestUser({
-      password: 'test_password',
       isUserEmailValidated: true,
     })?.then(({ emailAddress }) => {
       cy.visit('/register');
@@ -157,7 +156,6 @@ describe('Registration flow', () => {
 
   it('sends user an account exists email for user with existing account with password trying to register, clicks reset password on email', () => {
     cy.createTestUser({
-      password: 'test_password',
       isUserEmailValidated: true,
     })?.then(({ emailAddress }) => {
       cy.visit('/register');
--- a/cypress/integration/ete/registration/register.spec.ts
+++ b/cypress/integration/ete/registration/register.spec.ts
@@ -198,6 +198,7 @@ describe('Registration flow', () => {
   it('sends user an account exists without password email for user with existing account without password trying to register, clicks create password on email', () => {
     cy.createTestUser({
       isUserEmailValidated: false,
+      isGuestUser: true,
     })?.then(({ emailAddress }) => {
       cy.visit('/register');
       const timeRequestWasMade = new Date();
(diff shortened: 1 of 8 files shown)


ob6160:
[Extend test users endpoint (M)](https://trello.com/c/3pY6gCAT/3204-extend-test-users-endpoint-m)

cypress/integration/ete/registration/register.spec.ts
```diff
@@ -119,182 +95,156 @@ describe('Registration flow', () => {
...
-        sentTo: existingWithPassword.email,
-      },
-      {
-        receivedAfter: timeRequestWasMade,
-      },
-    ).then((email) => {
-      cy.getEmailDetails(email, /\/reset-password\/([^"]*)/).then(
-        ({ links, body, id, token }) => {
+    cy.createTestUser({
+      password: 'test_password',
```
coldlink, line 122: Is it worth using a random strong password here which we don't know the value of?

maybe even something simple like a randomly generated uuid (`window.crypto.randomUUID()`)  value as the password for each test.

Just thinking about potential security implications of having test users floating around with weak passwords, even in a code/dev environment 🤔

coldlink (approved):
Looking really good! Just one comment around passwords.

cypress/integration/ete/registration/register.spec.ts
ob6160: good idea, not too hard to implement and potentially good payoff if it stops a potential vuln

ob6160: added in latest commit

PR #1112, opened 2021-11-08
8th November 2021 | Yet more dependency updates

coldlink:
## What does this change?

## Notes
- ~`react-router-dom` is a major update, so will move into a separate PR~
  - Updated `react-router-dom` in this PR and fix all breaking changes using this [upgrade guide](https://reactrouter.com/docs/en/v6/upgrading/v5#upgrade-to-react-router-v6)
  - Had to refactor how we handle `404`/not found routes as react router wasn't picking it up otherwise and causing a `502` error on the server

## Tested?
- [x] Local
- [x] CI
- [ ] CODE


src/client/components/PasswordInput.tsx
```diff
@@ -29,7 +33,7 @@ export const isDisplayEyeOnBrowser = (browserName: string | undefined) => {
 
 const textInputBorderStyle = (error?: string) => css`
   border: ${error
-    ? `4px solid ${textInputDefault.textInput.textError}`
+    ? `4px solid ${errorColors[400]}`
```
ob6160, line 36: I think think one can be `textInput.borderError` maybe

ob6160 (commented):
lgtm, one small comment on colour token

src/server/routes/index.ts
```diff
@@ -47,8 +48,10 @@ uncachedRoutes.use(verifyEmail);
 uncachedRoutes.use(magicLink);
 
 // welcome routes
-uncachedRoutes.use(noCache, welcome);
+uncachedRoutes.use(welcome);
 
 router.use(uncachedRoutes);
 
+router.use(notFound);
```
QuarpT, line 55: if res.locals gets passed to 404 then this should be uncached to avoid leaking sensitive data 🤔

QuarpT (dismissed):
worthing looking at this since it could be a potential vulnerability
https://github.com/guardian/gateway/pull/1112/files?diff=split&w=1#r744637606

src/client/routes.tsx
```diff
@@ -34,101 +34,106 @@ export type RoutingConfig = {
...
-    </Route>
-    <Route exact path={`${Routes.WELCOME}${Routes.EMAIL_SENT}`}>
-      <EmailSentPage />
-    </Route>
-    <Route exact path={`${Routes.WELCOME}${Routes.TOKEN_PARAM}`}>
-      <WelcomePage />
-    </Route>
+  <RouterRoutes>
+    <Route path={Routes.SIGN_IN} element={<SignInPage />} />
+    <Route path={Routes.REGISTRATION} element={<RegistrationPage />} />
```
QuarpT, line 39: the `exact` prop has been removed from some of these. Does this break anything?

coldlink: From the upgrade guide https://reactrouter.com/docs/en/v6/upgrading/v5#upgrade-to-react-router-v6:

<Route exact> is gone. Instead, routes with descendant routes (defined in other components) use a trailing * in their path to indicate they match deeply

coldlink: So from that reasoning every route is "`exact`" by default, and you use a `*` if it's not exact.

In gateway every route is exact, so it won't break anything now by removing it.

src/server/routes/index.ts
coldlink: Good spot!

src/client/routes.tsx
QuarpT: 👍

ob6160 (approved):

PR #1131, opened 2021-11-11
Okta Registration Proof of Concept - Pre-activation step

rebecca-thompson:
## What does this change?
This is a Proof of Concept to test our registration flow using Okta. It is sitting behind a feature flag and is intended to provide the mobile apps teams something to develop and test against. It is not intended to go into prod as it does not have end-to-end tests or Okta registration hooks to update IDAPI.

This is the first step (pre-activation) of our registration flow. The next stage is for IDAPI to send the email with the Okta activation token and for the welcome page to use the activation token to set a password (this will be in a subsequent PR).


rebecca-thompson:
[Build the Okta initial email registration step (pre-activation)](https://trello.com/c/WJHjprX9/3559-build-the-okta-initial-email-registration-step-pre-activation)

src/server/lib/okta/registration.ts
```diff
@@ -0,0 +1,33 @@
...
+import { sendOktaActivationEmail } from '@/server/lib/idapi/user';
+
+export const register = (
+  email: string,
+  ip: string,
+  returnUrl?: string,
+  refViewId?: string,
+  ref?: string,
+  activate = false,
+  sendEmail = false,
```
coldlink, line 12: if we always want these to be `false`, as i can't see a use case right now where we want to override these, i suggest that they shouldn't be parameters and be hardcoded as `false` where they are used

rebecca-thompson: I have extracted these into variables instead of arguments

src/server/lib/okta/registration.ts
```diff
@@ -0,0 +1,18 @@
+import { CreateUserRequestOptions } from '@okta/okta-sdk-nodejs/src/types/models/CreateUserRequest';
+import { oktaClient } from '@/server/lib/okta/client';
+import { UserActivationToken } from '@okta/okta-sdk-nodejs';
+
+export const register = (email: string): Promise<UserActivationToken> => {
```
dominickendrick, line 5: Could this function be called `registerWithOkta` to make it clear [here](https://github.com/guardian/gateway/pull/1131/files#diff-0862d4de5f45856338489930372d58873f6f0cf4ba5423c7b29e04e99d93d131R180) that we are registering with Okta ?

dominickendrick (commented):
Looks really good!

dominickendrick (approved):

src/server/lib/okta/registration.ts
rebecca-thompson: 👍

PR #1132, opened 2021-11-11
Remove unused code, exports, and tests

coldlink:
## What does this change?
- We've ran some AB tests in the past where we forgot to clean up the code after the tests have finished. This PR does just that.
- Also adds [`ts-prune`](https://github.com/nadeesha/ts-prune) to find unused code, and removed any that were not in use.


rebecca-thompson (approved):
looks good!

ob6160 (approved):

PR #1133, opened 2021-11-12
MVP1 Design Emails

coldlink:
## What does this change?

Replaces the old layout/designs of the emails with a new layout with the template and designs defined in the [figma file](https://www.figma.com/file/ZDPonnBXbMhV2smf51Dylc/?node-id=4175%3A126191), and updates all the email components and emails to use the new designs.

Documentation on the MJML tags is available at https://documentation.mjml.io/, we use [`mjml-react`](https://github.com/wix-incubator/mjml-react) as a wrapper.

### General Notes

- Each component is it's own [`MjmlSection`/`mjml-section`](https://documentation.mjml.io/#mj-section), essentially equivalent to a "row". We define the spacing between "rows" of components as the `padding` property on the `MjmlSection` component. We also use this to define the `backgroundColor` of the row, and whether the background should go `fullWidth` too.
- Each of the "rows" has a single column, defined using [`MjmlColumn`/`mjml-column`](https://documentation.mjml.io/#mj-column). There's no special properties/attributes on the column, and in our layouts there's only one column per row. So we're using this as a purely semantical component.
- Inside each "column" is where the component is defined with all the attributes that relate to that component, you see the docs for the relevant attribute information
- The `Page` component is used to define the rest of the rest of the markup that mjml requires, and each of the emails are built by adding the components are children to the `Page` component
- We also use standard fonts rather than Guardian fonts for two reasons
  - `font-face` isn't supported by most email clients
  - We don't have the license to use the fonts outside of `theguardian.com` domains
  - So after speaking to design, they came up with closely matching default system fonts

## How to test/review this PR

- Open the [figma file](https://www.figma.com/file/ZDPonnBXbMhV2smf51Dylc/?node-id=4175%3A126191) on the `✅ - Sign in & register emails /Q3/21` page to see expected designs
- Open the [latest storybook](https://60929d168594f80039336501-bizcvmzouh.chromatic.com/?path=/story/email-components-page--default) for this branch to see new designs, check the `Email` section in the left side bar.
- Open the [latest chromatic UI tests](https://www.chromatic.com/build?appId=60929d168594f80039336501&number=1466) to see what's changed

## TODO
- [x] Update the logo one new image png url is received from design

## After merge
- Raise PRs in IDAPI to use the new email templates by exporting the HTML from Gateway


src/email/components/SubHeader.tsx
```diff
@@ -7,28 +7,29 @@ type Props = { children: React.ReactNode };
...
+    >
       <MjmlColumn>
         <MjmlText
-          padding="4px 48px"
-          font-size="20px"
-          line-height="20px"
+          padding="4px 0px"
+          fontSize="20px"
+          lineHeight="23px"
+          letterSpacing="-2%"
```
ob6160, line 27: I've not seen this with a negative value before, just curious about what it does :)

ob6160 (approved):
nice work, looks good!

src/email/components/SubHeader.tsx
coldlink: it reduces the horizontal spacing between the characters https://developer.mozilla.org/en-US/docs/Web/CSS/letter-spacing

coldlink: we use it in this case to as closely as possible match the system default fonts to the guardian fonts, as we can't include guardian fonts in emails

ob6160: makes sense, thanks for the info

PR #1176, opened 2021-11-19
reCAPTCHA protection for routes that send emails

ob6160:
## What does this change?

Introduces a reCAPTCHA check for all pages where a user can send or resend an email.

### Server side changes

Currently, only the `/register` route enforces reCAPTCHA.

These changes also enforce reCAPTCHA on the following routes:

1. `/set-password/resend`
2. `/welcome/resend`
3. `/register/email-sent/resend`
4. `/reset`
5. `/magic-link` (unused right now)
6. `/verify-email`

Recaptcha was already enforced on this route

7. `/register`

### Client side changes

#### Pages where reCAPTCHA has been enabled

The table below shows a mapping between each page with reCAPTCHA enabled and the paths set out above.

The third column indicates where test coverage was added either through our end-to-end tests or our mocked tests.

| Page Component                  | Form Action Route                                                                  | Test Coverage                                                                                                                                                                                                                                                                                    |
| ------------------------------- | ---------------------------------------------------------------------------------- | ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------ |
| RegistrationPage                | `/register` (7)                                                                    | [e2e](https://github.com/guardian/gateway/blob/03ab2b44d6e072cd9b8149c1532f4e3a396b90c2/cypress/integration/ete/registration/register.spec.ts#L235), [mocked](https://github.com/guardian/gateway/blob/a1ad9cc1cd2bbbe83d5fc0ac48bb4c6bc005b0e1/cypress/integration/mocked/register.spec.js#L62) |
| RegistrationEmailSentPage       | `/register/email-sent/resend` (3)                                                  | [e2e](https://github.com/guardian/gateway/blob/0dd1cd3eb3a782336e89ac33f9947b6ed5a5dca4/cypress/integration/ete/registration/register_email_sent.spec.ts#L202)                                                                                                                                   |
| ResetPasswordPage               | `/reset` (4)                                                                       | [e2e](https://github.com/guardian/gateway/blob/ob/recaptcha-integration/cypress/integration/ete/reset_password.spec.ts#L26)                                                                                                                                                                      |
| ResetPasswordSessionExpiredPage | `/reset` (4)                                                                       | [e2e](https://github.com/guardian/gateway/blob/ob/recaptcha-integration/cypress/integration/ete/reset_password.spec.ts#L74)                                                                                                                                                                      |
| SetPasswordResendPage           | `/set-password/resend` (1)                                                         | [mocked](https://github.com/guardian/gateway/blob/99559966077afb9a9181cf8e2fd4623e69676ebf/cypress/integration/mocked/set_password.spec.ts#L109)                                                                                                                                                 |
| SetPasswordSessionExpiredPage   | `/set-password/resend` (1)                                                         | [mocked](https://github.com/guardian/gateway/blob/99559966077afb9a9181cf8e2fd4623e69676ebf/cypress/integration/mocked/set_password.spec.ts#L109)                                                                                                                                                 |
| WelcomeResendPage               | `/welcome/resend` (2)                                                              | [mocked](https://github.com/guardian/gateway/blob/d0c655b2a0e556e24da62d9e25873777fb862095/cypress/integration/mocked/welcome.spec.ts#L229)                                                                                                                                                      |
| WelcomeSessionExpiredPage       | `/welcome/resend` (2)                                                              | [mocked](https://github.com/guardian/gateway/blob/d0c655b2a0e556e24da62d9e25873777fb862095/cypress/integration/mocked/welcome.spec.ts#L229)                                                                                                                                                      |
| EmailSentPage                   | `/set-password/resend` (1), `/reset` (4), `/welcome/resend` (2), `/magic-link` (5) | [e2e](https://github.com/guardian/gateway/blob/0dd1cd3eb3a782336e89ac33f9947b6ed5a5dca4/cypress/integration/ete/registration/register_email_sent.spec.ts#L202)                                                                                                                                   |
| MagicLink (**unused**)          | `/magic-link` (5)                                                                  | not used, so no coverage yet                                                                                                                                                                                                                                                                     |
| ResendEmailVerificationPage     | `/verify-email` (6)                                                                | [mocked](https://github.com/guardian/gateway/blob/a07b31f03db5bc5862926192c935fc090727146a/cypress/integration/mocked/verify_email.spec.ts#L263)                                                                                                                                                 |

#### Implementation details

The `MainForm` component now accepts the following extra optional props:

```
  recaptchaSiteKey,          // reCAPTCHA will be enabled on the form if this is provided.
  setRecaptchaErrorMessage,  // A callback to set state if the form reports a captcha error
  setRecaptchaErrorContext,  // A callback to set state if the form reports captcha error context
```

The consumer of `MainForm` is responsible for rendering the **recaptchaErrorMessage** and **recaptchaErrorContext** states using the props described above.

If `recaptchaSiteKey` is truthy, `MainForm` renders a component [^1] which wraps the `useRecaptcha` hook. This component sets state in the form through a callback as the hook state is updated. [^2] The reasoning behind wrapping the hook like this is so that we can run it conditionally, as not all pages using `MainForm` will want to load the reCAPTCHA script from Google.

**How does MainForm carry out the reCAPTCHA check?**

**1.** When the user clicks the Submit button, the following submit handler runs `preventDefault()` and executes `executeCaptcha()`. This sends a request to Google for a reCAPTCHA token.

```typescript
/**
 * Executes the reCAPTCHA check.
 * Prevents the form from submitting until the reCAPTCHA check is complete.
 */
const handleSubmit = useCallback(
  (event: React.FormEvent<HTMLFormElement>) => {
    if (onSubmitOverride) {
      onSubmitOverride(event);
    }

    if (recaptchaEnabled && !recaptchaState?.token) {
      event.preventDefault();
      recaptchaState?.executeCaptcha();
    }
  },
  [onSubmitOverride, recaptchaEnabled, recaptchaState]
);
```

**2.** Once the reCAPTCHA request completes, the `useRecaptcha()` hook will set state internally. If it was successful, the `token` state is set. If it failed, the `error` or `expired` state is set.
**3.** If `token` is set in (2), `MainForm` will run the following useEffect block. This submits the form as the reCAPTCHA check was successful.

```typescript
/**
 * Submits the form once the reCAPTCHA check has been successfully completed.
 */
useEffect(() => {
  if (recaptchaEnabled) {
    const registerFormElement = registerFormRef.current;
    if (recaptchaState?.token) {
      registerFormElement?.submit();
    }
  }
}, [recaptchaEnabled, recaptchaState, recaptchaState?.token, registerFormRef]);
```

**4.** If `expired` or `error` is set in (2), `MainForm` will run the following useEffect block. This sets `recaptchaErrorMessage` state and optionally `recaptchaErrorContext` state if the request has failed more than once. This allows the consumer to act upon an error state and display an appropriate error message.

```typescript
useEffect(() => {
  if (recaptchaEnabled) {
    // Determine is something went wrong with the check.
    const recaptchaCheckFailed =
      recaptchaState?.error || recaptchaState?.expired;

    // Used to show a more detailed reCAPTCHA error if
    // the user has requested a check more than once.
    const showErrorContext =
      recaptchaCheckFailed && recaptchaState?.requestCount > 1;

    // Default to generic reCAPTCHA error message.
    // Show the retry message if the user has requested a check more than once.
    const recaptchaErrorMessage = showErrorContext
      ? CaptchaErrors.RETRY
      : CaptchaErrors.GENERIC;

    const recaptchaErrorContext = showErrorContext ? (
      <DetailedRecaptchaError />
    ) : undefined;

    // Pass the error states back to the parent component if setters are provided.
    if (recaptchaCheckFailed && setRecaptchaErrorMessage) {
      setRecaptchaErrorMessage(recaptchaErrorMessage);
    }
    if (showErrorContext && setRecaptchaErrorContext) {
      setRecaptchaErrorContext(recaptchaErrorContext);
    }
  }
}, [
  recaptchaEnabled,
  recaptchaState,
  recaptchaState?.error,
  recaptchaState?.expired,
  recaptchaState?.requestCount,
  setRecaptchaErrorContext,
  setRecaptchaErrorMessage,
]);
```

Because `MainForm` now deals with some state around the reCAPTCHA check, this PR includes a comprehensive unit test for the component [^3]. Broadly speaking, the test covers these key areas:

1. Check that the form is rendered
2. Check recaptcha error message is set
3. Check recaptcha error context is set
4. Check submit override is called if recaptcha is enabled
5. Check recaptcha is only enabled when a site key is provided
6. Check form can be submitted with recaptcha enabled

The Storybook entry for each page listed above was also updated to include a Story for the reCAPTCHA error state.

[^1]: https://github.com/guardian/gateway/pull/1176/files#diff-1fd1f792cf686a380fb7821007467a7727237394531d60422df3d9912e2d3302R177
[^2]: https://github.com/guardian/gateway/pull/1176/files#diff-6b2012d0cb55fc2f05610053970775217e250df05c01989f885eaab985f8c34cR157
[^3]: https://github.com/guardian/gateway/pull/1176/files#diff-85d552f1ffd31680477db35514cc1040290115ab7059422aacf328376949157e


coldlink:
I haven't started reviewing this yet, but we may also want reCAPTCHA on the sign in page too, at least while i was looking at the sign in page I saw the recaptcha text, but it's not implemented there.

Screenshot 2021-11-25 at 15 19 43

ob6160:
> I haven't started reviewing this yet, but we may also want reCAPTCHA on the sign in page too, at least while i was looking at the sign in page I saw the recaptcha text, but it's not implemented there.
> 
> Screenshot 2021-11-25 at 15 19 43

Thanks for the heads up - I've left it out of this ticket because in Trello it was specifically covering endpoints that send emails. It shouldn't be difficult to add in a followup PR though!

coldlink:
Great! I've put in a trello card for the recaptcha on the sign in page in Sprint to do, can pick it up once this goes in!

ob6160:
> Great! I've put in a trello card for the recaptcha on the sign in page in Sprint to do, can pick it up once this goes in!

thanks for sorting the ticket!

ob6160:
[Add recaptcha to all endpoints/pages/routes in gateway that send an email through IDAPI](https://trello.com/c/sMI3MK2p/3524-add-recaptcha-to-all-endpoints-pages-routes-in-gateway-that-send-an-email-through-idapi)

ob6160:
Note: I've missed adding reCAPTCHA to the `/verify-email` route. Will push up this change too and update the description.

src/client/components/MainForm.tsx
```diff
@@ -42,23 +58,111 @@ export const MainForm = ({
...
+  ]);
 
   return (
     <form
       css={formStyles}
       method="post"
       action={formAction}
-      onSubmit={onSubmitOverride}
+      onSubmit={handleSubmit}
+      ref={registerFormRef}
```
coldlink, line 154: the name of this is probably not right here, as `registerFormRef` implies it's to do with the registration form, rather than a generic form

I suggest maybe just using `formRef` as the name

src/server/routes/reset.ts
```diff
@@ -11,11 +11,19 @@ import { PageTitle } from '@/shared/model/PageTitle';
...
+import { RecaptchaV2 } from 'express-recaptcha';
+import { getConfiguration } from '@/server/lib/getConfiguration';
 
 const router = Router();
+const {
+  googleRecaptcha: { secretKey, siteKey },
+} = getConfiguration();
+
+// set google recaptcha site key
+const recaptcha = new RecaptchaV2(siteKey, secretKey);
```
coldlink, line 26: rather than having to initialise this in every file, could you make a file in `lib` with method that could export the `recaptcha` const?

src/server/routes/reset.ts
```diff
@@ -46,6 +55,17 @@ router.post(
...
+      if (req.recaptcha?.error) {
+        logger.error(
+          'Problem verifying recaptcha, error response: ',
+          req.recaptcha.error,
+        );
+        throw new ApiError({
+          message: CaptchaErrors.GENERIC,
+          status: 400,
+        });
+      }
```
coldlink, line 67: similar to the previous comment, maybe a method which takes `req.recaptcha`, and throws if there's an error?

src/client/layouts/Main.tsx
```diff
@@ -21,6 +21,7 @@ interface MainLayoutProps {
   pageTitle?: string;
   successOverride?: string;
   errorOverride?: string;
+  errorContext?: React.ReactNode | string;
```
coldlink, line 24: `ReactNode` can be a `string`, so don't need the explicit `string` type here, see here: https://stackoverflow.com/a/58123882.

```suggestion
  errorContext?: React.ReactNode;
```

src/client/components/PasswordForm.tsx
```diff
@@ -32,6 +32,11 @@ type Props = {
   labelText: string;
   // for grid layout on consents page
   gridAutoRow?: AutoRow;
+  recaptchaSiteKey?: string;
+  setRecaptchaErrorMessage?: React.Dispatch<React.SetStateAction<string>>;
+  setRecaptchaErrorContext?: React.Dispatch<
+    React.SetStateAction<ReactNode | string>
```
coldlink, line 38:
```suggestion
    React.SetStateAction<ReactNode>
```

coldlink (commented):
Looking good! I've just left a few small comments.

Might also be worth having someone else look at this too!

src/server/routes/reset.ts
ob6160: refactored recaptcha initialisation and this into its own method in `recaptcha.ts`

src/client/components/MainForm.tsx
ob6160: nice catch, missed this!

src/server/routes/reset.ts
ob6160: https://github.com/guardian/gateway/pull/1176#discussion_r757529383

src/server/routes/reset.ts
ob6160: https://github.com/guardian/gateway/blob/a7641a7305dfe1fb89bbe1d9dcbeb3299bb01fb3/src/server/lib/recaptcha.ts

ob6160:
> **Note:** I've missed adding reCAPTCHA to the `/verify-email` route. Will push up this change too and update the description.

added in: https://github.com/guardian/gateway/pull/1176/commits/cf101fb1a15964d304b699262d5b4116b52b5fe4, test in : https://github.com/guardian/gateway/pull/1176/commits/a07b31f03db5bc5862926192c935fc090727146a

src/client/components/MainForm.tsx
```diff
@@ -42,23 +58,111 @@ export const MainForm = ({
...
+      const registerFormElement = registerFormRef.current;
+      if (recaptchaState?.token) {
+        registerFormElement?.submit();
+      }
+    }
+  }, [
+    recaptchaEnabled,
+    recaptchaState,
+    recaptchaState?.token,
+    registerFormRef,
```
QuarpT, line 106: I don't think changes to `registerFormRef` will trigger a rerender
https://stackoverflow.com/a/60476525

ob6160: Interesting, good spot! I'll read into it.

src/client/lib/hooks/useRecaptcha.tsx
```diff
@@ -173,6 +173,43 @@ const useRecaptcha: UseRecaptcha = (
...
+      widgetId,
+    });
+  }, [
+    error,
+    executeCaptcha,
+    expired,
+    requestCount,
+    setRecaptchaState,
+    token,
+    widgetId,
```
QuarpT, line 208: what's the advantage of using 'useEffect' when everything is a dependency?

ob6160: good point, I can't see any reason why it's needed here

src/server/lib/recaptcha.ts
```diff
@@ -0,0 +1,36 @@
...
+    logger.error(
+      'Problem verifying recaptcha, error response: ',
+      recaptchaResponse.error,
+    );
+    throw new ApiError({
+      message: CaptchaErrors.GENERIC,
+      status: 400,
+    });
+  }
+};
```
QuarpT, line 32: can we add this to the middleware so we don't have to call checkRecaptchaError independently of adding the middleware for each route?

ob6160: hmm, I initially thought about adding it as middleware. the thing was that in each route we catch the thrown error and return the error to the rendered page like this: https://github.com/guardian/gateway/blob/3ccfa267589e4f0a8f6d095c8208109691557c97/src/server/routes/verifyEmail.ts#L118 I can't think of a clean way of doing the check outside of this try/catch block without shifting a lot of stuff around. I could be missing something here about how we could do it cleanly though?

src/client/lib/hooks/useRecaptcha.tsx
QuarpT: Ah that's a good reason, calling setRecaptchaState is probably triggering the rerender above this

src/server/lib/recaptcha.ts
QuarpT: Maybe we can do something similar to how csrf errors are handled

ob6160: ah I think I see what you mean - will get back to you soon

src/client/components/MainForm.tsx
ob6160: after looking into this, I don't think this is too critical to fix, as we don't change the ref after rendering the page. it's a pretty static ref just used to trigger submit on the form

src/server/lib/recaptcha.ts
ob6160: after pairing on this @QuarpT  and I moved reCAPTCHA error handling into its own query param called: `recaptchaError` which is caught in `errorHandler.ts` here: https://github.com/guardian/gateway/pull/1176/files#diff-ab4263f049cc27297065729715aa9cc12772fd4786a620419d124c0ace700d9fR36

To protect a route using this middleware you now need to use the exported array on the route: https://github.com/guardian/gateway/pull/1176/files#diff-1a6c97bd62b9062b06450a4be0bc71b232e70540be8dbbde0d6e194bd7901157R43

src/server/lib/middleware/errorHandler.ts
```diff
@@ -33,6 +33,21 @@ export const routeErrorHandler = (
...
     );
     return next(err);
+  } else if (err.code === 'EBADRECAPTCHA') {
+    res.redirect(
+      303,
+      addQueryParamsToPath(
+        req.url,
+        {
+          ...res.locals.queryParams,
+          returnUrl: defaultReturnUri,
```
QuarpT, line 43: we does returnUrl get overridden?
I don't know why it does this on EBADCSRFTOKEN

src/client/components/MainForm.tsx
QuarpT: > after looking into this, I don't think this is too critical to fix, as we don't change the ref after rendering the page. it's a pretty static ref just used to trigger submit on the form

It would be good to remove it from the list of dependencies unless there's a reason for it being there. It's a bit strange seeing a ref as a useEffect dependency 🤔

src/server/lib/middleware/errorHandler.ts
ob6160: good question! not sure why. on the face of it I can't see why we'd want to do that in the case of handling the recaptcha error.. although it'd be good to know why it was there for the csrf check

QuarpT:
looks very good - are you going to add recaptcha to signin?
(we currently rely on recaptcha on signin to prevent password spraying attacks against identity frontend)

src/server/lib/middleware/errorHandler.ts
QuarpT: @coldlink 36d5feb69c322f58e2792f7218a1f9ccce31ae2d

ob6160:
> looks very good - are you going to add recaptcha to signin? (we currently rely on recaptcha on signin to prevent password spraying attacks against identity frontend)

thanks 🙂 – yup, we've got that planned as a followup: https://trello.com/c/FAx8xduG/3579-add-recaptcha-to-the-sign-in-page

src/client/components/MainForm.tsx
ob6160: I don't think it should matter, yeah - I've removed it and added a comment above explicitly ignoring it so the hook linter passes https://github.com/guardian/gateway/pull/1176/commits/b44b46dfe6bb3162ef4003e40701a9ff5f1398e6#

package.json
```diff
@@ -54,6 +54,7 @@
     "@types/cookie-parser": "^1.4.2",
     "@types/csurf": "^1.11.2",
     "@types/express": "^4.17.13",
+    "@types/http-errors": "^1.8.1",
```
coldlink, line 57: you've added the types for this package, but i don't see the dependency itself, especially as I see the dependency being used later on

src/server/lib/renderer.tsx
```diff
@@ -67,9 +67,21 @@ const clientStateFromRequestStateLocals = ({
...
+
+  // checking if recaptcha error exists in query params, and attaching it to the
+  // forms field errors
+  if (queryParams.recaptchaError) {
+    return {
+      ...clientState,
+      globalMessage: {
+        ...clientState.globalMessage,
+        error: CaptchaErrors.GENERIC,
+      },
```
coldlink, line 80: can we use `deepmerge` here as we do elsewhere in the project?

package.json
coldlink: i assume it's being pulled in by another dependency, but since we're using it now best to add it to our dependencies too

coldlink (commented):
A few more small comments, looking great and nearly there!

src/client/lib/hooks/useRecaptchaTestUtils.ts
```diff
@@ -0,0 +1,47 @@
+// Used to mock the `load` and `error` events for scripts so the @guardian/libs `loadScript` method functions as intended.
```
coldlink, line 1: this file is related to the test right?

it would be worth adding to a `__tests__` directory inside the `hooks` folder to separate it from the non-test code. I also noticed there's a couple of other files in the `hooks` folder too which should also go into the `__tests__` folder

src/client/components/MainForm.tsx
QuarpT: ah if `exhaustive-deps` complains about this, then we're probably supposed to do it this way
sorry for making such a fuss

QuarpT (approved):

src/client/components/MainForm.tsx
ob6160: no problem, it was a good thing to point out regardless

package.json
ob6160: ah, great point! It was added so we can use types for the `createError` method imported from `http-errors`. I'll add it in as an explicit dep!

src/client/lib/hooks/useRecaptchaTestUtils.ts
ob6160: will do!

ob6160: moved all the relevant files into a `__tests__` folder

src/server/lib/renderer.tsx
ob6160: switched to using `deepmerge`

src/server/lib/middleware/errorHandler.ts
ob6160: thanks for explaining how this bit is working! Sounds like it's a tried and tested solution, I can make us a ticket to look into this if it doesn't need changing immediately here

ob6160:
Tested all the routes I could on CODE (magic-link not available), all looks good

src/server/lib/middleware/errorHandler.ts
coldlink: yeah it will only affect a small number of users who encounter CSRF errors, so we can add something to the backlog to look at it, it's not urgent :)

coldlink (approved):
💯

ob6160:
Thanks both for the thorough reviews 😊

I'm going to leave this PR up for the rest of today in case anyone else wants to look through, and merge tomorrow morning.

PR #1178, opened 2021-11-22
Okta Registration Proof of Concept - Post-activation step

rebecca-thompson:
## What does this change?
This is the second half of the Proof of Concept to register users in Okta. It uses the emailed activation token to set a password in Okta on the first step of the onboarding journey. The activation token is exchanged for a state token which is used to set the password. After a password is set, Okta returns a session token which can be passed into the `/authorize` endpoint to obtain access and refresh tokens via the OAuth Code Flow.


rebecca-thompson:
[Build the Okta post activation registration step (set password page etc)](https://trello.com/c/9uOfMahu/3564-build-the-okta-post-activation-registration-step-set-password-page-etc)

PR #1183, opened 2021-11-22
Upgrade Cypress to version 9.0.0

ob6160:
## What does this change?
Upgrades Cypress to version 9.0.0

Rebased onto #1182


coldlink (approved):
looks good! going through the changelog there doesn't seem to be anything that should affect us: https://docs.cypress.io/guides/references/changelog#9-0-0

PR #1187, opened 2021-11-22
MVP1: Geolocated logo in header

coldlink:
## What does this change?
- When a user is geolocated to the `GB`, show the `bestWebsite` logo in the header, if coming from anywhere else show the `anniversary` logo
  - We reuse the existing `geolocation` property from the `clientState` `pageData` object, in turn provided by the `x-gu-geolocation` header from fastly.

### GB
chrome_8DSnFKkTkD

### Not GB
chrome_ceM6OJ0lxZ

### DIFF
--- a/src/client/components/ConsentsHeader.tsx
+++ b/src/client/components/ConsentsHeader.tsx
@@ -3,15 +3,17 @@ import { GlobalError } from '@/client/components/GlobalError';
 import { GlobalSuccess } from '@/client/components/GlobalSuccess';
 import { Header } from '@/client/components/Header';
 import { getErrorLink } from '@/client/lib/ErrorLink';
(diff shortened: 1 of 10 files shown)


ob6160 (approved):
lgtm, TIL that fastly give us geolocation.. neat!

coldlink:
> lgtm, TIL that fastly give us geolocation.. neat!

Yeah it's super handy

See: https://github.com/guardian/gateway/blob/main/src/server/lib/getGeolocationRegion.ts
and https://github.com/guardian/identity-platform/blob/8de33d40f06a0f98a4fe6bb1609b294f0d6bfedc/fastly/profile-theguardian-com.vcl#L54

PR #1210, opened 2021-11-29
Refactor, update, and add additional metrics

coldlink:
## What does this change?

- Refactors `Metrics` from an `enum` to a `type` to make them typesafe using a combination of [template literal types](https://www.typescriptlang.org/docs/handbook/2/template-literal-types.html) and [union types](https://www.typescriptlang.org/docs/handbook/2/everyday-types.html#union-types)
- Adds `Register::Success` metric as it was included
- Adds `emailSendMetric` to track if an email send was successful or failed
- Removes the old `SEND_PASSWORD_RESET_SUCCESS` and `SEND_PASSWORD_RESET_FAILURE` metrics as these have been superseded using `emailSendMetric`, and they were not being used in existing alarms
- ~Adds `SigninInactivityAlarm` and `RegisterInactivityAlarm` based on the equivalent alarms in [identity-frontend](https://github.com/guardian/identity-frontend/blob/main/cloudformation/identity-frontend.yaml#L428-L492)~
  - Alarms moved to https://github.com/guardian/gateway/pull/1211

## Post Merge
- Update the [`identity-gateway` dashboard](https://eu-west-1.console.aws.amazon.com/cloudwatch/home?region=eu-west-1#dashboards:name=identity-gateway) with any metrics
- Update the [Business metrics dashboard](https://eu-west-1.console.aws.amazon.com/cloudwatch/home?region=eu-west-1#dashboards:name=Business-Metrics) with gateway metrics


ob6160 (approved):
looks good 👍

coldlink:
Actually I'm going to split out the alarms bit into it's own PR so we can merge the new metrics in now, and the alarms separately when we go live.

rebecca-thompson (approved):
As a future refactor it would be nice if all metric names were generated the same way, rather than some with the `::Success | ::Failure` suffixes in the enum string and others having it added based on passing in a boolean value.

Another nice to have would be the isSuccess argument being a `'success' | 'failure'` string value rather than a boolean, making it clearer what is happening when it is called from within a block of code.

rebecca-thompson (approved):
Looks great!

PR #1222, opened 2021-12-01
Emails: use px instead of pt

coldlink:
## What does this change?

- Fixes a validation issue where `pt` is not valid as a unit for `letter-spacing` in mjml, so using `px` after speaking to design
  - Design said that in this case they are equivalent enough
  - `Attribute letter-spacing has invalid value: -0.02pt for type Unit, only accepts (px, em) units and 1 value(s)`


ob6160 (approved):
📐📏

PR #1224, opened 2021-12-01
Make all routes type safe

dominickendrick:
This add type checking to all routes in the gateway project.

There are 5 main changes:

All routes are defined as string literals in `RoutePaths` and `ApiRoutePaths`

`buildUrl` and `buildUrlWithQueryParam` these functions only accept `RoutePaths` so you are guaranteed type safety when adding routes client side

Screen Shot 2021-12-01 at 16 57 15
And you will receive a type error if you enter an incorrect value
Screen Shot 2021-12-01 at 16 57 29

Type checking for routes also supports types for tokenised urls ie `'/reset-password/:token'`
The type checker would require an object of type { token: "value" } for token substitution

Screen Shot 2021-12-01 at 17 00 01

These types are defined by `ExtractRouteParams<T>` and `PathParam<P>`
 
`renderer` this also only accepts `RoutePaths` so you are guaranteed type safety when adding routes server side when rendering a page

`<TypedRoute>` component and the `typedRouter` helper only accept `RoutePaths` so `react-router` is type safe


src/client/layouts/ConsentsLayout.tsx
```diff
@@ -111,7 +110,11 @@ export const ConsentsLayout: FunctionComponent<ConsentsLayoutProps> = ({
         <ConsentsSubHeader autoRow={autoRow} title={title} current={current} />
         <form
           css={form}
-          action={`${Routes.CONSENTS}/${page}${queryString}`}
+          action={buildUrlWithQueryParams(
+            '/consents/:page',
+            { page: page },
```
coldlink, line 115:
```suggestion
            { page },
```

src/client/pages/ChangePasswordPage.tsx
```diff
@@ -30,7 +30,13 @@ export const ChangePasswordPage = () => {
     <ChangePassword
       headerText="Reset password"
       buttonText="Confirm new password"
-      submitUrl={`${Routes.CHANGE_PASSWORD}/${token}${search}`}
+      submitUrl={buildUrlWithQueryParams(
+        '/reset-password/:token',
+        {
+          token: token,
```
coldlink, line 36:
```suggestion
          token,
```

src/client/pages/Registration.tsx
```diff
@@ -20,14 +19,16 @@ import useRecaptcha, {
...
+import { QueryParams } from '@/shared/model/QueryParams';
 import { GeoLocation } from '@/shared/model/Geolocation';
 
 export type RegistrationProps = {
   returnUrl?: string;
   email?: string;
   recaptchaSiteKey: string;
   oauthBaseUrl: string;
-  queryString?: string;
+  queryString: QueryParams;
```
coldlink, line 31: not sure if the name of the prop makes sense anymore as it's not a string we're passing in now, it's an object, in this file and other places this is being used

my suggestion would be something like

```suggestion
  queryParams: QueryParams;
```

src/server/controllers/changePassword.ts
```diff
@@ -20,6 +20,10 @@ import {
 } from '@/server/lib/encryptedStateCookie';
 import { ApiError } from '@/server/models/Error';
 
+type ValidPassworUrls = '/reset-password' | '/set-password' | '/welcome';
+
+type ValidResendEmailUrls = '/reset' | '/welcome' | '/set-password';
```
coldlink, line 25: I suggest inheriting from `RouteProps` here, using `Extract` for even more type safety as it has to get the strings from RoutePaths, so if they were to change this type would show an error: https://www.typescriptlang.org/docs/handbook/utility-types.html#extracttype-union

Something like this could work:
```suggestion
type ValidPassworUrls = Extract<RoutePaths, '/reset-password', '/set-password', '/welcome'>;

type ValidResendEmailUrls = Extract<RoutePaths, '/reset', '/welcome', '/set-password'>;
```

src/server/lib/middleware/errorHandler.ts
```diff
@@ -52,7 +52,7 @@ export const routeErrorHandler = (
 
   logger.error('unexpected error', err);
 
-  const html = renderer(`${Routes.UNEXPECTED_ERROR}`, {
+  const html = renderer(`${'/error'}`, {
```
coldlink, line 55:
```suggestion
  const html = renderer('/error', {
```

coldlink (commented):
Just a few comments!

docs/development.md
```diff
@@ -28,34 +28,38 @@ When a browser makes a request to gateway, routes must render an output, this go
...
+
 Here's an example of a `GET` route, which renders a page and returns it to the client.
 
 ```ts
 ...
-const router = Router();
+import { typedRouter as router } from '@/server/lib/typedRoutes';
 ...
 // tell express the method (GET)
 router.get(
```
coldlink, line 60:
```suggestion
typedRouter.get(
```

coldlink, line 60:
```suggestion
typedRouter.get(
```

src/server/controllers/changePassword.ts
dominickendrick: Nice !

dominickendrick: The syntax was 

```ts 
type ValidResendEmailUrls = Extract< '/reset' | '/welcome' | '/set-password',  RoutePaths>;
``` 

but it works a charm

docs/development.md
dominickendrick: I added the import alias `{ typedRouter as router }` so I wouldn't need to rename all the instances of `router.get` to `typedRouter.get` but I could be persuaded to do that if it makes it clearer ?

coldlink: Ah don't mind me! I got confused. They way you've done it fine with me.

src/server/lib/typedRoutes.ts
```diff
@@ -0,0 +1,16 @@
...
+export const typedRouter = {
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  get: (path: RoutePaths, ...handlers: Array<any>) => {
+    return router.get(path, ...handlers);
+  },
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  post: (path: RoutePaths, ...handlers: Array<any>) => {
+    return router.post(path, ...handlers);
+  },
+  router: router,
```
coldlink, line 15:
```suggestion
  router,
```

src/server/lib/typedRoutes.ts
```diff
@@ -0,0 +1,16 @@
+import { RoutePaths } from '@/shared/model/Routes';
+import { Router } from 'express';
+
+const router = Router();
+
+export const typedRouter = {
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  get: (path: RoutePaths, ...handlers: Array<any>) => {
```
coldlink, line 8: trying to figure out a way we could avoid the `any` here, could the `Paramters` type  possibly work? https://www.typescriptlang.org/docs/handbook/utility-types.html#parameterstype

coldlink: actually i tried this and it doesn't quite seem to work 😅
might just have to rely on the `any`

dominickendrick: Cool, I did play around with it too, but couldn't get anything else to work

PR #1227, opened 2021-12-03
3rd December 2021 | Yet more dependency updates

coldlink:
## What does this change?

## Notes
- `react-router-dom` is a major update, so will move into a separate PR
- Storybook/Chromatic also updated react router, so can only update them once we update react router/at the same time
- Fixed `cypress-mailosaur` version to `2.5.0` for now as types were removed in `2.5.1` for some reason

## Tested?
- [x] Local
- [x] CI
- [x] CODE


vlbee (approved):

ob6160:
just checked the Mailosaur repo: this could've been the commit that introduced the problem https://github.com/mailosaur/cypress-mailosaur/commit/1bab3bf8770a17dc89d9c36eaa3dd26b8af29614 perhaps we could open a fix PR / raise an issue 🙂

PR #1228, opened 2021-12-03
Integrate sign in endpoint and page with reCAPTCHA 

ob6160:
## What's changed?
Completes reCAPTCHA integration for the Sign In flow, using the `useRecaptcha` hook from #961. 

Largely uses the same code introduced in #968. I haven't done much refactoring to share code between Registration and SignIn though, that's because we'd like to move both of these pages to use the more generic reCAPTCHA implementation in `MainForm` introduced in #1176. I've left a comment to that effect.

The manual test plan for MVP1 will also be updated along with this PR so that we cover these changes during our testing too.

## Testing

Adds an e2e and mocked cypress test that checks that we handle an error state returned from the reCAPTCHA service and how we handle a scenario in which the clients' browser goes offline.

Uses end to end testing keys provided publicly from Google which bypass the reCAPTCHA challenge so that our tests can complete the whole process.


ob6160:
[Add recaptcha to the Sign In page](https://trello.com/c/FAx8xduG/3579-add-recaptcha-to-the-sign-in-page)

coldlink (approved):

ob6160:
thanks for the review @coldlink  will merge on Monday

PR #1233, opened 2021-12-03
Update react router to 6.0.2 and related deps

coldlink:
## What does this change?

- Updates `react-router` to major version 6 outside of other dependency updates, and also updates storybook/chromatic at the same time which use `react-router` as a dependency
  - Followed this [upgrade guide](https://reactrouter.com/docs/en/v6/upgrading/v5)
  - We initially updated it in https://github.com/guardian/gateway/pull/1112 however this caused issues which we had to revert
  - Testing this new branch on CODE doesn't cause any 5xx errors like we were previously getting
- We also had to update how routes are defined in [`src/client/routes.tsx`](src/client/routes.tsx) to use a typed array to define the `path` and `element` to render
  - We were using a custom component called `TypedRoute` to enforce type safety on the route, however custom route components are [disallowed](https://reactrouter.com/docs/en/v6/upgrading/v5#refactor-custom-routes) in v6
  - By using an array, we can enforce type safety on the array, and then map over the array creating a normal react-router `Route` component, avoiding the need to use the custom component

## Tested
- [x] Local
- [x] CODE

### DIFF
--- a/src/shared/lib/routeUtils.ts
+++ b/src/shared/lib/routeUtils.ts
@@ -20,9 +20,9 @@ import {
 export type ExtractRouteParams<T> = string extends T
   ? Record<string, string>
   : T extends `${infer _Start}:${infer Param}/${infer Rest}`
-  ? { [k in Param | keyof ExtractRouteParams<Rest>]: string }
+  ? { [k in Param | keyof ExtractRouteParams<Rest>]?: string }
   : T extends `${infer _Start}:${infer Param}`
-  ? { [k in Param]: string }
+  ? { [k in Param]?: string }
   : {};
 
 /**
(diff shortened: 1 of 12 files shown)


ob6160 (approved):
Looks good, second time lucky!

src/shared/lib/routeUtils.ts
```diff
@@ -20,9 +20,9 @@ import {
 export type ExtractRouteParams<T> = string extends T
   ? Record<string, string>
   : T extends `${infer _Start}:${infer Param}/${infer Rest}`
-  ? { [k in Param | keyof ExtractRouteParams<Rest>]: string }
+  ? { [k in Param | keyof ExtractRouteParams<Rest>]?: string }
   : T extends `${infer _Start}:${infer Param}`
-  ? { [k in Param]: string }
+  ? { [k in Param]?: string }
```
coldlink, line 25: Updating react router also meant that the route parameter (e.g. how `/:token` becomes `{ token }`) is defined is also changed. React router now enforces these as an optional string, so we have to update the types here so they are an optional string.

ob6160 (approved):
changes to routing look good, shame about not being able to use the old solution anymore. this seems solid though 👍

coldlink:
> changes to routing look good, shame about not being able to use the old solution anymore. this seems solid though 👍

imo this solution is better in some ways, as we're not modifying any existing components or types to get the type safety

ob6160:
> 

agreed, it's clearer that we're changing the default functionality 👍

PR #1250, opened 2021-12-06
Fixes for LastPass on email and password input fields

ob6160:
## What does this change?

Fixes email input validation when autofilling with LastPass. When LastPass submits the form it fires a `keyDown` event with an empty `keyCode`. The fix was to listen on that event and check the field validity on that condition.

The LastPass autofill icon also clashed with our absolutely positioned password view/hide button. To fix that I added the button back into the flow and wrapped in a flex container. This means that the button is no longer overlayed so the two co-exist happily. This should work for other password managers/plugins which use the absolute positioning trick.

This PR also removes the blue autofill background from the two input fields for consistency across browsers.

## Icons Before
Screenshot 2021-12-07 at 13 21 53

## Icons After
Screenshot 2021-12-07 at 13 21 38


ob6160:
[Error when autofilling email with LastPass and problem where you have to click twice to dismiss error message](https://trello.com/c/46a4OLDW/3621-error-when-autofilling-email-with-lastpass-and-problem-where-you-have-to-click-twice-to-dismiss-error-message)

src/client/components/EmailInput.tsx
```diff
@@ -67,10 +68,17 @@ export const EmailInput: React.FC<EmailInputProps> = ({
...
       onBlur={(e) => {
         // Transition error state when the input box loses focus.
         transitionState(e.target.validity);
       }}
+      onKeyDown={(e) => {
+        if (e.code === '') {
+          e.currentTarget.checkValidity();
+          transitionState(e.currentTarget.validity);
+        }
+      }}
```
liamness, line 81: Is this unrelated to the lastpass fixes?

src/client/components/PasswordInput.stories.tsx
```diff
@@ -1,23 +1,38 @@
...
 WithSupporting.storyName = 'with supporting text';
+
+export const WithoutEye = (props: PasswordInputProps) => (
+  <PasswordInput
+    {...props}
+    label="New password"
+    supporting="Must be between 8 and 72 characters"
+    displayEye={false}
+  />
+);
```
liamness, line 37: The spread props might need to come after the other props, if we want people viewing the story to be able to override all values.

src/client/styles/Shared.ts
```diff
@@ -18,3 +18,12 @@ export const topMargin = css`
...
   }
 `;
+
+export const disableAutofillBackground = css`
+  :-webkit-autofill,
+  :-webkit-autofill,
+  :-webkit-autofill {
+    background-color: transparent !important;
+    box-shadow: 0 0 0 50px white inset;
+  }
```
liamness, line 28: Am I correct in assuming this draws the "shadow" over the brower default background?

I am not sure why `:-webkit-autofill` is targeted three times.

liamness (commented):
Looks good, I have some questions though.

src/client/components/EmailInput.tsx
ob6160: this one is so that when LastPass autofills the email field we run a validation check

src/client/components/PasswordInput.stories.tsx
ob6160: good spot, will change

src/client/styles/Shared.ts
ob6160: yep - it adds an inner shadow that occludes the default autofill background. seemed like the best suggested solution.
good spot re the three usages, will remove

ob6160: I've changed so the selector targets:
```
:autofill,
:-webkit-autofill {
...
}
```

as recommended here https://developer.mozilla.org/en-US/docs/Web/CSS/:autofill#examples

liamness (approved):

PR #1258, opened 2021-12-08
Track server side reCAPTCHA middleware error and success as metric

ob6160:
## What does this change?
Adds a metric to track the frequency of successful vs failed reCAPTCHA verification requests: `RecaptchaMiddleware::Success`, `RecaptchaMiddleware::Failure`.

This could be useful if we are experiencing an extended bot attack for example and would like to track how many requests are failing the reCAPTCHA validation middleware.


ob6160:
[Add failed registration/signin metric for reCAPTCHA server side fail](https://trello.com/c/ScbtIR6c/3620-add-failed-registration-signin-metric-for-recaptcha-server-side-fail)

coldlink (approved):

PR #1273, opened 2021-12-13
Show Jobs specific terms and conditions

ob6160:
## What does this change?
This change adds support jobs user registration. We now persist the `clientId` parameter across the Sign In, Register and Reset Password pages if it is set.

When `clientId=jobs` we show the Jobs specific terms and conditions on the Sign In and Registration pages.

Currently on `identity-frontend` we also have a checkbox which allows the user to agree to receive marketing communication related to Jobs. After some discussion the decision was made to not provide the checkbox and opt-out users automatically. This meant that we didn't have to recreate this behaviour in Gateway.

The changes to the persisted `clientId` and new terms & conditions text are tested through e2e, mocked and modified unit tests.

## Images
Screenshot 2021-12-21 at 16 29 31


QuarpT (approved):

src/shared/model/QueryParams.ts
```diff
@@ -20,7 +21,7 @@ export interface PersistableQueryParams
   extends TrackingQueryParams,
     StringifiableRecord {
   returnUrl: string;
-  clientId?: string;
+  clientId?: typeof validClientId[number];
```
dominickendrick, line 24: I know this is already merged, but could this type not just be `ValidClientId` (that is exported here https://github.com/guardian/gateway/pull/1273/files#diff-ae8ba4ba4c4c6df1c98f5fc76009bd75c0c658fc4ebff6b591d68a672dfa9c1fR9)

ob6160: ah yeah, I added this and forgot to change to use that. good spot! Another jobs PR is likely coming, so I can change it in that

PR #1275, opened 2021-12-14
Refactor password controller

rebecca-thompson:
## What does this change?
This is a refactoring PR which breaks up the `changePassword.ts` controller into two controllers - one to check the password reset token and one to change the password. It also makes the following changes:

- migrate the `PageTitle.ts` from an enum to a type. This is consistent with the changes made in this pr https://github.com/guardian/gateway/pull/1224
- differentiate between 'PageTitle' which is used to populate the <title> tag by the renderer and 'PageHeader' which is used in some templates to pass in the h1 header.
- updates some of the reset password routes to be consistent with the `set-password` route patterns. Specifically:
  - /password/reset-confirmation --> /reset-password/complete
  - /reset/resend --> /reset-password/resend
  - /reset/expired --> /reset-password/expired
  - /reset --> /reset-password
  - /reset/email-sent --> /reset-password/email-sent
- Break the `changePassword.ts` controller up into three files, extract the password validation into libs and add tests

## How to test
The following scenarios have been tested in `CODE`:

Registration path (new user)
- [x] set password
- [x] resend email (link expired)
- [x] resend email (email sent page)

Registration path (guest user)
- [x] set password
- [x] resend email (link expired)
- [x] resend email (email sent page)

Reset password path
- [x] send email
- [x] resend email
- [x] set password
- [x] resend email (link expired)


rebecca-thompson:
[Refactor password controller](https://trello.com/c/dqrBNCXh/3673-refactor-password-controller)

src/shared/model/Routes.ts
```diff
@@ -10,13 +10,12 @@ export type RoutePaths =
...
+  // end routes to be refactored
   | '/reset-password'
+  | '/reset-password/expired'
   | '/reset-password/:token'
-  | '/password/reset-confirmation'
-  | '/reset/resend'
-  | '/reset/expired'
-  // end routes to be refactored
+  | '/reset-password/complete'
+  | '/reset-password/resend'
```
coldlink, line 18: I wonder if it's worth refactoring `/reset` into `/reset-password` like you did with some of these so that they're all the same and clear what they're doing.

And then add a rewrite rule in fastly for backwards compatibility?

```vcl
if (req.url ~ "^/reset/") {
    set req.url = regsub(req.url, "^/reset/","/reset-password/");
}
```

src/server/routes/changePassword.ts
```diff
@@ -3,74 +3,63 @@ import deepmerge from 'deepmerge';
...
-  ),
-);
-
-router.get(
-  '/password/reset-confirmation',
+  '/reset-password/complete',
   (req: Request, res: ResponseWithRequestState) => {
     const email = readEmailCookie(req);
 
-    const html = renderer('/password/reset-confirmation', {
```
coldlink, line 48: and maybe a rewrite rule for this one too in vcl

coldlink (commented):
Looking good! I like the changes, just a thought about refactoring the rest of the routes before I approve this?

src/server/routes/changePassword.ts
coldlink:
```vcl
if (req.url ~ "^/password/reset-confirmation") {
    set req.url = regsub(req.url, "^/password/reset-confirmation","/reset-password/complete");
}
```

coldlink (approved):
💯

PR #1279, opened 2021-12-15
Refactor logging: Remove logstash, use Kinesis

coldlink:
## What does this change?
Use AWS Kinesis directly in gateway for logging, replacing logstash to resolve the log4j vulnerability. Example of logs coming through in Kibana: https://logs.gutools.co.uk/s/identity/goto/5718bb0ff3699215b6bd453bec4b5662

- Remove logstash configuration from the cloudformation.
- Add environment variables (only for `CODE`/`PROD`) for the `LOGGING_KINESIS_STREAM` and `EC2_INSTANCE_ID`
- Set up new ["winston transport"](https://github.com/winstonjs/winston-transport#usage) to pipe logs directly to aws kinesis through our logging solution "winston"
  - We have to use `any` in typescript as the shape of the `info` object can change.
- Type the `stage` variable so it has to be one of `DEV | CODE | PROD`
- Update `getOrThrow` to add another parameter to do strict checking
  - strict checking checks for both `undefined` and empty string `""`
  - disabling strict checking only checks for `undefined`
  - strict checking enabled by default
  - this option is needed as in CI the kinesis stream variable should be an empty string, but was previously failing as `getOrThrow` was doing a simple falsey check
- Update the aws configuration options so that the following happens to help prevent promises from leaking:
  - `maxRetries = 0` - Set retries to 0 so that aws doesn't attempt to retry the request
  - `httpOptions.connectTimeout = 500` - The maximum time in milliseconds that the connection phase of the request should be allowed to take.
  - `httpOptions.timeout = 250` - Sets the socket to timeout after timeout milliseconds of inactivity on the socket. Defaulted previously two minutes (120000).
  - This also affects `trackMetric` too
  - Use `includes` for certain error codes as some errors don't exactly match
 
## Post-merge
- [ ] Update the AMI to remove logstash

### DIFF
--- a/src/server/lib/logger.ts
+++ b/src/server/lib/logger.ts
@@ -47,11 +47,9 @@ class KinesisTransport extends Transport {
           .promise()
           .catch((error: AWSError) => {
             if (error.code === 'ExpiredToken' && stage === 'DEV') {
-              logger.warn(
+              console.warn(
                 'AWS Credentials Expired. Have you added `Identity` Janus credentials?',
               );
-            } else {
-              logger.error('Kinesis Logging Error', error);
             }
           });
       }
--- a/src/server/lib/logger.ts
+++ b/src/server/lib/logger.ts
@@ -1,9 +1,68 @@
+import { AWSError, Kinesis } from 'aws-sdk';
 import { Logger, LogLevel } from '@/server/models/Logger';
 import { createLogger, transports } from 'winston';
+import Transport, { TransportStreamOptions } from 'winston-transport';
 import { formatWithOptions, InspectOptions } from 'util';
+import { awsConfig } from './awsConfig';
+import { getConfiguration } from './getConfiguration';
+
+const {
+  stage,
+  aws: { instanceId, kinesisStreamName },
+} = getConfiguration();
+
+// custom "Winston Transport" to send logs to the AWS kinesis stream
+// see https://github.com/winstonjs/winston-transport#usage
+class KinesisTransport extends Transport {
+  private kinesis: Kinesis;
+
+  constructor(opts?: TransportStreamOptions) {
+    super(opts);
+
+    this.kinesis = new Kinesis(awsConfig);
+  }
+
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  public log(info: any, callback: () => void) {
+    const { level, message } = info;
+
+    setImmediate(() => {
+      this.emit('logger', level);
+      if (kinesisStreamName) {
+        this.kinesis
+          .putRecord({
+            StreamName: kinesisStreamName,
+            PartitionKey: stage,
+            Data: JSON.stringify({
+              type: 'app',
+              app: 'identity-gateway',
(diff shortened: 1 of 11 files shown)


dominickendrick (approved):

dominickendrick (approved):

dominickendrick (dismissed):

src/server/lib/logger.ts
```diff
@@ -1,9 +1,68 @@
...
+              instance_id: instanceId,
+              stage,
+              level,
+              message,
+            }),
+          })
+          .promise()
+          .catch((error: AWSError) => {
+            if (error.code === 'ExpiredToken' && stage === 'DEV') {
+              logger.warn(
```
QuarpT, line 50: does calling log on logger failure cause an infinite loop?

src/server/lib/logger.ts
```diff
@@ -1,9 +1,68 @@
...
+  }
+
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  public log(info: any, callback: () => void) {
+    const { level, message } = info;
+
+    setImmediate(() => {
+      this.emit('logger', level);
+      if (kinesisStreamName) {
+        this.kinesis
```
QuarpT, line 32: what happens if the app logs too much, and the kinesis stream starts blocking or is slow to respond?

In 2018 there was a major incident caused by bad logging behaviour on dotcom frontend, so in the Scala apps we use either AsyncAppender with never block = true
https://github.com/guardian/ophan/blob/main/shared-lib/src/main/scala/ophan/util/ElkLogging.scala#L66

or a logging circuit breaker when this can't be used
https://github.com/guardian/identity/blob/main/identity-api/src/main/scala/com/gu/identity/api/logging/SafeBlockingKinesisAppender.scala#L39

QuarpT (changes requested):
https://github.com/guardian/gateway/pull/1279#discussion_r769742984

src/server/lib/logger.ts
coldlink: good spot! i'll change it to a `console.warn`, and i'll remove the logging failure error

src/server/lib/logger.ts
coldlink: After discussion we decided to update the aws configuration options so that the following happens:
- `maxRetries = 0` - Set retries to 0 so that aws doesn't attempt to retry the request
- `httpOptions.connectTimeout = 500` - The maximum time in milliseconds that the connection phase of the request should be allowed to take.
- - `httpOptions.timeout = 250` - Sets the socket to timeout after timeout milliseconds of inactivity on the socket. Defaultsed previously two minutes (120000).

coldlink: We also updated the `trackMetric` method to this too.

This should hopefully prevent any promises from leaking, although we may lose a few logs/metrics if they're long running, but this is unlikely and would only happen under huge loads.

QuarpT (approved):
looks great

PR #1280, opened 2021-12-16
Add `ref` and `refViewId` tracking parameters to gateway flows

coldlink:
## What does this change?

- Adds the `ref` and `refViewId` tracking parameters to endpoints in gateway where we want to track the users journey through email.
- The first commit adds it to the client side
  - We use a custom hook (`useRefTracking`) to get the `ref` and `refViewId` parameters if they are defined.
  - This is consumed by a custom component (`RefTrackingFormFields`) which are hidden input fields which can be added to forms. This means that on a form submit these parameters will also be included in the request body
  - We then append it to any form which could rely on this: `MainForm`, `Registration`, and `SignIn`
- The second commit modifies the server to be able to consume these parameters from the body
  - The `parseExpressQueryParams` method is modified to be able to read parameters from `req.query` and `req.body`
    - We also improve the way we type this method by using utility types
  - The `requestStateMiddleware` is then modified to consume this new way of parsing query params
    - As we're using utility types to type the query parameters now, we introduce a `RequestWithTypedQuery` type to expand the express `Request` type with exactly which query parameters we're expecting at compile time.
  - We can then read these query parameters on the `ResponseWithRequestState`, so the `res.locals.requestState` object, and then use them in the relevant IDAPI controllers.
  - This commit also refactors usage of `getConfiguration` and makes sure that it is destructed where its used rather than a config object being passed around.
    - This is to avoid any consuming methods that were using the `config` object to read any variables that it shouldn't have access to
- Also use `jest-mock` instead of `ts-jest/utils` as the functionality of mocks has moved from the latter (where it's been deprecated) to the former. 

## Related
[Identity API PR](https://github.com/guardian/identity/pull/2037) which should be reviewed and go in before this PR.

### DIFF
--- a/src/client/lib/hooks/useRefTracking.ts
+++ b/src/client/lib/hooks/useRefTracking.ts
@@ -1,0 +1,12 @@
+// react hook for getting the `ref` and `refViewId` for tracking
+// `ref` is the current url of the page we want to track
+// `refViewId` is the current ophan page view id we want to track
+export const useRefTracking = (): { ref?: string; refViewId?: string } => {
+  if (typeof window !== 'undefined') {
+    return {
+      ref: `${window.location.origin}${window.location.pathname}`,
+      refViewId: window.guardian.ophan?.viewId,
+    };
+  }
+  return {};
+};
--- a/src/client/lib/hooks/useRefTracking.ts
+++ b/src/client/lib/hooks/useRefTracking.ts
@@ -1,7 +1,9 @@
+import { TrackingQueryParams } from '@/shared/model/QueryParams';
+
 // react hook for getting the `ref` and `refViewId` for tracking
 // `ref` is the current url of the page we want to track
 // `refViewId` is the current ophan page view id we want to track
-export const useRefTracking = (): { ref?: string; refViewId?: string } => {
+export const useRefTracking = (): TrackingQueryParams => {
   if (typeof window !== 'undefined') {
     return {
       ref: `${window.location.origin}${window.location.pathname}`,
--- a/src/client/pages/Registration.tsx
+++ b/src/client/pages/Registration.tsx
@@ -22,6 +22,7 @@ import { EmailInput } from '@/client/components/EmailInput';
 import { buildUrlWithQueryParams } from '@/shared/lib/routeUtils';
 import { QueryParams } from '@/shared/model/QueryParams';
 import { GeoLocation } from '@/shared/model/Geolocation';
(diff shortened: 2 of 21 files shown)


src/client/lib/hooks/useRefTracking.ts
```diff
@@ -0,0 +1,12 @@
+// react hook for getting the `ref` and `refViewId` for tracking
+// `ref` is the current url of the page we want to track
+// `refViewId` is the current ophan page view id we want to track
+export const useRefTracking = (): [string | undefined, string | undefined] => {
```
ob6160, line 4: if this returned an object instead of an array it might be easier for the user to know that the hook returns `ref` and `refViewId` at the point of destructuring the return val

src/server/lib/queryParams.ts
```diff
@@ -45,18 +45,8 @@ export const parseExpressQueryParams = (
...
-    emailSentSuccess?: string;
-    csrfError?: string;
-    recaptchaError?: string;
-    refViewId?: string;
-    ref?: string;
-    encryptedEmail?: string;
-    error?: string;
-  },
+  }: Record<keyof QueryParams, string | undefined>, // parameters from req.query
+  bodyParams: Pick<QueryParams, 'ref' | 'refViewId'> = {}, // some parameters may be manually passed in req.body too
```
ob6160, line 49: wondering if it might be nice to extract `'ref' | 'refViewId'` into a type called `AllowedBodyParams` or something along those lines

ob6160 (commented):
nice work! `jest-mock` is cool

src/client/pages/Registration.tsx
```diff
@@ -130,6 +131,7 @@ export const Registration = ({
         >
           <RecaptchaElement id="register-recaptcha" />
           <CsrfFormField />
+          <RefTrackingFormFields />
```
ob6160, line 134: I like this pattern!

src/server/lib/queryParams.ts
coldlink: The main reason I didn't extract it into a separate type is that this is the only place that it needs to be used. If it was being used in multiple places then that would be an option.

coldlink: Actually I did end up refactoring it into a shareable `TrackingQueryParams` interface, as we could use the same interface as a type in your previous comment..

ob6160: nice! :)

ob6160 (approved):
🎄

PR #1281, opened 2021-12-16
Mh/update onboarding flow

micr-urus:
## Accessibility


package.json
```diff
@@ -130,6 +130,7 @@
     "deepmerge": "^4.2.2",
     "express": "^4.17.1",
     "express-recaptcha": "^5.0.2",
+    "gifsicle": "^6.1.0",
```
coldlink, line 133: Is this new dependency used anywhere?

src/client/components/ConsentsSubHeader.tsx
```diff
@@ -11,8 +11,6 @@ import {
 } from '@guardian/source-foundations';
 import { AutoRow, gridItemColumnConsents, gridRow } from '@/client/styles/Grid';
 import { CONSENTS_PAGES_ARR } from '@/client/models/ConsentsPages';
-import { CONSENTS_MAIN_COLOR } from '@/client/layouts/shared/Consents';
-import { SvgCheckmark } from '@guardian/source-react-components';
```
liamness, line 15: This seems to have been removed in error? Results in `make dev` failing locally if I rebase off this branch, due to `<SvgCheckmark />` being used on line 259.

.gitignore
```diff
@@ -50,3 +50,5 @@ cypress/videos
 
 # Cypress screenshots
 cypress/screenshots
+
+.DS_Store
```
coldlink, line 54: I'm still seeing a few `.DS_Store` files have been committed below:

chrome_40gJZ3IfeX

Worth removing them manually, and possibly using the following to also match subdirectories
```suggestion
**/.DS_Store
```

package.json
```diff
@@ -132,6 +132,7 @@
     "deepmerge": "^4.2.2",
     "express": "^4.17.2",
     "express-recaptcha": "^5.0.2",
+    "gifsicle": "^6.1.0",
```
coldlink, line 135: Can we remove this dependency if it's not being used?

dominickendrick:
Hey, Could we provide some before and after screenshots of all the design changes in the PR description ? It would also be good to get a summary of the changes made.

src/client/components/NewsletterCard.tsx
```diff
@@ -108,51 +106,134 @@ const subtitleDiv = css`
...
+    border-bottom: 1px solid ${neutral[86]};
+  }
+`;
+
+const checkboxGroup = css`
+  ${from.tablet} {
+    // position: absolute;
+    // bottom: 0;
+  }
+`;
```
coldlink, line 151: this css doesn't seem to be doing anything with the `//` comment blocks, can this be removed?

src/client/components/NewsletterCard.tsx
```diff
@@ -108,51 +106,134 @@ const subtitleDiv = css`
...
+  const { description, frequency, name, nameId } = props.newsletter;
   const subtitle = frequency ? (
-    <div css={subtitleDiv}>
-      {clockSVG}
-      <h2 css={h2}>{frequency}</h2>
+    <div css={[subtitleDiv, `svg {fill: ${idColor(nameId)}}`]}>
+      {/* {clockSVG} */}
+      {/* <h2 css={[h2, `color: ${idColor(nameId)}`]}>{frequency}</h2> */}
     </div>
   ) : null;
```
coldlink, line 196: The commented out code here suggests to be that nothing is displayed?
Just an empty `div` with some styling applied. Is this intentional? Can the commented code be removed?

src/client/pages/ConsentsCommunication.tsx
```diff
@@ -1,53 +1,100 @@
 import React from 'react';
 import { css } from '@emotion/react';
-import { from, space } from '@guardian/source-foundations';
+import {
+  from,
+  space,
+  brand,
+  neutral,
+  // textSans,
```
coldlink, line 8:
```suggestion
```

src/client/pages/ConsentsData.tsx
```diff
@@ -75,15 +83,31 @@ export const ConsentsData = ({ consented, description }: ConsentsDataProps) => {
...
+                    label="Yes"
+                    defaultChecked={consented}
+                    css={radio}
+                  />
+                </div>
+                <div css={radio}>
+                  <Radio
+                    value="consent-option"
+                    label="No"
+                    // defaultChecked={consented}
```
coldlink, line 105:
```suggestion
```

coldlink: or possibly
```suggestion
                    defaultChecked={!consented}
```

src/client/pages/ConsentsNewsletters.tsx
```diff
@@ -57,26 +57,31 @@ export const ConsentsNewsletters = ({
   newsletters,
 }: ConsentsNewslettersProps) => {
   const autoRow = getAutoRow(1, gridItemColumnConsents);
-
+  console.log('NEWSLETTERS', newsletters);
```
coldlink, line 60:
```suggestion
```

src/client/pages/Welcome.tsx
```diff
@@ -38,6 +46,8 @@ export const Welcome = ({
 }: Props) => {
   const autoRow = getAutoRow(1, gridItemColumnConsents);
 
+  console.log('FIELDERRORS:', fieldErrors);
```
coldlink, line 49:
```suggestion
```

coldlink (changes requested):

coldlink:
There's a bunch of cypress tests failing.

I suggest running
`make cypress-mocked` for the cypress mocked tests (running against a mocked version of identity api)
and
`make cypress-ete` for the end to end tests running again identity api as defined in your `.env` file

liamness (commented):
This is representative of half a review really, haven't looked at all the files yet, paused for now as Michael tells me there is some wrangling with git he needs to do.

I noticed that the cypress tests were failing before FYI. Might want to check that once the branch is in a good state.

src/client/pages/ConsentsCommunicationPage.tsx
```diff
@@ -1,13 +1,6 @@
...
 import { ConsentsCommunication } from '@/client/pages/ConsentsCommunication';
 
 export const ConsentsCommunicationPage = () => {
-  const clientState = useContext<ClientState>(ClientStateContext);
-
-  const { pageData = {} } = clientState;
-  const { consents = [] } = pageData;
-
-  return <ConsentsCommunication consents={consents} />;
+  return <ConsentsCommunication />;
```
liamness, line 5: If `pageData` is no longer used, likely there is code that can be cleaned up in `src/server/routes/consents.ts` which prepares this data.

src/client/components/NewsletterCard.tsx
```diff
@@ -108,51 +106,134 @@ const subtitleDiv = css`
...
+  if (id === 'the-long-read') {
+    return brand[400];
+  }
+  if (id === 'green-light') {
+    return news[400];
+  }
+  if (id === 'bookmarks') {
+    return culture[400];
+  }
+};
```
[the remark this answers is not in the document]
liamness: Had another thought. We already have the enum `Newsletters` which can be imported from `'@/shared/model/Newsletter'`, would be preferable to use those ids instead of the `nameId` just so we don't end up with two different approaches.

Also noticed you are only handling `today-us`. For other locales (AU, GB, ROW) a colour won't be defined.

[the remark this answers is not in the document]
liamness: Pushed an approach on my rebased branch that you might find useful? https://github.com/guardian/gateway/commit/de91169373fcbe28e6dbef8b809bd2c6ad89b8a1

.gitignore
liamness: Actually I think just `.DS_Store` should work, but it is tracking them anyway because they have already been committed.

src/client/components/NewsletterCard.tsx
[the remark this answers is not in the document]
micr-urus: Very useful @liamness. Thank you!

.gitignore
micr-urus: This file should be gone now, as of the last commit

src/client/components/NewsletterCard.tsx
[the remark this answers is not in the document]
liamness: Hey as you mentioned you were waiting for reviews at standup, I thought it would be worth asking, have you looked into this yet? Looks like there will still be issues in non-US locales.

src/client/pages/ConsentsCommunicationPage.tsx
liamness: Similarly to my other comment, wondering if you've looked at this? Maybe not a blocker if you've got a follow-up PR coming though, as I don't think this will cause any noticable issues for users, it will just be firing an unneccessary request to fetch the consents.

.gitignore
vlbee: I can see the `.DS_Store` was committed on the third commit (SHA 582d34f) so you might need to rebase or do an interactive rebase to remove it from there.

src/client/styles/Consents.ts
```diff
@@ -32,3 +32,29 @@ export const text = css`
...
+  padding-bottom: ${space[1]}px;
+`;
+
+export const passwordButton = css`
+  width: 100%;
+  justify-content: center;
+
+  // ${from.mobileMedium} {
+  //   width: max-content;
+  // }
```
vlbee, line 59: Some commented out code here

src/client/pages/ConsentsCommunicationPage.tsx
liamness: Thinking about it now, I actually think this was removed in error. Seems the consent information has been hardcoded into `src/client/pages/ConsentsCommunication.tsx, which means the copy won't change even if the response from IDAPI does.

src/client/pages/ConsentsCommunication.tsx
```diff
@@ -1,53 +1,99 @@
...
+  brand,
+  neutral,
+  headline,
+} from '@guardian/source-foundations';
 import {
   getAutoRow,
   gridItemColumnConsents,
   consentsParagraphSpanDef,
 } from '@/client/styles/Grid';
-import { CommunicationCard } from '@/client/components/CommunicationCard';
```
liamness, line 9: I don't believe there are any other usages of `CommunicationCard`, could be cleaned up perhaps.

liamness:
Couple of issues I noticed running this branch locally (as I'm looking into what I need to use from the branch in another story):

Firstly, opting into the supporter consent permission does not seem to work anymore. The POST request does not have the correct id for the required consent. I think this is because the attributes of the checkbox are incorrect? It also no longer displays the checkbox as ticked if the user has already consented. This will prevent a user from viewing their previous choice should they go back to this page.

Secondly, the image used for the supporter consent is pretty low quality and will likely look poor on mobile screens. I have figured out how to export full quality images from figma if you want help with that. Or if we are okay with releasing this as is, I could fix in my branch, as the same image is also used on the prompt page.

Thirdly, the styles for the newsletter card and the supporter consent card seem extremely similar, but are defined separately. I would consider this more of a missed opportunity than a blocker however. Again, something I could look into on my branch.

.gitignore
vlbee: I'm still seeing 3 .DS_Store files on here.

liamness:
`8ae16df82ba4fb07c65269759b3e39c3-2.png` is committed twice, and I don't see any actual usages of it. Could do with a nicer filename too?

src/client/pages/ConsentsCommunication.tsx
```diff
@@ -57,23 +59,30 @@ export const ConsentsCommunication = ({
...
+              autoRow(communicationCardSpanDef),
+            ]}
+          >
+            {consents.map((consent) => (
+              <CommunicationCard
+                key={consent.id}
+                title={consent.name}
+                body={consent.description}
+                value={consent.id}
+                checked={!!consent.consented}
```
coldlink, line 82: i think the `image` prop is missing from here, which is why the image isn't showing up for the consent

coldlink (commented):
As mentioned looks like `8ae16df82ba4fb07c65269759b3e39c3-2.png` is commited twice, once in the `/` folder.

Think it should be renamed to `supporter.png` and then it's missing from the communication card where it should be being used.

Other than that I think it's nearly there! Give me a shout if you need a hand with anything/want to pair on finishing it off.

vlbee:
I think this is supposed to be an information "i" on the password form. It's been like this for ages but there are two different versions on the figma but it makes more sense as an "i" since you can't click it to go away :) 

image

vlbee:
@coldlink Does it make sense to have a "Go Back" button on the Welcome Page from a UX perspective? what do you go back to? resetting your password?

coldlink:
> @coldlink Does it make sense to have a "Go Back" button on the Welcome Page from a UX perspective? what do you go back to? resetting your password?

There isn't anywhere to go back to on the welcome page, you land on that page from your email, so doesn't make sense to have one there. It's also not possible to go back to the welcome page either as it's technically a reset password page where the token has been used.

coldlink:
> I think this is supposed to be an information "i" on the password form. It's been an X for ages but there are two different versions on the figma but it makes more sense as an "i" since you can't click it to go away :)
> 
> image

This might be outside the scope of this ticket, we use this input field all over gateway, and it's been implemented currently with the `x`: https://60929d168594f80039336501-njyyebfjfg.chromatic.com/?path=/story/components-passwordform--default

vlbee:
Apologies I meant the Stay in Touch page after you've set a password. It was on the design but not in the code which I think makes more sense (not to have a go back button) so just double checking as not super familiar with the existing UX

> > @coldlink Does it make sense to have a "Go Back" button on the Welcome Page from a UX perspective? what do you go back to? resetting your password?
> 
> There isn't anywhere to go back to on the welcome page, you land on that page from your email, so doesn't make sense to have one there. It's also not possible to go back to the welcome page either as it's technically a reset password page where the token has been used, and you're now authenticated.

vlbee:
> > I think this is supposed to be an information "i" on the password form. It's been an X for ages but there are two different versions on the figma but it makes more sense as an "i" since you can't click it to go away :)
> > image
> 
> This might be outside the scope of this ticket, we use this input field all over gateway, and it's been implemented currently with the `x`: https://60929d168594f80039336501-njyyebfjfg.chromatic.com/?path=/story/components-passwordform--default

Agree out of scope. Just noticing.

liamness:
New screenshots:

image

image

image

image

image

src/client/components/NewsletterCard.tsx
```diff
@@ -114,45 +110,105 @@ const checkBoxBackgroundColorBugFix = css`
...
+  width: 100%;
+`;
+
+const frequencyStyles = css`
+  display: flex;
+  align-items: center;
+  padding: ${space[2]}px 0;
+`;
+
+const idColor: (arg0: string) => string | undefined = (id) => {
```
coldlink, line 155: was slightly confused by the type definition for this method, think this suggestion is clearer?

```suggestion
const idColor = (id: string): string => {
```

src/client/components/ConsentsSubHeader.tsx
```diff
@@ -91,38 +76,80 @@ const li = (numPages: number, isLastPage: boolean) => css`
...
+  }
+  &:first-child.complete::before {
+    left: 0;
+  }
+
+  &:last-child::after {
+    border: ${BORDER_SIZE}px solid ${neutral[60]};
+    border-radius: 50%;
     ${circle}
+    z-index: 100;
```
dominickendrick, line 122: Why does the `z-index` need to be set so high here ?

src/client/components/NewsletterCard.tsx
coldlink: as i don't think there is a scenario where `undefined` is returned, and was getting confused by the whole `arg0`

vlbee (approved):
👍

PR #1282, opened 2021-12-16
Test adding different newsletters into the onboarding flow

liamness:
## What does this change?

This sets up an A/B test where 50% of users will see different newsletter options when going through the onboarding flow.

## How to test

Set a `GU_mvt_id` cookie to `1` to see the changes being tested. Set it to `2` to see the control group behaviour, which shows the same list as is currently used.
Still looking into how best to test region-specific behaviour.

## Accessibility


src/server/routes/consents.ts
```diff
@@ -279,9 +294,23 @@ router.get(
...
       pageTitle = _pageTitle;
 
+      // TODO: decide which approach to use
+      const newslettersTestActive =
+        state.abTesting.participations[ONBOARDING_NEWSLETTERS_TEST_ID]
+          ?.variant === ONBOARD_NEWSLETTERS_TEST_VARIANT;
+      // const newslettersTestActive = state.abTestAPI.isUserInVariant(
+      //   ONBOARDING_NEWSLETTERS_TEST_ID,
+      //   ONBOARD_NEWSLETTERS_TEST_VARIANT,
+      // );
```
coldlink, line 304: in general this is the approach that we've taken in the past

src/server/routes/consents.ts
```diff
@@ -229,7 +244,7 @@ export const consentPages: ConsentPage[] = [
         page: 'review',
         consents: await getUserConsentsForPage(ALL_CONSENT, ip, sc_gu_u),
         newsletters: await getUserNewsletterSubscriptions(
-          NewsletterMap.get(geo) as string[],
+          TestNewsletterMap.get(geo) as string[],
```
coldlink, line 247: This change seems to fall outside the test, which may cause issues with reading newsletter subscriptions

src/server/routes/consents.ts
liamness: Okay, will stick with that then.

liamness: (discussed away from this PR with Mahesh)

For some reason `abTestAPI` was not available during cypress runs, so have switched to the other approach to unblock this feature.

PR #1283, opened 2021-12-16
Update documentation with latest information

coldlink:
## What does this change?

- Update the documentation with up to date information 
- Remove explicit type notation in `src/server/lib/getABTesting.ts` since we've updated past Typescript 4.2

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,11 +4,11 @@
 
 The (new) profile (dot) theguardian (dot) com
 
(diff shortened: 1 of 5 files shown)


ob6160 (approved):
looks great 

Thanks for adding that bit about how we run our tests in GH (requesting review/re-requesting) as that's a bit of a quirk in our setup

PR #1284, opened 2021-12-17
Tracking | Add form submit/input focus-blur/page load tracking/social button click

coldlink:
## What does this change?
Adds additional tracking through the use of ophan interaction events to help us measure how specific parts of the MVP1 flow are being used.

There are 4 main parts we're tracking
- Form submit
  - Does a user submit and go to a new page?
  - Does the same page reload? Hinting at error. Does user drop off?
  - `trackFormSubmit` helper method which wraps the `sendOphanInteractionEvent` and fires on a form submission
- Input field focus and blur
  - Does a user interact with specific fields?
    - e.g. are there some fields they have more trouble with, for hypothetical example, the password input box
  - `trackFormFocusBlur` helper method which wraps the `sendOphanInteractionEvent` and fires on a form focus/blur event, and uses the `event.target.tagName` to filter for input elements.
- Page load tracking
  - Specifically for the `ResetPassword` component, which is used as a generic resend email page, e.g. for forgot password, link expired, and session expired pages.
  - Since the urls are often in the same format e.g. `/reset-password/:token`, for both a valid/invalid/expired token, we want to differentiate between the different types of pages that are rendered
  - Create a `usePageLoadOphanInteraction` hook, which only runs on first render/hydration to fire the ophan event
- Social button click event tracking using [`data-link-name`](https://github.com/guardian/ophan/blob/ccaa57bcee3f5f3f83ec28973074c9b3f98f1438/tracker-js/assets/coffee/ophan/click-path-capture.coffee#L23)

## Other changes
- Adds `autocomplete` attributes to help browsers filling out certain field values.

### Form submit
chrome_TWjLYQhYnr
### Input field blur (loses focus)
chrome_rQ5v4f3g81
### Input field focus
chrome_RSSYJUHXYK
### Page load (render)
chrome_0wNe04NB10
### Social button click
chrome_d3g4iJ152f

### DIFF
--- a/src/client/components/MainForm.tsx
+++ b/src/client/components/MainForm.tsx
@@ -19,6 +19,7 @@ import {
 import { CaptchaErrors } from '@/shared/model/Errors';
 import { DetailedRecaptchaError } from '@/client/components/DetailedRecaptchaError';
 import { RefTrackingFormFields } from '@/client/components/RefTrackingFormFields';
+import { trackFormFocusBlur, trackFormSubmit } from '@/client/lib/ophan';
 
 export interface MainFormProps {
   formAction: string;
@@ -32,6 +33,7 @@ export interface MainFormProps {
   >;
   hasGuardianTerms?: boolean;
   onSubmitOverride?: React.FormEventHandler<HTMLFormElement>;
+  formTrackingName?: string;
 }
 
 const formStyles = css`
@@ -64,6 +66,7 @@ export const MainForm = ({
   setRecaptchaErrorContext,
   hasGuardianTerms = false,
   onSubmitOverride,
+  formTrackingName,
 }: PropsWithChildren<MainFormProps>) => {
   const recaptchaEnabled = !!recaptchaSiteKey;
   const hasTerms = recaptchaEnabled || hasGuardianTerms;
@@ -73,11 +76,15 @@ export const MainForm = ({
     useState<UseRecaptchaReturnValue>();
 
   /**
-   * Executes the reCAPTCHA check.
+   * Executes the reCAPTCHA check and form submit tracking.
    * Prevents the form from submitting until the reCAPTCHA check is complete.
    */
   const handleSubmit = useCallback(
     (event: React.FormEvent<HTMLFormElement>) => {
+      if (formTrackingName) {
+        trackFormSubmit(formTrackingName);
+      }
+
       if (onSubmitOverride) {
         onSubmitOverride(event);
       }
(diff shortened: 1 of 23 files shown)


src/client/components/MainForm.tsx
```diff
@@ -73,11 +76,15 @@ export const MainForm = ({
...
 
   /**
-   * Executes the reCAPTCHA check.
+   * Executes the reCAPTCHA check and form submit tracking.
    * Prevents the form from submitting until the reCAPTCHA check is complete.
    */
   const handleSubmit = useCallback(
     (event: React.FormEvent<HTMLFormElement>) => {
+      if (formTrackingName) {
+        trackFormSubmit(formTrackingName);
```
QuarpT, line 85: can we use data-link-name element attributes using click events, instead of calling ophan interaction events manually?
e.g.
image

coldlink: Is there any documentation on this? e.g. is it only for click events? does it work on form submit etc.

QuarpT: I can't find documentation.
This is the implementation, so looks like it's just a click event listener... https://github.com/guardian/ophan/blob/ccaa57bcee3f5f3f83ec28973074c9b3f98f1438/tracker-js/assets/coffee/ophan/click-path-capture.coffee#L23

looks like you can also add `data-component` attribute to the parent element

QuarpT: there are lots of examples of usage
https://github.com/search?q=org%3Aguardian+%22data-link-name%22&type=code

coldlink: Thanks! I can probably refactor it somewhat still, I'll switch to using the `data-link-name`. But keep the manual events for form submit/focus/blur.
For the social button tracking we can easily switch to the `data-link-names'

QuarpT: ah I see, for form submit a user can hit enter key rather than press the button, so `data-link-name` probably won't pick up these events 🤔

coldlink: Yeah, using the `onSubmit` option picks up both clicking the submit button, and using enter to submit the form. The `data-link-name` only works on click.

So I didn't refactor anything around submit/focus/blur, and left them as interaction events (which were already being used in the onboarding flow for tracking).

I've changed the social buttons to use `data-link-name` as thats a good use case for it!

QuarpT (approved):

PR #1316, opened 2021-12-23
Fix styling of password input and page progression

ob6160:
## What does this change?

This PR makes a few small styling tweaks to improve the user experience on iOS and other mobile viewports:

1. Makes the `PageProgression` component fill the container so that it looks better visually on mobile (before this change it was left aligned well but it didn't use all of the available space.
2. Turns on the "show password" eye for the rest password and welcome flows.
3. Fixes the border of the email and password input boxes on iOS so that they override the system default which applies a slight border radius.

Tested in IE11, Firefox, Chrome, Safari.

### Visually centring PageProgress and fixing iOS password input border radius

| Before  | After |
| ------------- | ------------- |
|   |  |


src/client/components/ConsentsSubHeader.tsx
```diff
@@ -78,8 +78,9 @@ const li = (numPages: number) => css`
     ${textSans.small()}
   }
   position: relative;
-  width: ${100 / numPages}%;
+  width: ${lastPage ? 'min-content' : 100 / (numPages - 1) + '%'};
```
ob6160: This makes it so that the last column width is as small as possible and we spread the remaining columns evenly.

We do this because we want to make component look visually centred. Without this modification, the furthest column would push the content further to the left than we want.

src/client/components/ConsentsSubHeader.tsx
```diff
@@ -78,8 +78,9 @@ const li = (numPages: number) => css`
...
   display: flex;
+  flex-grow: ${lastPage ? '0' : '1'};
```
ob6160: Similarly to the line above, this ensures that the last column doesn't grow and cause the steps to shift to the left

src/client/components/ConsentsSubHeader.tsx
```diff
@@ -158,6 +159,7 @@ const pageProgression = css`
 const ul = css`
   display: flex;
   list-style: none;
+  justify-content: space-between;
```
ob6160, line 162: We use the flex rule `space-between` here to spread each item out evenly across the span of the progress step container

src/client/styles/Grid.ts
```diff
@@ -229,7 +229,7 @@ export const gridItem = (spanDefinition?: SpanDefinition) => {
 export const gridItemColumnConsents: SpanDefinition = {
   TABLET: { start: 2, span: 10 },
   DESKTOP: { start: 2, span: 10 },
-  LEFT_COL: { start: 2, span: 10 },
+  LEFT_COL: { start: 2, span: 12 },
```
ob6160, line 232: with the changes to the column width, this need changing so that they spread across the correct number of columns

ob6160:
[Bugs on MVP1 - mobile](https://trello.com/c/l47NN7RQ/3686-bugs-on-mvp1-mobile)

rebecca-thompson (approved):
Looks good!

dominickendrick:
Hmmmmmmm, I wonder about the design on the progress bar - 
If you look at other places where this style is used - like the tube map
Screen Shot 2022-01-05 at 17 15 07
The left most item is left justified, the right most item is right justified, and the middle items are centred. If design are happy with the current setup, then I'm happy too, just thought I'd raise it.

ob6160:
> Hmmmmmmm, I wonder about the design on the progress bar - If you look at other places where this style is used - like the tube map Screen Shot 2022-01-05 at 17 15 07 The left most item is left justified, the right most item is right justified, and the middle items are centred. If design are happy with the current setup, then I'm happy too, just thought I'd raise it.

hm, interesting! It has sign off from design as-is ..  Although! In the Figma designs it does follow something closer to what you describe:
Screenshot 2022-01-05 at 17 22 20

Perhaps a future improvement we make to this?

dominickendrick:
Yeah looks good, maybe this will be picked up by the redesign anyway ?

ob6160:
> Yeah looks good, maybe this will be picked up by the redesign anyway ?

Very possibly! Either way, changing it to look like Figma/the tube map style seems like a big improvement to my untrained eyes, so I hope it is picked up

PR #1317, opened 2021-12-24
Sanatise the routing config/initial state

coldlink:
## What does this change?

In some cases it might be possible for a user to insert uncontrolled state data into the initial state object that we render as a string to be parsed by the browser. This could potentially allow for possible XSS attacks.

By sanatising the configuration/state as we render it, we can make sure that any dangerous strings are safely dealt with.

To do this we use the popular [`serialize-javascript`](https://github.com/yahoo/serialize-javascript) library with the [`isJSON`](https://github.com/yahoo/serialize-javascript#optionsisjson) option.


ob6160 (approved):
🔒 👍

QuarpT (approved):

PR #1349, opened 2022-01-05
Use GitHub Actions for CI

coldlink:
## What does this change?

Switches CI from TeamCity to GitHub Actions (with the use of https://github.com/guardian/actions-assume-aws-role).

There are a few benefits to using GHA individually:
- No build queue
- No need to manage `node` docker versions
- No need to globally install `yarn`
- Build logs are co-located with the PR (via the Checks tab)
- Collectively, use of GHA means we don't have to maintain our own CI infrastructure.

In RiffRaff the build numbers will count up from 5000 now, as the teamcity builds are currently in the 4000s.

To help perform the migration, I used [migration doc](https://docs.google.com/document/d/1Xz-gzptICIza0oio0Kza3LWg2a6n_015o0MPu0LGfuU/view#) and example prs from [Prism](https://github.com/guardian/prism/pull/259) and [tag-janitor](https://github.com/guardian/tag-janitor/pull/213).

## How to test

`identity:identity-gateway` should appear as a deployable project in Riff-Raff. It should be successfully deployed to CODE.

https://riffraff.gutools.co.uk/deployment/view/efa79594-2d2f-4b0d-8339-3e591bf3873f

## Pre merge
- Change the branch protection rules to remove TeamCity from required, and add the CI Github Action to required

## Post merge
- Ask people to rebase their branches on `main` to deploy to `CODE`/`PROD` going forward
- Remove project from TeamCity


ci.sh
```diff
@@ -2,7 +2,18 @@
 
 set -ae
 
+if [[ ! -z "${TEAMCITY_VERSION}" ]]; then
+  echo "Running in TeamCity. Nope!"
```
ob6160, line 6: Haha brutal 😆

ob6160 (approved):

.github/workflows/ci.yml
```diff
@@ -0,0 +1,20 @@
...
+      id-token: write
+      contents: read
+    steps:
+      - uses: actions/checkout@v2
+      - uses: guardian/actions-assume-aws-role@v1.0.0
+        with:
+          awsRoleToAssume: ${{ secrets.GU_RIFF_RAFF_ROLE_ARN }}
+      - uses: actions/setup-node@v2
+        with:
+          node-version: '16'
```
akash1810, line 18: https://github.com/actions/setup-node/pull/338 allows you to specify a `node-version-file`, which makes it easier to keep things inline.

ci.sh
```diff
@@ -2,7 +2,18 @@
...
+  exit 0
+fi
```
akash1810, line 8: Not needed, but is a good marker to see which PR you were inspired by 😄 .

.github/workflows/ci.yml
coldlink: Thanks! Even better 😃

ci.sh
coldlink: Haha, I'll remove it.

ci.sh
```diff
@@ -2,7 +2,13 @@
...
 
+# Don't remove this line as it's used to number the build based on the
+# last build number in teamcity
+LAST_TEAMCITY_BUILD=5000
+export GITHUB_RUN_NUMBER=$(( $GITHUB_RUN_NUMBER + $LAST_TEAMCITY_BUILD ))
+
 yarn install --frozen-lockfile
 yarn lint
 yarn test
 yarn build
```
dominickendrick, line 13: Do we also want to add `yarn tsc` in here, as its been removed from the other action ...

coldlink: `yarn test` actually already runs `yarn tsc`

https://github.com/guardian/gateway/blob/85ec804a51aa1b5a8c669729d7d8702bccc53079/package.json#L18

coldlink: we could separate it out into two separate commands for clarity?

coldlink: or combine the linting into the test command too as it's a kind of test?

```
"test": "yarn lint && yarn tsc && yarn test:unit", 
```

dominickendrick: Hmm, If they are all seperate commands, then when build fails, you may get more clarity about the step that is an issue, whereas if they are all combined, its harder to work out what failed ?

coldlink: internally yarn just pipes those commands to the shell, so it will still show on which command it failed on

dominickendrick: ok, so combining them seems like the right approach : )

coldlink: Code_LrZ5LxP8EA

dominickendrick (approved):

PR #1351, opened 2022-01-06
wip: feat(tracking): add server side ophan tracking;

coldlink:
## Accessibility


src/server/lib/ophan.ts
```diff
@@ -0,0 +1,47 @@
...
+      logger.warn(`Ophan: Failed to record Ophan event`, error);
+    });
+  } else {
+    logger.warn(`Ophan: Missing bwid or viewId`);
+  }
+};
+
+export const sendOphanInteractionEventServer = (
+  interaction: OphanInteraction,
+  config: OphanConfig,
```
dominickendrick, line 46: Should this be optional with a default `{}`, it seems to be called with an empty object a few times ...

coldlink: Good point, I've made it optional.

ob6160 (approved):
nice work!

PR #1354, opened 2022-01-07
Add maintenance page

coldlink:
## What does this change?

Add a maintenance page (`/maintenance`) to use for all of `profile.theguardian.com` while the identity database is being updated.

Returns page with 503 status as that's what MDN recommends: https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/503
chrome_i9AnyWKwk8

## TODO
- [x] Copy
- [x] Design
- [x] [Identity Platform PR](https://github.com/guardian/identity-platform/pull/479)
- [x] Testing on CODE

## Image

60929d168594f80039336501-yzjpkbcecw chromatic com_iframe html_id=pages-maintenancepage--default args= viewMode=story(iPad Mini)

### DIFF
--- a/src/client/pages/MaintenancePage.stories.tsx
+++ b/src/client/pages/MaintenancePage.stories.tsx
@@ -1,0 +1,12 @@
+import React from 'react';
+import { Meta } from '@storybook/react';
+
+import { MaintenancePage } from './MaintenancePage';
+
+export default {
+  title: 'Pages/MaintenancePage',
+  component: MaintenancePage,
+  parameters: { layout: 'fullscreen' },
+} as Meta;
+
+export const Default = () => <MaintenancePage />;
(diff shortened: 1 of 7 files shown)


src/server/routes/core.ts
```diff
@@ -20,4 +22,12 @@ router.get('/healthcheck', (_, res: Response) => {
   res.status(200).send('200 OK');
 });
 
+router.use('/maintenance', (_, res: ResponseWithRequestState) => {
```
QuarpT, line 25: I think the core routes include cache headers

Can you add this to uncached routes instead?
https://github.com/guardian/gateway/blob/main/src/server/routes/index.ts#L22

coldlink: Refactored to add to uncached routes

ob6160 (approved):
looks good!

PR #1355, opened 2022-01-07
Integrate with Sentry for client side logging in Gateway

ob6160:
## What does this change?
We'd like some visibility on client side Javascript errors in Gateway now that we have an integration with reCAPTCHA and a growing amount of critical JS logic tied to successfully logging in and registering on the Guardian.

This PR adds integration with the Sentry logging service so that we report successful sessions, client side errors and failed reCAPTCHA checks.

Sentry is disabled by default on DEV environments and is set to a sample rate of 0.2. This PR also uses the `GITHUB_RUN_NUMBER` environment variable set in `ci.sh` (https://github.com/guardian/gateway/blob/main/ci.sh) so that we can update Sentry with a release number to tie errors to specific releases.

The  public `SENTRY_DSN` parameter/env variable is also introduced so that we can connect to Sentry successfully to post logs.

Sentry has been configured to only accept requests from: `profile.theguardian.com profile.code.dev-theguardian.com theguardian.com`

## Testing
I simulated a reCAPTCHA error in CODE and confirmed that it is successfully logged in the Sentry dashboard. Successful sessions are also recorded.


ob6160:
[Finish client side logging integration with Sentry on Gateway](https://trello.com/c/zfkOMcTy/3728-finish-client-side-logging-integration-with-sentry-on-gateway)

src/client/static/hydration.tsx
```diff
@@ -15,8 +17,21 @@ export const hydrateApp = () => {
...
 
-  const { abTesting: { mvtId = 0, forcedTestVariants = {} } = {} } =
-    clientState;
+  const {
+    abTesting: { mvtId = 0, forcedTestVariants = {} } = {},
+    sentryConfig: { stage, build, dsn },
+  } = clientState;
+
+  // Disable Sentry client side logging during local development.
+  if (stage !== 'DEV') {
```
coldlink, line 26: rather than disabling on dev, I would suggest enable it if the `dsn` is supplied, as we'd want to replicate as much as possible between DEV/CODE/PROD

coldlink (commented):
looking good, have left a few small comments

.github/workflows/cypress-mocked.yml
```diff
@@ -39,6 +39,8 @@ jobs:
           OKTA_CLIENT_SECRET: oktaclientsecret
           LOGGING_KINESIS_STREAM: ''
           EC2_INSTANCE_ID: ''
+          # This is a public URL+key sent to Sentry on the client side
+          SENTRY_DSN: 'https://fa6dbd97a7af4ea892d811250bb14b03@o14302.ingest.sentry.io/6138895'
```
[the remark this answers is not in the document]
coldlink: actually, i'm guessing for cypress tests we don't want to run sentry logging, especially on the mocked tests, might be worth just setting this to blank now you've made the `dsn` change

src/client/lib/sentryLogger.ts
```diff
@@ -0,0 +1,41 @@
...
+    case LogLevel.INFO:
+      return Sentry.Severity.Info;
+    case LogLevel.WARN:
+      return Sentry.Severity.Warning;
+    default:
+      return Sentry.Severity.Log;
+  }
+};
+
+class SentryLogger extends BaseLogger {
```
coldlink, line 18: Would `ClientLogger` make more sense to differentiate between sentry and winston, and where it should be used?

src/client/components/MainForm.tsx
```diff
@@ -116,6 +117,10 @@ export const MainForm = ({
       const recaptchaCheckFailed =
         recaptchaState?.error || recaptchaState?.expired;
 
+      if (recaptchaCheckFailed) {
+        Sentry.captureMessage('Recaptcha check failed', Sentry.Severity.Info);
```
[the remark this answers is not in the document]
ob6160: I've made a similar abstraction to the one we use for Winston for Sentry.

 I've taken the common parts out into an abstract class: https://github.com/guardian/gateway/blob/090aeb01d9c9fd9af077c2290f9fdf59db406314/src/shared/lib/baseLogger.ts

and then used that here for Winston:
https://github.com/guardian/gateway/pull/1355/files#diff-d20b558e79b8583d3e04655a9c7f1538d4cc55b577a8316770343a71cc77a6a2R81

.. and here for Sentry:
https://github.com/guardian/gateway/blob/090aeb01d9c9fd9af077c2290f9fdf59db406314/src/client/lib/sentryLogger.ts

src/shared/lib/baseLogger.ts
```diff
@@ -0,0 +1,21 @@
+import { LogLevel, Logger } from '../model/Logger';
```
coldlink, line 1: i'm being pedantic 

```suggestion
import { LogLevel, Logger } from '@/shared/model/Logger';
```

src/server/lib/winstonLogger.ts
```diff
@@ -77,7 +78,7 @@ const loggingOptions: InspectOptions = {
 const formatLogParam = (message?: any) =>
   formatWithOptions(loggingOptions, message);
 
-export const logger: Logger = {
+class WinstonLogger extends BaseLogger {
```
coldlink, line 81: Same as about this, could this be called `ServerLogging` to help differentiate?

.github/workflows/cypress-mocked.yml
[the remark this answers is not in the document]
ob6160: yeah I think that's the way to go :)

src/client/lib/sentryLogger.ts
ob6160: have changed both to follow a client/server naming convention

coldlink (approved):

PR #1346, opened 2022-01-04
4th January 2022 | Yet more dependency updates

coldlink:
## What does this change?

- Updates dependencies to latest versions where possible

## Notes

- `@babel/preset-react` not updated as this causes an issue with Storybook and chromatic
- `helmet` is a major upgrade with breaking changes that affect our project, so will separate out into it's own PR
- `http-errors` a major update which causes error when deploying to code
```
  /home/mahesh/Downloads/identity-gateway/node_modules/http-errors/node_modules/toidentifier/index.js:27
    .split(' ')
     ^

TypeError: Cannot read properties of undefined (reading 'split')
    at toIdentifier (/home/mahesh/Downloads/identity-gateway/node_modules/http-errors/node_modules/toidentifier/index.js:27:6)
    at forEachCode (/home/mahesh/Downloads/identity-gateway/node_modules/http-errors/index.js:269:16)
    at Array.forEach (<anonymous>)
    at populateConstructorExports (/home/mahesh/Downloads/identity-gateway/node_modules/http-errors/index.js:265:9)
    at Object.<anonymous> (/home/mahesh/Downloads/identity-gateway/node_modules/http-errors/index.js:31:1)
    at Module._compile (node:internal/modules/cjs/loader:1101:14)
    at Object.Module._extensions..js (node:internal/modules/cjs/loader:1153:10)
    at Module.load (node:internal/modules/cjs/loader:981:32)
    at Function.Module._load (node:internal/modules/cjs/loader:822:12)
    at Module.require (node:internal/modules/cjs/loader:1005:19)
```
- `express` update causes errors when deploying to code
```
[...]

PR #1374, opened 2022-01-10
Update deps where possible (10th Jan '22)

ob6160:
## What does this change?
Upgrades dependencies where possible.

## Notes
`@babel/preset-react`, `helmet` and `http-errors` to be upgraded separately as detailed in #1346 

Tested on CODE, all looks good.


coldlink (approved):
Glad we're able to update `express` now, seems moving to GHA has meant that dependencies are working properly.

PR #1375, opened 2022-01-11
Support Jobs user registration in Gateway

ob6160:
## Background
A Jobs account is a user account that has been added to the GRS user group. Users who are signed in have to join the GRS group or sign out before they can access Guardian Jobs.

Users are currently added to the GRS group (in identity-frontend) under these conditions:

* When a logged in user not belonging to the GRS group navigates to the Jobs site and gives consent.
* When a user originating from Jobs registers an account.

In both cases, the user is added to the GRS group through a POST request made on the client side to the /agree/GRS endpoint in IDAPI.

Following this process, when the user who now belongs to GRS next visits jobs.theguardian.com, their madgex account is created and the welcome email is sent.

## What does this change?
This PR introduces support for adding users to the `GRS` group if the `clientId` query parameter is set to `jobs`.

It builds upon the prior work in #1273 to persist the `clientId` query parameter across the different pages on Gateway.

To add the users to the GRS group, the `/user/me/group/:groupCode` endpoint in IDAPI is called when the password is set by users originating from the welcome flow using the `SC_GU_U` cookie they are granted

Two changes were made in IDAPI (https://github.com/guardian/identity/pull/2045) to support passing the `clientId` through as a query param to the `/guest` endpoint and the `/user/send-account-verification-email` endpoints.

This ensures that on the registration confirmation email and any subsequent resend requests, the `clientId` will be included in the generated link.

## How to test
A separate PR has been raised against identity-platform: https://github.com/guardian/identity-platform/pull/482, this will let us test with `clientId=jobs` in PROD, by setting `gateway=true` as an extra query param in the request.

Users should be added to the GRS group when originating from jobs.theguardian.com and be redirected without having to go through the full confirmation flow.

### DIFF
--- a/src/server/controllers/changePassword.ts
+++ b/src/server/controllers/changePassword.ts
@@ -93,7 +93,7 @@ export const setPasswordController = (
 
       // When a jobs user is registering, we'd like to add them to the GRS group.
       //
-      // We'd like to do this so they aren't shown a confirmation page when-
+      // Once they belong to this group, they aren't shown a confirmation page when-
       // they first visit the jobs site.
       //
       // If the SC_GU_U cookie exists, we try to add the user to the group.
--- a/src/server/controllers/changePassword.ts
+++ b/src/server/controllers/changePassword.ts
@@ -21,6 +21,7 @@ import { PasswordRoutePath, RoutePaths } from '@/shared/model/Routes';
 import { PasswordPageTitle } from '@/shared/model/PageTitle';
 import { validatePasswordField } from '@/server/lib/validatePasswordField';
 import { addQueryParamsToPath } from '@/shared/lib/queryParams';
+import { addToGroup, GroupCode } from '@/server/lib/idapi/user';
 
 export const setPasswordController = (
   path: PasswordRoutePath,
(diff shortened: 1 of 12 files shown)


ob6160:
[Register for a jobs account successfully through Gateway](https://trello.com/c/iYXItXFT/3701-register-for-a-jobs-account-successfully-through-gateway)

src/server/controllers/changePassword.ts
```diff
@@ -77,6 +81,25 @@ export const setPasswordController = (
...
+      //
+      // Once they belong to this group, they aren't shown a confirmation page when-
+      // they first visit the jobs site.
+      //
+      // If the SC_GU_U cookie exists, we try to add the user to the group.
+      // If the cookie doesn't exist for some reason, we log the incident.
+      if (clientId === 'jobs' && path === '/welcome') {
+        const SC_GU_U = cookies?.values.find(({ key }) => key === 'SC_GU_U');
+        if (SC_GU_U) {
+          addToGroup(GroupCode.GRS, req.ip, SC_GU_U.value);
```
dominickendrick, line 95: do we need to `await` here ? Could the request lifecycle complete before the API call to Identity is complete ?

dominickendrick (approved):
Looks good, I've left a comment

PR #1378, opened 2022-01-11
Registration with Okta behind feature flag

rebecca-thompson:
## What does this change?
This adds registration with Okta and puts it behind a feature flag which is currently turned off in PROD. It should have no impact on the current registration process.

The process for registering a user is:

1. create, activate or reactivate a user in Okta when a valid email is entered into the registration form (this triggers an email containing an activation token)
2. clicking on the link will take you to the first step of the onboarding journey. In the backend, gateway exchanges the activation token for a state token via the `/authn` endpoint of Okta's Authentication API.
3. when a user enters a valid password, the new password along with the state token is passed to Okta's `/reset_password` endpoint which sets the user's password.
4. if any of the tokens expire throughout the process, resending the email will restart the process

## Excluded in this PR
This PR has excluded the reset password flows and authentication, both to be added in later PRs. This means that registration will fail for ACTIVE users (rather than sending an email to reset their password) and the call to the `/authorize` endpoint to sign the user in after they see a password on the onboarding flow will result in a 404 until authentication with Okta is merged/built. More details on Authentication with Okta can be found in this PR - https://github.com/guardian/gateway/pull/638

## How to test
This can be deployed in the CODE environment and can be triggered by passing the parameter flag `useOkta=true`


rebecca-thompson:
[Build the Okta post activation registration step (set password page etc)](https://trello.com/c/9uOfMahu/3564-build-the-okta-post-activation-registration-step-set-password-page-etc)

src/server/controllers/checkPasswordToken.ts
```diff
@@ -12,63 +12,133 @@ import { renderer } from '@/server/lib/renderer';
...
 
 export const checkPasswordTokenController = (
   path: PasswordRoutePath,
   pageTitle: PasswordPageTitle,
 ) =>
   handleAsyncErrors(async (req: Request, res: ResponseWithRequestState) => {
-    let requestState = res.locals;
-    const { token } = req.params;
+    const { useOkta } = res.locals.queryParams;
+    if (okta.registrationEnabled && useOkta && path === '/welcome') {
```
dominickendrick, line 26: Maybe this controller is now getting a bit long with these extra okta changes added. Could it be worth moving this ```if (okta.registrationEnabled && useOkta && path === '/welcome') { ... } else { ... }``` logic up into the `routes/setPassword.tx` and make a seperate Okta only controller file ? Then this file can just be dedicated to Identity stuff?

src/server/models/okta/Error.ts
```diff
@@ -0,0 +1,87 @@
...
+  errorSummary: string;
+}
+
+export type ErrorCode = 'E0000001' | 'E0000007' | 'E0000016' | 'E0000038';
+
+export class OktaError extends Error {
+  name: string;
+  code?: ErrorCode;
+  constructor(message?: string, name = 'OktaError', code?: ErrorCode) {
+    super(message);
```
dominickendrick, line 19: Does `Object.setPrototypeOf(this, new.target.prototype);` need to be set after calling `super()` see - https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-2.html#support-for-newtarget I'm not sure if its applicable

src/server/lib/okta/register.ts
```diff
@@ -0,0 +1,64 @@
...
+          throw new OktaError(
+            `Okta registration failed with unaccepted Okta user status: ${user.status}`,
+          );
+      }
+    } else {
+      throw error;
+    }
+  }
+};
+
```
dominickendrick, line 45: Is it worth explaining in comments why the user could have different status? The broader question is what is the best way to document the Okta registration integration.

src/shared/model/OktaQueryParams.ts
```diff
@@ -0,0 +1,9 @@
+import { StringifiableRecord } from 'query-string';
+
+/**
+ * Okta are query parameters
```
ob6160, line 4: wording of comment might need fixing slightly

src/server/lib/okta/register.ts
```diff
@@ -0,0 +1,64 @@
...
+  OktaError,
+  ResourceAlreadyExistsError,
+} from '@/server/models/okta/Error';
+
+export const register = async (email: string): Promise<User> => {
+  try {
+    return await createUser(email);
+  } catch (error) {
+    if (error instanceof ResourceAlreadyExistsError) {
+      const user: User = await fetchUser({ id: email });
```
ob6160, line 18: do we need the type assertion here?

src/server/lib/okta/api/errors.ts
```diff
@@ -0,0 +1,89 @@
...
+export const handleErrorResponse = async (response: Response) => {
+  const error = await extractErrorResponse(response);
+  if (error.errorId && error.errorSummary) {
+    const { errorCode, errorSummary } = error;
+    switch (errorCode) {
+      case 'E0000001': {
+        if (error.errorCauses) {
+          const { errorCauses } = error;
+          if (
+            errorCauses.some((cause) =>
```
ob6160, line 25: wondering if it might help readability here if this was abstracted into a method called `errorCausePresent` or something similar because it's repeated a lot?

src/server/lib/fetch.ts
```diff
@@ -6,7 +6,7 @@
  * Solution taken from: https://github.com/node-fetch/node-fetch/issues/1279#issuecomment-915063354
  */
 
-import { RequestInfo, RequestInit, Response } from 'node-fetch';
+import type { RequestInfo, RequestInit, Response } from 'node-fetch';
```
ob6160, line 9: not seen `import type` before, just wondering why we're changing it here

src/server/models/okta/Token.ts
```diff
@@ -0,0 +1,10 @@
+import { User } from '@/server/models/okta/User';
+
+export interface Token {
+  stateToken?: string;
+  sessionToken?: string;
+  expiresAt: string;
+  _embedded?: {
```
ob6160, line 7: wondering if a small doc comment on this or link to the docs might be useful to explain what _embedded is

src/server/controllers/checkPasswordToken.ts
rebecca-thompson: That's funny, because the first time I started adding to this controller I ended up doing a refactoring PR which broke it up from one controller into two and moved some code out into libs (https://github.com/guardian/gateway/pull/1275), so it's a lot smaller than it used to be! I can refactor it further if you think that it's still too big

rebecca-thompson: Personally I think having them in the same file is less error-prone because you can see all the logic in one place and if something needs to be changed in both scenarios, it is less likely to be missed

src/server/lib/okta/register.ts
rebecca-thompson: I think it would be best to put this information in `/docs` along with the user state diagram when it's ready. It's still a bit fluid for now while we're finalising the forgot password flow

src/server/models/okta/Error.ts
rebecca-thompson: I'm not sure, but the current IDAPI errors also extend Error without including this line, so if it were an issue it would already be affecting what is currently in prod

dominickendrick: 👍

src/server/controllers/checkPasswordToken.ts
dominickendrick: 👍

src/server/lib/fetch.ts
rebecca-thompson: This is related to `node-fetch` being a pure ESM package and us importing it asynchronously. There is a comment here on the same thread linked in the comments in this file that explains - https://github.com/node-fetch/node-fetch/issues/1279#issuecomment-915061092

src/shared/model/OktaQueryParams.ts
rebecca-thompson: have updated

src/server/lib/okta/api/errors.ts
rebecca-thompson: I have refactored this into a function

src/server/models/okta/Token.ts
rebecca-thompson: have linked do the docs in a comment

dominickendrick (approved):
Excellent work! Great to see this all come together

src/server/lib/fetch.ts
ob6160: ahh thanks, that's really interesting

ob6160 (approved):
great work!

PR #1380, opened 2022-01-14
Fix flakey reCaptcha mocked test

rebecca-thompson:
## What does this change?
This mocked test fails roughly 1 in 10 times. It times out when trying to find the text on the page related to the second reCaptcha failure message.

My hypothesis is that reCaptcha's behaviour is non-deterministic, so we expect it to fail twice but only mock one call to the reCaptcha api whereas it may be that it occasionally makes a second call and passes the second reCaptcha check, causing the test to fail as the second failure message never displays.

The fix is to mock all calls to the reCaptcha api and only test for failure messages. This is in line with tests that we have for other flows (welcome, sign_in, etc).

### DIFF
--- a/cypress/integration/mocked/verify_email.spec.ts
+++ b/cypress/integration/mocked/verify_email.spec.ts
@@ -60,7 +60,7 @@ describe('Verify email flow', () => {
   });
 
   context('Verify email', () => {
-    it('successfuly verifies the email using a token and sets auth cookies', () => {
+    it('successfully verifies the email using a token and sets auth cookies', () => {
       // mock validation success response (200 with auth cookies)
       cy.mockNext(200, authCookieResponse);
 
@@ -87,7 +87,7 @@ describe('Verify email flow', () => {
       cy.url().should('include', '/consents');
     });
 
-    it('successfuly verifies the email using a token and sets auth cookies, and preserves encoded return url', () => {
+    it('successfully verifies the email using a token and sets auth cookies, and preserves encoded return url', () => {
       // mock validation success response (200 with auth cookies)
       cy.mockNext(200, authCookieResponse);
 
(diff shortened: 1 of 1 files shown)


ob6160 (approved):
this looks great, thanks for taking the time to look into this!

cypress/integration/mocked/verify_email.spec.ts
```diff
@@ -284,25 +284,18 @@ describe('Verify email flow', () => {
...
       cy.contains(VerifyEmail.CONTENT.SEND_LINK).click();
       cy.contains('Google reCAPTCHA verification failed. Please try again.');
 
+      // show extended reCaptcha message on the second failure
       cy.contains(VerifyEmail.CONTENT.SEND_LINK).click();
       cy.contains('Google reCAPTCHA verification failed.');
       cy.contains('If the problem persists please try the following:');
       cy.contains('userhelp@');
-
-      cy.contains(VerifyEmail.CONTENT.SEND_LINK).click();
```
ob6160, line 302: I think this test was checking to see if we can still put the form through when reCAPTCHA eventually succeeds, by removing this we're no longer checking that. Maybe we could intercept it and show it as a successful check after the second message is shown so we can this final assertion in there too?

ob6160 (commented):
just one comment after spotting this

cypress/integration/mocked/verify_email.spec.ts
rebecca-thompson: This would reintroduce the flakiness as there is no way to know in advance if reCaptcha will make 2 or 3 calls to the API

ob6160: aha, I see. probably not a big deal as we check the form submission with recaptcha in other places 👍

ob6160 (approved):

PR #1382, opened 2022-01-17
Polyfill `Object.fromEntries` for browsers without support

coldlink:
## What does this change?

Userhelp reported that some browsers were unable to sign in/reset password (effectively not being able to use any forms on gateway) as they were seeing a recaptcha error.

We determined that this was due to the console error showing that `Object.fromEntries is not a function`, stopping the Javascript bundle from fully executing, meaning that recaptcha was not initialising.

We use [polyfill.io](https://polyfill.io/) to polyfill this method, polyfill.io works by only polyfilling the method if it's not supported by the browser.

Using the same method as https://github.com/guardian/frontend/pull/24473

## Tested
- [x] CODE


src/server/lib/renderer.tsx
```diff
@@ -133,14 +133,17 @@ export const renderer: <P extends RoutePaths>(
         <link rel="icon" href="https://static.guim.co.uk/images/${favicon}">
         <title>${pageTitle} | The Guardian</title>
         <script>window.gaUID = "${gaUID.id}"</script>
+
+        <script src="https://polyfill.io/v3/polyfill.min.js?features=Object.fromEntries" defer></script>
```
ob6160, line 137: wondering if we should use this URL or the assets.guim one used in the PR Tom linked: `https://assets.guim.co.uk/polyfill.io/v3/polyfill.min.js` https://github.com/guardian/frontend/pull/24473/files#diff-8b9cd752fdbccf643a1ce3fe63206eb1ebf9295f81bc915f6d7cb8fa6589e426R1

coldlink: I suppose `assets.guim.co.uk` is behind our own caching policies, so we can use that.

ob6160 (approved):

PR #1405, opened 2022-01-18
Remove PLAY_SESSION_2 cookie support (identity-frontend)

ob6160:
## What does this change?
This PR removes our support for the `PLAY_SESSION_2` cookie, set by `identity-frontend` to store the users' email during registration.

This can be safely carried out now that Gateway has been released to 100% of users for sign in and registration.


rebecca-thompson (approved):
Looks good 🎉

PR #1409, opened 2022-01-20
Fix: IE11 polyfill and transpilation issues

coldlink:
## What does this change?

This PR fixes two issues with IE11 introduced after the safari fixes in #1403 and #1406 :

### 1.
We were mixing polyfills provided by polyfill.io and corejs causing an uncaught error to be thrown in IE11. Ideally we want to move all polyfills to using polyfill.io rather than corejs as we can target only features that the users browser does not support rather than providing polyfills for all features using corejs.

chrome_JId4TGVBOk

As a temp migration measure we set the `type="module"` attribute on the polyfill.io script tag so that it is not executed by legacy browsers such as IE11. We'll migrate fully to using polyfill.io for all browsers after this PR as it's a bit more work to implement.

We also add all supported features we transpile with to polyfill.io namely: `es2015` through `es2021` and `fetch`. This includes polyfills for `Object.fromEntries` and `Promise.prototype.finally`.

### 2.
Although we provide a polyfill for `Promise.finally`, our usage of it causes a `Object doesn't support property or method 'finally'` only in the case of the `PasswordForm.tsx`, after investigation it seems that the transpilation is behaving in a strange way for this promise in legacy browsers, where as in other cases `.finally` behaves as expected.

chrome_AhOAJsvWQH

We solve this issue by refactoring the breached check to use async/await avoiding the need for the `.finally` in this scenario.

We also add sentry logging for checking when the breached password check fails for any reason.

## Tested
- [x] CODE
  - [x] Browserstack IE11 Win 10
  - [x] Browserstack Safari 10
  - [x] Browserstack Safari 11
  - [x] Modern browser

### DIFF
--- a/src/client/components/PasswordForm.tsx
+++ b/src/client/components/PasswordForm.tsx
@@ -28,6 +28,7 @@ import AwesomeDebouncePromise from 'awesome-debounce-promise';
 import { AutoRow, passwordFormSpanDef } from '@/client/styles/Grid';
 import { MainForm } from '@/client/components/MainForm';
 import { trackFormFocusBlur, trackFormSubmit } from '@/client/lib/ophan';
+import { logger } from '../lib/clientSideLogger';
 
 type Props = {
   submitUrl: string;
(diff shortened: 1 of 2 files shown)


ob6160 (approved):
looks good! hopefully can go fully polyfill.io soon 🤞

PR #1441, opened 2022-01-24
Update GitHub action workflows

coldlink:
## What does this change?

- Set cypress `NODE_ENV` to production to use production javascript
bundles, hopefully speeding up development
  - cypress was creating development bundles, which are large, take longer to build, and download
- add `actions/setup-node` to use defined node version and yarn cache
  - hopefully speeds actions up a little too
- Set dependabot to only open 10 PRs max
  - this is to reduce spam, yet enough to force us to update weekly


ob6160 (approved):
great improvements, thanks!

PR #1442, opened 2022-01-27
Test adding different newsletters into the onboarding flow

liamness:
## What does this change?

Replaces #1282, and currently branched off `mh/update-onboarding-flow`. This sets up an A/B test where 50% of users will see different newsletter options when going through the onboarding flow.

## How to test

Set a `GU_mvt_id cookie` to `1` to see the changes being tested. Set it to `2` to see the control group behaviour, which shows the same list as is currently used.

Unsure how best to test region-specific behaviour (different regions have different newsletters). Locally I have been hardcoding the relevant region to test.

I suppose we look at which variant leads to the most signups?

### Control (US)

image

### Variant (US)

image

## Accessibility


src/shared/model/experiments/tests/onboarding-newsletters.ts
```diff
@@ -0,0 +1,30 @@
+import { ABTest } from '@guardian/ab-core';
+
+export const TEST_ID = 'OnboardingNewslettersTest';
+export const CONTROL_VARIANT = 'control';
+export const TEST_VARIANT = 'test';
+
+export const onboardingNewslettersTest: ABTest = {
+  id: TEST_ID,
+  // TODO: change start / expiry before release
```
liamness, line 9: Just commenting here to actually remind myself to change this before release.

coldlink (approved):
📰

coldlink:
In terms of the UI tests, that looks related to #1281 so I'll have a look at those first.

src/server/routes/consents.ts
```diff
@@ -152,15 +162,20 @@ export const consentPages: ConsentPage[] = [
...
-        sc_gu_u,
-      ),
-      previousPage: '/communication'.slice(1),
-    }),
+    read: async (ip, sc_gu_u, geo, newsLettersTestActive) => {
+      const newsletterMap = newsLettersTestActive
+        ? TestNewsletterMap
+        : NewsletterMap;
+      return {
+        page: '/newsletters'.slice(1),
```
ob6160, line 170: realise you didn't add this, but I'm confused by our pattern here of doing `.slice(1)` instead of just putting `newsletters` 🤔  we do it below as well

coldlink: Actually it looks like a hangover from when we did the type safety on the `Routes` object a couple of months back https://github.com/guardian/gateway/pull/1224 before we'd take the `Routes` enum, the route, and remove the `/` (`Routes.CONSENTS_COMMUNICATION.slice(1)`), however since we're not using enums anymore it does look a bit strange.

Since they're just `string` types now, i think you can just remove the `/` and `.slice(1)`. Ideally we could make these type safe too, specifically the `page` and `previousPage` properties as the are related to the `path`.

liamness: Yeah I noticed this and agree it was a bit strange. Didn't touch as it wasn't neccessarily related to my changes, but I can clean this up.

liamness: I think we could define the paths with template literal types, e.g.

```typescript
type ConsentPath = 'communication' | 'newsletters' | 'data' | 'review';
type RoutePaths = `/consents/${ConsentPath}` | '/consents/:page';
```

But might be best to leave this for another PR.

ob6160 (approved):

PR #1443, opened 2022-01-27
Show prompt to opt into supporter marketing after sign in

liamness:
## What does this change?

After sign in, this redirects users to a page prompting them to opt in to support marketing, if:

- They are in the a/b test (so we can control the rollout)
- The `returnUrl` query param is the default (this is to avoid adding extra friction into critical paths)
- They have not seen the prompt previously on the device they are using
- They have not previously consented to supporter marketing

~~Currently branched off `mh/update-onboarding-flow`.~~

Still need to do the following, hence the PR being in draft:

- [x] Implement design (needs to use bits from Michael's branch)
- [x] Add tests 

## How to test

Set `GU_mvt_id cookie` to `1` and sign in with a user who is has not already consented to supporter marketing.

### Desktop

image

### Tablet

image

# Mobile

image

## Accessibility

### DIFF
--- a/package.json
+++ b/package.json
@@ -52,6 +52,7 @@
     "@storybook/builder-webpack5": "^6.5.0-alpha.34",
     "@storybook/manager-webpack5": "^6.5.0-alpha.34",
     "@storybook/react": "^6.5.0-alpha.34",
+    "@testing-library/cypress": "^8.0.2",
     "@testing-library/jest-dom": "^5.16.2",
     "@testing-library/react": "^12.1.3",
     "@testing-library/react-hooks": "^7.0.2",
--- a/src/client/components/CommunicationCard.tsx
+++ b/src/client/components/CommunicationCard.tsx
@@ -17,6 +17,7 @@ import { greyBorderBottom, greyBorderTop } from '../styles/Consents';
 
 interface CommunicationCardProps {
   title: string;
+  titleLevel?: 1 | 2 | 3 | 4 | 5;
   body: string;
   value: string;
   checked: boolean;
@@ -116,17 +117,20 @@ const ieFlexOverflowFix = css`
 
 export const CommunicationCard: FunctionComponent<CommunicationCardProps> = ({
   title,
+  titleLevel = 3,
   body,
   value,
   image,
   checked,
 }) => {
+  const HeadingTag = `h${titleLevel}` as const;
+
   return (
     <div css={[greyBorderTop, blueBorderBottom, ieFlexOverflowFix]}>
       <div css={[communicationCardInfoContainer]}>
         <img css={img} src={image} alt={title} />
         <div css={ieFlexOverflowFix}>
-          <h3 css={communicationCardHeadingText}>{title}</h3>
+          <HeadingTag css={communicationCardHeadingText}>{title}</HeadingTag>
           <div css={communicationCardBodyContainer}>
             <p css={communicationCardBodyText}>{body}</p>
           </div>
--- a/src/client/lib/consentsTracking.ts
+++ b/src/client/lib/consentsTracking.ts
@@ -77,6 +77,7 @@ export const onboardingFormSubmitOphanTracking = (
   switch (`/${page}`) {
     case '/communication':
     case '/data':
+    case '/signin/success':
       return consentsFormSubmitOphanTracking(inputElems, pageData);
     case '/newsletters':
       return newslettersFormSubmitOphanTracking(inputElems, pageData);
--- a/src/client/pages/SignInSuccess.tsx
+++ b/src/client/pages/SignInSuccess.tsx
@@ -1,18 +1,174 @@
 import React, { useContext } from 'react';
-import { Checkbox } from '@guardian/source-react-components';
+import { css } from '@emotion/react';
+import {
+  body,
+  brand,
+  from,
+  headline,
+  neutral,
+  space,
+  titlepiece,
+  until,
+} from '@guardian/source-foundations';
+import { Button, Checkbox } from '@guardian/source-react-components';
+
 import { Header } from '@/client/components/Header';
 import { Footer } from '@/client/components/Footer';
 import { GeoLocation } from '@/shared/model/Geolocation';
 import { Consent } from '@/shared/model/Consent';
-import { QueryParams } from '@/shared/model/QueryParams';
 import { buildUrlWithQueryParams } from '@/shared/lib/routeUtils';
 import { CsrfFormField } from '@/client/components/CsrfFormField';
 import { onboardingFormSubmitOphanTracking } from '@/client/lib/consentsTracking';
 import { ClientState } from '@/shared/model/ClientState';
 import { ClientStateContext } from '@/client/components/ClientState';
+import {
+  gridItem,
+  gridRow,
+  gridRowPadding,
+  innerGridRow,
+} from '@/client/styles/Grid';
+import { EnvelopeImage } from '@/client/components/EnvelopeImage';
 
-export type SignInSuccessProps = {
-  queryParams: QueryParams;
+const titleSpanDefinition = {
+  DESKTOP: { start: 2, span: 10 },
+  LEFT_COL: { start: 2, span: 11 },
+  WIDE: { start: 2, span: 11 },
+};
+
+const envelopeSpanDefinition = {
+  DESKTOP: { start: 9, span: 3 },
+  LEFT_COL: { start: 11, span: 3 },
+  WIDE: { start: 13, span: 3 },
+};
+
+const formSpanDefinition = {
+  DESKTOP: { start: 2, span: 6 },
+  LEFT_COL: { start: 2, span: 6 },
+  WIDE: { start: 2, span: 6 },
+};
+
+const titleWrapper = css`
+  background: ${brand[400]};
+  border-top: 1px solid ${brand[600]};
+
+  ${from.desktop} {
+    border-top: none;
+    margin-top: 1px;
+  }
+`;
+
+const autoMargin = css`
+  margin: 0 auto;
+`;
+
+const title = css`
+  ${titlepiece.small()}
+  ${gridItem(titleSpanDefinition)}
+  font-size: 36px;
+  color: ${neutral[100]};
+  margin: 60px 0 70px;
+
+  ${from.desktop} {
+    ${titlepiece.large()}
+    margin: 60px 0 130px;
+  }
+`;
+
+const formStyles = css`
+  ${gridItem(formSpanDefinition)}
+
+  ${until.desktop} {
+    padding: 0 ${space[5]}px;
+  }
+`;
+
+const formBackground = css`
+  background: ${neutral[97]};
+  flex-grow: 1;
+`;
+
+const formWrapper = css`
+  background: ${neutral[100]};
+  border: 1px solid ${neutral[86]};
+  position: relative;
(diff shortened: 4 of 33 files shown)


src/client/lib/consentsTracking.ts
```diff
@@ -77,6 +77,7 @@ export const onboardingFormSubmitOphanTracking = (
   switch (`/${page}`) {
     case '/communication':
     case '/data':
+    case 'signin/success':
```
coldlink, line 80: missing the `/` at the start here i think

```suggestion
    case '/signin/success':
```

src/client/pages/SignInSuccessPage.tsx
```diff
@@ -0,0 +1,18 @@
+import React, { useContext } from 'react';
```
coldlink, line 1: can we  get a storybook for this page?

src/server/lib/experiments.ts
```diff
@@ -0,0 +1,44 @@
+import { Request } from 'express';
+import { parse, stringify } from 'query-string';
+import ms from 'ms';
+
+import { ResponseWithRequestState } from '@/server/models/Express';
+
+const RAN_EXPERIMENTS_COOKIE_NAME = 'ranExperiments';
```
coldlink, line 7: in general we preface cookies with `GU_`
as a possible suggestion maybe something like:
```suggestion
const RAN_EXPERIMENTS_COOKIE_NAME = 'GU_gateway_experiments';
```

src/server/lib/experiments.ts
```diff
@@ -0,0 +1,44 @@
...
+      [experimentId]: Date.now(),
+    };
+  } else {
+    // Using destructuring to omit property, so extracted variable is unused
+    // eslint-disable-next-line @typescript-eslint/no-unused-vars
+    const { [experimentId]: omit, ...otherExperiments } = ranExperiments;
+    newExperiments = otherExperiments;
+  }
+  const newValue = stringify(newExperiments);
+  res.cookie(RAN_EXPERIMENTS_COOKIE_NAME, newValue, { maxAge: ms('1yr') });
```
coldlink, line 43: In terms of the cookie options I also suggest adding on the `httpOnly` (if not being read on the clientside), `secure`, and `sameSite: "strict"` flags

coldlink (commented):
looks pretty good so far, just a few minor suggestions

src/server/lib/experiments.ts
liamness: It's not been read on the client currently, but I'm thinking there is a possibility that it might be in the future though, and we wouldn't be able to retroactively change this. So I'd err towards allowing it to be read on the client, unless the information is considered sensitive for some reason (I don't think it is).

src/client/lib/consentsTracking.ts
liamness: Whoops, adjusted now

src/server/lib/experiments.ts
liamness: Have taken your suggestion.

src/server/lib/experiments.ts
coldlink: We've generally gone with `httpOnly` cookies in the past as in general for gateway we've made sure that core functionality isn't reliant on client side code, so avoid reading cookies in the client.

If a client did need a value in a cookie, we'd read it on the server and pass it to the `RequestState` to propagate to the client through one of the properties on there, e.g. `pageData`, and that could be read on the client through the `ClientState`.

liamness: Ah okay, I see the rationale now. Will change.

src/server/lib/idapi/consents.ts
```diff
@@ -67,3 +68,49 @@ export const update = async (
...
+
+export const getConsentValueFromRequestBody = (
+  key: string,
+  body: { [key: string]: string },
+): boolean => {
+  if (body[key] === undefined || typeof body[key] !== 'string') {
+    return false;
+  }
+
+  switch (body[key]) {
```
ob6160, line 108: I wonder if this switch block could just be replaced with `return !!body[key];`

ob6160 (approved):
looks good, two small comments

src/server/lib/idapi/consents.ts
liamness: I don't think so, as `!!'false'` would return `true`.

This and the above function have been moved from `src/server/routes/consents.ts` btw, I haven't actually changed them at all. Just moved them here as they are needed in the sign in routes now also.

ob6160: ah yes, didn't notice it was a string. all good!

package.json
```diff
@@ -52,6 +52,7 @@
     "@storybook/builder-webpack5": "^6.5.0-alpha.34",
     "@storybook/manager-webpack5": "^6.5.0-alpha.34",
     "@storybook/react": "^6.5.0-alpha.34",
+    "@testing-library/cypress": "^8.0.2",
```
ob6160, line 55: what's this library giving us? interested in why we haven't needed it until now

liamness: It allows you to use the same queries you would be using with `@testing-library/react`. So for instance:

```js
const checkbox = cy.findByLabelText('Yes, sign me up');
```

instead of

```
const checkbox = cy
    .contains('label', 'Yes, sign me up')
    .invoke('attr', 'for')
    .then((id) => {
      cy.get('#' + id)
    })
```

If you associated a label with an input by just wrapping it, then you'd probably need to do something even more complicated.

Unlike `@testing-library/react` it doesn't give you the same nice assertion functions like `.toBeInTheDocument()` etc but I still think it's worth having, should encourage us to avoid testing implementation details, and make context switching between the jsdom and cypress tests a bit easier too hopefully

ob6160: brilliant, thanks for covering the key changes, sounds like a great addition to our testing stack

liamness:
I've used the wrong design here, moving this back into draft

liamness:
This now shares styling with the consents layout, and particularly the communication and newsletters pages. The way the styles are shared could use some work, but I was already aiming to clean that up as part of [this card](https://trello.com/c/ncPxVzLu/3810-onboarding-flow-redesign-follow-up), so haven't tried to improve matters as part of this PR.

src/client/pages/SignInSuccess.tsx
```diff
@@ -1,233 +1,99 @@
...
+              form as HTMLFormElement,
+            );
+          }}
+        >
+          <CsrfFormField />
+          <ConsentsContent>
+            <h1 css={[heading, autoRow()]}>
+              Get the latest offers sent to your inbox
+            </h1>
+            <div css={[autoRow()]}>
```
ob6160:
```suggestion
            <div css={autoRow()}>
```

src/client/pages/SignInSuccess.tsx
```diff
@@ -1,233 +1,99 @@
...
-            <div css={formWrapper}>
-              <div css={[innerGridRow, formGridRow]}>
-                <div css={envelopeImageWrapper}>
-                  <EnvelopeImage invertColors cssOverrides={envelopeImage} />
-                </div>
-                {renderForm()}
-              </div>
+      <main css={mainStyles}>
+        <form
+          css={[autoRow()]}
```
ob6160:
```suggestion
          css={autoRow()}
```

src/client/pages/SignInSuccess.tsx
```diff
@@ -1,233 +1,99 @@
...
-          <div css={[autoMargin, gridRowPadding]}>
-            <div css={formWrapper}>
-              <div css={[innerGridRow, formGridRow]}>
-                <div css={envelopeImageWrapper}>
-                  <EnvelopeImage invertColors cssOverrides={envelopeImage} />
-                </div>
-                {renderForm()}
-              </div>
+      <main css={mainStyles}>
+        <form
```
ob6160, line 63: I wonder if this can make use of our central `MainForm` component. I don't think it really needs changing, but thought I'd throw the idea in there

src/client/styles/Grid.ts
```diff
@@ -93,24 +93,27 @@ export enum MAX_WIDTH {
...
+    padding-left: ${px(padding)};
+    padding-right: ${px(padding)};
+    max-width: ${maxWidth ? px(maxWidth) : '100%'};
 `;
 
-export const innerGridRow = css`
+// this styles should be applied to an element to make it behave as
+// the grid container, it defines the css on how all the breakpoints
+// should behave
+// anything items should be added inside this container, and use
```
ob6160, line 114: overall reads well, just wondering about maybe clearing up this part: "anything items.."

src/client/pages/SignInSuccess.tsx
```diff
@@ -1,233 +1,99 @@
...
 
-const formStyles = css`
-  ${gridItem(formSpanDefinition)}
+const heading = css`
+  ${headline.small({ fontWeight: 'bold' })}
 
-  ${until.desktop} {
-    padding: 0 ${space[5]}px;
+  ${from.tablet} {
+    font-size: 32px;
```
ob6160, line 31: shame we can't use font sizes from the Source typography scale here. I think there's a 34 but no 32

ob6160 (commented):
looks good, few questions/suggestions/tweaks

src/client/pages/SignInSuccess.tsx
liamness: I should mention that `src/client/styles/Grid.ts` is actually completely unchanged now that I'm using the correct design.

src/client/pages/SignInSuccess.tsx
liamness: Intending to clean up the shared styles / logic as part of [this card](https://trello.com/c/ncPxVzLu/3810-onboarding-flow-redesign-follow-up).

src/client/pages/SignInSuccess.tsx
liamness: This is odd, this diff isn't showing as outdated, but this definitely isn't the most up to date code?

If you ever have need to check for usages along these lines, here's a regex `css=\{\[[a-zA-z]+?\]\}`. Seems there's only four times we do this elsewhere in the codebase.

src/client/styles/Grid.ts
liamness: I should mention that src/client/styles/Grid.ts is actually completely unchanged now that I'm using the correct design.

src/client/components/CommunicationCard.tsx
```diff
@@ -116,17 +117,20 @@ const ieFlexOverflowFix = css`
...
 }) => {
+  const HeadingTag = `h${titleLevel}` as const;
+
   return (
     <div css={[greyBorderTop, blueBorderBottom, ieFlexOverflowFix]}>
       <div css={[communicationCardInfoContainer]}>
         <img css={img} src={image} alt={title} />
         <div css={ieFlexOverflowFix}>
-          <h3 css={communicationCardHeadingText}>{title}</h3>
+          <HeadingTag css={communicationCardHeadingText}>{title}</HeadingTag>
```
coldlink, line 133: I had no idea you could do this.. amazing!

coldlink (approved):

src/client/components/CommunicationCard.tsx
liamness: I've used strings as tags before, just needed to figure out how to make tsc happy (the `as const` did the trick).

PR #1446, opened 2022-01-31
Add Sentry logging around token expiry flows

ob6160:
## What does this change?
We'd like to have enhanced visibility around when and how our token expiry pages are shown.

The reason for this is that we have had several reports of users being wrongly shown an expired token page after requesting a password reset through Gateway.

This PR adds Sentry logging for the following pages:
* Welcome page token expiry `setTimeout` redirect.
* Set password page token expiry `setTimeout` redirect.
* Reset password page token expiry `setTimeout` redirect.
* Expired token page rendering (including the token that was passed if present).

Having logging on the redirect will help us determine if the timeout is causing the issue. Logging for the expired token page will give us visibility on whether a malformed value or token is being passed from certain email clients, causing the page to be shown mistakenly.


ob6160:

ob6160:
> looking good
> 
> one thing that might also be worth adding is logging the "happy path", i.e when the normal change/set/welcome page is shown, to make it easier to compare numbers when researching this

thanks for the review–I've pushed some changes to log when the page loads successfully. also fixed a logic problem when logging out the token value on failure in `ResendPasswordPage.tsx`

PR #1448, opened 2022-01-31
Remove `http-errors` as a dependency

coldlink:
## What does this change?

The latest update to [`http-errors`](https://www.npmjs.com/package/http-errors) includes breaking changes which cause our application to fail to start, due to existing dependencies requiring older version of this package.

After investigation we saw that this was only being used in one place in our codebase outside of libraries/other dependencies, so rather than update the dependency, it's better to remove it if we can replicate it's functionality.

We can see that the main usage for `http-errors` was the inclusion of the `code` property. We can replicate this functionality by making a new class which extends our exiting error functionality which we use in other places, thus created `HttpError` class which extends `ApiError` with the additional `code` property. The rest of the properties are the same.

We can verify that this works as expected by running the mocked tests with the recaptcha working as expected.

I also noticed that this was error was being generated for every request, although it was constant in it's usage, as a small optimisation I moved this into the outer scope, so it's not being created on every request.


ob6160 (approved):
nice work

PR #1450, opened 2022-02-03
Refactor: Change password error messages

rebecca-thompson:
## What does this change?
This makes the change password error message more clear and removes duplicate messages and messages that should not be displayed to the end-user.

In addition, it changes `validatePasswordField.ts` to throw an error rather than populating a list or errors. This reduces the need for duplicate code in the `changePassword.ts` controller as these errors can be picked up by the try-catch block.


ob6160 (approved):
thanks for tidying this up, looks great

src/client/components/ConsentsHeader.tsx
```diff
@@ -14,7 +14,7 @@ type Props = {
 export const ConsentsHeader = ({ error, success, geolocation }: Props) => (
   <>
     <Header geolocation={geolocation} />
-    {error && <GlobalError error={error} link={getErrorLink(error)} left />}
+    {error && <GlobalError error={error} link={getErrorLink()} left />}
```
dominickendrick, line 17: Could `link` just be `DEFAULT_ERROR_LINK` and remove the `getErrorLink` function, as it only return a single value now ?

coldlink (approved):

PR #1451, opened 2022-02-03
Changes to client side logging with Sentry

ob6160:
## What does this change?
We recently added some extra client-side event tracking around the reset password, change password and welcome page flows in #1446

The logs introduced in that PR inserted some volatile variables in logged Sentry messages. This meant that it didn't group them correctly. Here we modify the logging so that these values are contained instead in the Sentry `extras` object as additional debug data.

Secondly we tweak the client side logger to ensure that each log message is wrapped in a [Sentry Transaction](https://docs.sentry.io/product/performance/transaction-summary/). This lets us override the sample rate of 20% and ensure that Sentry keeps track of all intentional logging that we add to the platform.

Finally we add a check for user consent before initialising the Sentry logger. If the user consent changes, the Sentry logger is initialised or destroyed as appropriate.

### DIFF
--- a/src/client/lib/clientSideLogger.ts
+++ b/src/client/lib/clientSideLogger.ts
@@ -1,6 +1,7 @@
 import { BaseLogger } from '@/shared/lib/baseLogger';
 import { LogLevel } from '@/shared/model/Logger';
 import * as Sentry from '@sentry/browser';
+import { Extras } from '@sentry/types';
 
 const getSentryLevel = (level: LogLevel) => {
   switch (level) {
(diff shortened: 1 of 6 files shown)


ob6160:
[Change Sentry configuration so that users have to give consent for it to be enabled. Improve logging so we log all info events](https://trello.com/c/lfgbK18t/3796-change-sentry-configuration-so-that-users-have-to-give-consent-for-it-to-be-enabled-improve-logging-so-we-log-all-info-events)

coldlink (approved):

PR #1462, opened 2022-02-07
Lazy load Sentry (WIP)

ob6160:
## Accessibility


ob6160:
[Change our Sentry implementation so that we only load the library when needed. Abstract into a common library for use across DCR and Gateway if feasible in the timeframe (S)](https://trello.com/c/lfgbK18t/3796-change-our-sentry-implementation-so-that-we-only-load-the-library-when-needed-abstract-into-a-common-library-for-use-across-dcr)

PR #1463, opened 2022-02-08
Revert check for Sentry consent

ob6160:
## What does this change?
It was confirmed that we do not need explicit consent for users to enable Sentry as it is listed as an "essential vendor". As a result we're removing the check in this PR

A wider set of improvements to dynamically import Sentry and add a logging queue for before it is in progress here: #1462


ob6160:
[Change our Sentry implementation so that we only load the library when needed. Abstract into a common library for use across DCR and Gateway if feasible in the timeframe (S)](https://trello.com/c/lfgbK18t/3796-change-our-sentry-implementation-so-that-we-only-load-the-library-when-needed-abstract-into-a-common-library-for-use-across-dcr)

coldlink (approved):

PR #1470, opened 2022-02-09
Okta POC | Running Sign In e2e tests against Okta 

ob6160:
## What does this change?
This is a proof of concept for how we can share code between our existing Cypress test suites and a separate suite specifically for testing the work in progress Okta integration.

The changes here cover the sign in e2e tests only so that we can prove out the concept. The key changes:

* The test logic for the Sign In tests was abstracted out into a shared file: `sign_in.shared.ts`.
* Both Sign In tests now share the test logic moved to `sign_in.shared.ts`
  * Cypress tests were organised into folders for when we introduce more shared code.
  * Each test is stored as an array with the test name and test method
  * When each test is initialised, a flag `isOkta` is captured in a closure and used to determine if the test will be run with the `useOkta=true` query param.
  * The key modification to each test, using this technique, was appending the `useOkta` query param to the visitUrl. Example: `const visitUrl = isOkta ? '/signin?useOkta=true' : '/signin';`
* A new Cypress suite for Okta tests was created, listed as `cypress-ete-okta`.
* A new makefile entry was added: `make cypress-ete-okta` so that we can test locally.
* Github action secrets were updated with the latest Okta CODE environment details
* A fix was added to (src/server/lib/getProfileUrl.ts) to ensure that on `localhost` we use `http` instead of `https`


ob6160:
[Research: Setup test users in Okta for end to end testing (1 week timebox)](https://trello.com/c/BHJVKF9z/3786-research-setup-test-users-in-okta-for-end-to-end-testing-1-week-timebox)

.github/workflows/cypress-ete-okta.yml
```diff
@@ -0,0 +1,61 @@
...
+        uses: actions/checkout@v2
+      - uses: actions/setup-node@v2
+        with:
+          node-version-file: '.nvmrc'
+          cache: 'yarn'
+      - name: Cypress run
+        uses: cypress-io/github-action@v2
+        env:
+          PORT: 8861
+          ENVIRONMENT: production
```
rebecca-thompson, line 23: Is this meant to say production? Does this variable even get used anymore?

rebecca-thompson (commented):
Just a couple of small comments around code duplication and potentially unused/misleading variables

cypress-ete-okta.sh
```diff
@@ -0,0 +1,27 @@
...
+  echo "This ID is required to run these ete Cypress tests. You can find the ID here: https://mailosaur.com/app/servers" 
+  echo
+else
+  CI_ENV=$(cat .env | tr '\n' ',')
+  CI_ENV=${CI_ENV%?}
+  yarn build
+  yarn start &
+  yarn wait-on:server
+  yarn cypress run --env $CI_ENV --config '{"testFiles":["ete-okta/**/*.ts"]}'
+fi
```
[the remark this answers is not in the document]
ob6160: good idea! I'll change to do this

.github/workflows/cypress-ete-okta.yml
ob6160: hm, I think the only place it's used is to set the `mode` in our webpack configuration. can probably be removed in a separate pr in favour of checking the `STAGE` variable  (as it'll need to be wiped from our .env files in aws etc)

rebecca-thompson (approved):

PR #1472, opened 2022-02-10
Okta | Change password

rebecca-thompson:
## What does this change?
This implements the change password flow in Okta behind the feature flag. There are three main sections to changing a password which are now handled by three separate controllers:
1.  Send reset password email -> `src/server/controllers/sendChangePasswordEmail.ts`
2. Check password token when landing on the password page -> `src/server/controllers/checkPasswordToken.ts`
3. Change password -> `src/server/controllers/changePassword.ts`

The reset password flow in Okta works as follows:
1. User enters an email in the forgotten password page (`/reset-password`)
2. Okta sends the email to reset the user's password
3. User clicks on the emailed link to reset their password (`/reset-password/:token`)
4. Gateway takes the token parameter (this is called a recovery token in Okta) and exchanges it for an Okta state token*
5. The user enters their new password, Gateway uses the state token to set the password in Okta. If successful, Okta returns a session token which can be used to log in
6. Gateway kicks off an OAuth code flow in Okta using the session token to log the user in

*The recovery token can be used to generate multiple state tokens up until the point at which it expires/is used. If there are problems setting the password, Gateway will re-verify the recovery token which generates a new state token.
There is a good overview on Okta's recovery, state and session tokens here: https://developer.okta.com/docs/reference/api/authn/#tokens

## Caveats and Limitations
1. When completing registration using Okta, the welcome flow will fail at the point at which it tries to set cookies in IDAPI. This is because we have yet to build the Registration hook to write a user into IDAPI after it has been created in Okta. There is a card in our backlog for this.
2. The set and reset password flows for users without a password that try to register will currently fail when using Okta. This is because we are planning to send these emails from gateway and this work is still pending (these are the 'Nearly there...' emails)
3. Signing the user into Okta after they have reset their password may fail if there is a pre-existing login session. This will be fixed when sign out is implemented in Okta. A TODO has been added to the code to sign the user out before kicking off the oauth flow.

## Additional Refactoring
This also includes some changes to the Okta error handling to make it easier to perform logic on the error responses.

## How to test
You can test this by deploying to `CODE` and using the `useOkta=true` parameter.

## Images
Untitled (13)


rebecca-thompson:
[Build reset password step for MVP3](https://trello.com/c/noZsXFh5/3722-build-reset-password-step-for-mvp3)

cypress/integration/mocked/okta_change_password.spec.ts
```diff
@@ -0,0 +1,322 @@
+/**
+ * Test Scenarios:
```
ob6160, line 2: I really like this way of summarising the tests, it helped me understand what was happening a lot faster

cypress/integration/mocked/okta_change_password.spec.ts
```diff
@@ -0,0 +1,322 @@
...
+ * Unexpected error changing the password           | show password page with global error: There was a problem changing your password, please try again.
+ * Password is too short                            | show password page with field error: Please make sure your password is at least 8 characters long.
+ * Password is too long                             | show password page with field error: Please make sure your password is not longer than 72 characters.
+ * Password cannot be your current password         | show reset password page with field error: Please use a password that is different to your current password.
+ * Password cannot be a common or breached password | show reset password page with field error: Please use a password that is hard to guess.
+ * Password set successfully                        | show password updated confirmation page
+ */
+
+describe('Change password in Okta', () => {
+  context.skip('reset password page', () => {
```
ob6160, line 21: just checking, is the skip there until we turn on Okta or was it left over from testing?

ob6160 (commented):
very nice work, ran it locally and it all looked great 👍  
a few general comments so far. I have a few things still to read through so will put any updates on here later if I spot anything.

cypress/integration/mocked/okta_change_password.spec.ts
ob6160: By skipping these tests if there's a breaking change in the intervening time until we enable Okta could we potentially miss it?

src/server/controllers/changePassword.ts
```diff
@@ -19,204 +19,218 @@ import {
...
-        redirect_uri: `https://${baseUri}/oauth/authorization-code/callback`,
-        sessionToken,
-      };
-      return res.redirect(
-        303,
-        `${okta.orgUrl}/oauth2/${okta.authServerId}/v1/authorize?${stringify(
-          params,
-        )}`,
-      );
+      // TODO: once sign out with Okta has been implemented, invalidate current sessions before kicking off code flow
```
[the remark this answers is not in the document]
rebecca-thompson: Moving over the ete tests for Okta should pick this up, as well as some other bits that are missing (passing in `ref` and `refviewid` is also missing as we can't pass them as url parameters when sending emails from Okta). I think this can be left to a future pr, so as not to bloat this one

ob6160 (approved):

PR #1476, opened 2022-02-11
Migrate sign out like-for-like from Identity-Frontend to Gateway

coldlink:
## What does this change?

Like for like migration of the `/signout` endpoint from [identity-frontend](https://github.com/guardian/identity-frontend) to gateway.

## Checklist
- [x] Set up IDAPI `/unauth` call
- [x] Clearing IDAPI cookies
- [x] Clearing dotcom cookies
  - Cookie options for `clearCookie`:
  - chrome_VIKkyP5fQP
  - There are seemingly lots of other user cookies when researching this https://github.com/guardian/frontend/blob/main/static/src/javascripts/projects/common/modules/commercial/user-features.ts#L45 but we've only migrated the existing dotcom cookies as this is a like-for-like
- [x] Clearing gateway cookies
- [x] VCL routing in [identity-platform](https://github.com/guardian/identity-platform)
  - https://github.com/guardian/identity-platform/pull/497
- [x] Metrics
- [x] Mocked cypress tests
- [x] E2E cypress tests

## Testing plan

https://docs.google.com/document/d/1O5k1PynYpDdOXXgvwZyPpvT51GNOAAXYFZ66jMPwtmM/edit

## Other Changes
- Use `303` in login middleware, so that the redirect is always a `GET`
- Use separate environment variable file for ete cypress tests
- Allow `returnUrl` parameter to read from both query parameters and the `Referrer` header


cypress/integration/mocked/welcome.spec.ts
```diff
@@ -63,7 +63,6 @@ describe('Welcome and set password page', () => {
     );
     cy.mockAll(200, allConsents, CONSENTS_ENDPOINT);
     cy.mockAll(200, verifiedUserWithNoConsent, USER_ENDPOINT);
-    setAuthCookies();
```
rebecca-thompson, line 66: Previously the backend was not setting cookies on the correct domain for localhost, so these were set again manually in the tests. Now they rely on the cookies set in the response headers

docs/setup.md
```diff
@@ -19,7 +19,7 @@ Make sure you have one or the other or both:
 
 ### Environment File
 
-Populate a `.env` file by using the examples from [`.env.example`](../.env.example). The `.env` file should **never** be committed.
+Populate `.env` and `cypress-ete.env` files by using the examples from [`.env.example`](../.env.example) and [`cypress-ete.env.example`](../cypress-ete.env.example.example). These `.env` files should **never** be committed as they store secrets.
```
rebecca-thompson, line 22: the `.env` file was being used for both local development and running ete tests locally (we have a separate env file for mocked tests). Extracting the ete tests into their own env file makes it easier to test for the presence or absence of ccokies set on the server response.

coldlink: Is it worth uploading a `cypress-ete.env` file to s3 too, and a link in the documentation to download it?

dominickendrick (approved):

rebecca-thompson (approved):

PR #1489, opened 2022-02-18
Add ToggleSwitchInput

vlbee:
## What does this change?

Adds a new Toggle Switch Input component that can be used for uncontrolled React form submissions

We had to create new switch component that functioned as a form input, not a button. 

This component is an adaptation of the Guardian Source Toggle but simplified for our use case. Stylings have been lifted and simplified. 

## Images

image

Intended usage: 
image

## Accessibility

- [x] [Tested with screen reader](https://guardian.github.io/source/?path=/docs/docs-06-accessibility--page#screen-readers)
- [x] [Navigable with keyboard](https://guardian.github.io/source/?path=/docs/docs-06-accessibility--page#keyboard-navigation)
- [x] [The change doesn't use only colour to convey meaning](https://guardian.github.io/source/?path=/docs/docs-06-accessibility--page#use-of-colour)

### Cross browser and device testing

-  [x] Tested with touch screen device
- [x] Chrome 77+ 
- [x] Firefox 68+ 
- [x] Edge 17+ 
- [x] Safari 12+ 
- [x] IE11

### Responsiveness

-   [x] Tested at all breakpoints
-   [x] Tested with with long text
-   [x] Stretched to fill a wide container

### Documentation

-   [ ] Full API surface area is documented in the README
-   [x] Examples in Storybook

### DIFF
--- a/src/client/components/ToggleSwitchInput.tsx
+++ b/src/client/components/ToggleSwitchInput.tsx
@@ -1,0 +1,154 @@
+import React from 'react';
+import type { EmotionJSX } from '@emotion/react/types/jsx-namespace';
+import type { Props } from '@guardian/source-react-components';
+import { css } from '@emotion/react';
+import {
+  neutral,
+  success,
+  textSans,
+  focusHalo,
+  visuallyHidden,
+  descriptionId,
+  generateSourceId,
+} from '@guardian/source-foundations';
+
+const inputStyles = css`
+  ${visuallyHidden}
+`;
+
+const labelStyles = css`
+  user-select: none;
+  position: relative;
+  ${textSans.small()};
+  display: flex;
+  color: ${neutral[7]};
+  align-items: center;
+  cursor: pointer;
+`;
+
+const siblingStyles = css`
+  input + span {
+    background-color: rgba(153, 153, 153, 0.5);
+  }
+
+  input:focus + span {
+    ${focusHalo};
+  }
+
+  input:checked + span {
+    background: ${success[500]};
+  }
+
+  input:checked + span:before {
+    opacity: 1;
+    z-index: 1;
+  }
+
+  input:checked + span:after {
+    left: 1.375rem;
+    background: ${neutral[100]};
+  }
+`;
+
+const switchStyles = css`
+  flex: none;
+  border: none;
+  margin: 0px 0px 0px 8px;
+  padding: 0;
+  display: inline-block;
+  text-align: center;
+  position: relative;
+  transition: background-color 150ms cubic-bezier(0.4, 0, 0.2, 1) 0ms;
+  cursor: pointer;
+  width: 2.75rem;
+  height: 1.5rem;
+  border-radius: 15.5px;
+
+  &:before,
+  &:after {
+    box-sizing: border-box;
+  }
+
+  &:before {
+    content: '';
+    position: absolute;
+    top: 6px;
+    height: 11px;
+    width: 6px;
+    right: 10px;
+    opacity: 0;
+    border-bottom: 2px solid ${success[400]};
+    border-right: 2px solid ${success[400]};
+    transform: rotate(45deg);
+    transition: opacity 0.1s ease-in;
+  }
+
+  &:after {
+    content: '';
+    position: absolute;
+    border-radius: 50%;
+    background: #fff;
+    will-change: left;
+    transition: left 0.15s ease-in-out;
+    height: 18px;
+    width: 18px;
+    top: 3px;
+    left: 4px;
+  }
+`;
+
+export interface ToggleSwitchInputProps extends Props {
+  /**
+   * Whether the ToggleSwitch is checked. This is necessary when using the
+   * [controlled approach](https://reactjs.org/docs/forms.html#controlled-components)
+   * (recommended) to form state management.
+   *
+   * Note: if you pass the `checked` prop, you MUST also pass an `onChange`
+   * handler, or the field will be rendered as read-only.
+   */
+  checked?: boolean;
+  /**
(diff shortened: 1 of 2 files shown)


src/client/components/ToggleSwitchInput.tsx
```diff
@@ -0,0 +1,213 @@
...
+    <label id={labelId} css={[labelStyles, siblingStyles, cssOverrides]}>
+      {labelPosition === 'left' && label}
+      <input
+        css={[inputStyles]}
+        name={switchName}
+        type="checkbox"
+        role="switch"
+        defaultChecked={defaultChecked != null ? defaultChecked : undefined}
+        checked={checked != null ? isChecked() : undefined}
+        aria-checked={isChecked()}
```
ob6160, line 201: As it is, this won't update the `aria-checked` parameter when it's uncontrolled. I think we could get around this by setting some state called `uncontrolledChecked` or similar and handling the `onChange` event on the input.

Something like:

```typescript
const [uncontrolledChecked, setUncontrolledChecked] = React.useState(
  !!defaultChecked,
);

<input
onChange={(e) => setUncontrolledChecked(e.target.checked)}
aria-checked={checked != undefined ? checked : uncontrolledChecked}
// .. everything else
></input>
```

perhaps

src/client/components/ToggleSwitchInput.tsx
```diff
@@ -0,0 +1,213 @@
...
+    }
+
+    return !!defaultChecked;
+  };
+
+  return (
+    <label id={labelId} css={[labelStyles, siblingStyles, cssOverrides]}>
+      {labelPosition === 'left' && label}
+      <input
+        css={[inputStyles]}
```
ob6160, line 195:
```suggestion
        css={inputStyles}
```

src/client/components/ToggleSwitchInput.tsx
```diff
@@ -0,0 +1,213 @@
...
+        name={switchName}
+        type="checkbox"
+        role="switch"
+        defaultChecked={defaultChecked != null ? defaultChecked : undefined}
+        checked={checked != null ? isChecked() : undefined}
```
ob6160, line 200: I think we can get rid of the `isChecked()` method entirely and replace this line with:

`checked != undefined ? checked : defaultChecked` because `defaultChecked` can either be true/false/undefined

src/client/components/ToggleSwitchInput.tsx
```diff
@@ -0,0 +1,213 @@
...
+  labelPosition = 'right',
+  checked,
+  defaultChecked,
+  platform = 'web',
+  cssOverrides,
+}: ToggleSwitchInputProps): EmotionJSX.Element => {
+  const switchName = id ?? generateSourceId();
+  const labelId = descriptionId(switchName);
+
+  const isChecked = (): boolean => {
```
ob6160, line 183: I think this adds a lot of unnecessary obfuscation, I believe it can be removed by making the change I suggested below

ob6160 (commented):
really nice work! from a functional standpoint it seems great.

I realise this is based on the version in the kitchen, but I have some suggestions to improve the implementation that we could maybe apply to both.

One request, would it be possible to write a few tests for this? We have a unit testing library called `jest-dom` which you can use to render the component and make assertions on stuff like `aria-checked` to make sure that it behaves as advertised. Happy to pair on this!

src/client/components/ToggleSwitchInput.tsx
vlbee: `aria-checked` seems to be redundant on an actual input type=checkbox so think we can do away with this.

SiAdcock:
> We had to create new switch component that functioned as a form input, not a button.

Before we fork this component, I think it would be useful to understand why the button implementation is not working for you. Is it because you had trouble getting the button with a switch role to work in an uncontrolled way?

src/client/components/ToggleSwitchInput.tsx
```diff
@@ -0,0 +1,204 @@
...
+  }
+`;
+
+const siblingStyles = css`
+  input + span {
+    background-color: rgba(153, 153, 153, 0.5);
+  }
+
+  /* TODO focus-visible not fully supported */
+  input:focus-visible + span {
```
vlbee, line 60: Struggling to fine a better solution to this (showing outline on tab focus but not mouse hover/click)

vlbee:
Tests will be added in subsequent PR when this component is used in the Consents Data page

src/client/components/ToggleSwitchInput.tsx
```diff
@@ -0,0 +1,199 @@
...
+  /* TODO does not work in chrome/edge/firefox if :focus is used for Safari*/
+  input:focus:not(:focus-visible) + span {
+    outline: none !important;
+  }
+
+  /* Focus-visible not fully supported in Safari <15.4 and IE 11 */
+  input:focus-visible + span {
+    transition: ${transitions.medium};
+    ${focusHalo};
+  }
```
vlbee, line 74: Couldn't find a solution that worked across all browsers so replicating the focus style behaviour of the check box which shows focus on mouse events still.

vlbee: Any better solutions let me know pls.

SiAdcock: Take a look at [https://guardian.github.io/source/?path=/story/packages-source-foundations-focusstylemanager--page](`FocusStyleManager`) which [Gateway is already using](https://github.com/guardian/gateway/blob/6653d8c3faa31a3ae743c74d7f337258a7c64c6a/src/client/static/sourceAccessibility.tsx)

vlbee: Thanks!

coldlink (approved):

PR #1492, opened 2022-02-19
Rate Limiting (pt. 1): Implement redis backed token bucket rate limiter

ob6160:
## Description
We want to rate limit requests that are made to Gateway in a safe and performant manner.

We have an existing implementation of rate limiting in IDAPI. This is based on a Redis backed token bucket strategy [^1]. It been tried and tested in production and has been proven to be stable and effective.

[^1]: https://en.wikipedia.org/wiki/Token_bucket

This pull request ports the rate limiting mechanism implemented for IDAPI over to Gateway so that we can benefit from it as well.

It should be noted that we want to keep the configuration for the rate limiter private once put into use in production. A subsequent pull request will introduce private configuration and appropriate middleware to be applied to our routes.

## Method

A Token "bucket" of fixed capacity is defined and given an initial amount of tokens. The bucket is stored in Redis with a maximum timeout before the record removed.

Each call to the rate limiter with the same value consumes a token until all are depleted. At this point a delay kicks in, limiting the rate at which new tokens are added back to the bucket and hence the rate of requests that can be made by the user.

The number of new tokens to add back to the bucket can be calculated upon each request by working out the difference between the time of the request and when the bucket was last written to. Knowing this time delta and the rate at which new tokens are added, we can derive the amount of tokens that will have been added to the bucket since the last request.

### Buckets

Individual bucket lookup keys in Redis follow the scheme:

`rl-{rateLimiterName}-{bucketName}-{hashedValue}`

We use these keys to index named buckets for unique values in Redis. For example, this record: `rl-signin-email-hashedEmail` might represent a bucket that limits based on a unique email under the sign-in route in Gateway.

Each bucket includes the following information encoded as JSON:

* Capacity – the maximum burst capacity for the bucket.
* addTokenMs – the time delay in milliseconds before a new token is added to the bucket.

In tandem, we make a call to the Redis `pttl` method to determine the remaining time to live of the bucket. This is then  used to work out how many new tokens to add to the bucket since the last request was made (see the method section for more detail).

If a bucket has not been created in Redis, we instantiate a new bucket and set an expiry (time to live). Each time the bucket is updated this expiry is reset so that we always know how long it has been since the last request using `pttl`.

Rate limit buckets are given a precedence. If a lower bucket rate limit is hit, the buckets higher in the precedence are not updated.

Pipelining the bucket read/write operations to Redis keeps the whole operation to just two calls to the database to read the bucket values and write back the new token values.

### Order of precedence

Each rate limiter has a `global` request bucket that rate limits based on every request and no extra parameter.  If the rate limit of this bucket is reached, all requests to that route will be denied. All of the other buckets are optional, leaving flexibility to the user in terms of which values they would like to limit against.

The Global bucket has the highest precedence; so if a lower bucket is rate limited, it won't be updated. This ensures that individual users cannot carry out a denial of service attack as they are stopped by a more specific bucket.

### Usage example

```typescript
  await rateLimit({
    name: '/signin',
    redisClient,
    bucketConfiguration: {
      ipBucket: { addTokenMs: 500, capacity: 2 },
      globalBucket: { addTokenMs: 500, capacity: 5 },
    },
    bucketValues: {
      ip: '127.0.0.1',
    }
  })

/**
If the example above is executed with the same ip two
times within 500ms the ipBucket will be empty and the
method will return false. User is not rate limited.

Because the two requests above were made successfully,
the global rate limit bucket now has three tokens left.

If a further request is made before 500ms has passed,
the rate limiter will kick in and return 'ip', indicating
that the ip bucket has been limited.

Because the ip has been rate limited against, the global
rate limit token bucket has no further tokens removed 
from it until the ip bucket has more tokens available to
allow a request through.
*/
```

## Testing
An integration test suite using an in-memory Redis instance covers the core of the `rateLimit` method. The next pull request which puts the rate limiter into use on our express routes will include mocked and end-to-end tests to ensure that the custom configuration and middleware is working as expected.


ob6160:
[Develop unit and integration tests for rate limiting (S)](https://trello.com/c/mJ7MrqtJ/3814-develop-unit-and-integration-tests-for-rate-limiting-s)

ob6160:
[Build out global rate limit strategy before individual token buckets (M)](https://trello.com/c/6MVmQWtD/3812-build-out-global-rate-limit-strategy-before-individual-token-buckets-m)

ob6160:
[Implement token bucket rate limiting strategy in Gateway based on approach used in IDAPI (S)](https://trello.com/c/xCY7mrfT/3813-implement-token-bucket-rate-limiting-strategy-in-gateway-based-on-approach-used-in-idapi-s)

package.json
```diff
@@ -100,6 +102,7 @@
     "fork-ts-checker-webpack-plugin": "^7.2.1",
     "husky": "^7.0.4",
     "image-webpack-loader": "^8.1.0",
+    "ioredis-mock": "^7.1.0",
```
dominickendrick, line 105: is it ok that the types are on a different version number ? `"@types/ioredis-mock": "^5.6.0"` vs `"ioredis-mock": "^7.1.0"`

src/server/lib/rate-limit/bucket.ts
```diff
@@ -0,0 +1,170 @@
+import type Redis from 'ioredis';
+
+import type {
+  ParsedRateLimitBucket,
+  RateLimitBucketContents,
+  BucketKeys,
+  BucketConfiguration,
+  PipelinedBucketData,
+} from './types';
+
```
dominickendrick, line 10: Could we put some of the detail from the PR description about token bucket as comments here ? Maybe a link to the wikipedia article etc

dominickendrick (approved):
Great PR description and implementation, I wonder if we want to setup any logging / alarms ? Could be for the next PR tho.

package.json
ob6160: both are the latest versions respectively, and it looks like the typings in Definitely Typed for `ioredis-mock` were introduced by somebody outside of the `ioredis-mock` project, so I think that they're just  unrelated

ob6160:
> Great PR description and implementation, I wonder if we want to setup any logging / alarms ? Could be for the next PR tho.

thanks for taking a look. I had some logging in originally but decided to move that and adding metrics to the next PR as working on that will probably give me a better idea of what kind of metrics/logging we'll need

PR #1503, opened 2022-02-21
Upgrade to latest dependency versions (21/02/22)

ob6160:
## What does this change?
Upgrades our dependencies to the latest versions. Keeps `@storybook/*` packages pinned at alpha 34 until the full `6.5.0` release

- [ ] Tested in CODE
- [x] Tested in IE11


package.json
```diff
@@ -46,14 +46,14 @@
...
-    "@storybook/addon-links": "^6.5.0-alpha.34",
-    "@storybook/builder-webpack5": "^6.5.0-alpha.34",
-    "@storybook/manager-webpack5": "^6.5.0-alpha.34",
-    "@storybook/react": "^6.5.0-alpha.34",
+    "@storybook/addon-actions": "^6.5.0-alpha.40",
+    "@storybook/addon-essentials": "^6.5.0-alpha.40",
+    "@storybook/addon-links": "^6.5.0-alpha.40",
+    "@storybook/builder-webpack5": "^6.5.0-alpha.40",
+    "@storybook/manager-webpack5": "^6.5.0-alpha.40",
+    "@storybook/react": "^6.5.0-alpha.40",
```
liamness, line 54: Do we think it's a good idea to bump prerelease versions unless there's a specific change we need? I would suggest we leave these alone until the next proper semver release.

ob6160: agreed, will revert

PR #1506, opened 2022-02-22
Add disableOnSubmit prop to MainForm

ob6160:
## What does this change?
We believe that some users are submitting the change password form multiple times, causing unwanted behaviour.

This PR introduces a `disableOnSubmit` prop to `MainForm`. When this is set to `true`, the submit button is disabled upon form submit to prevent users from clicking the button more than once.

For now, this is only enabled for the password entry form.

If the reCAPTCHA check fails, the button is re-enabled; re-enabling the button lets the user have another attempt to resubmit the form.

A PR is open in Source to introduce a loading spinner so that users have a visual indication that something is happening [^1] Until this has been merged we set `cursor: not-allowed` and apply `aria-disabled` so that the user has a cue that something is happening as an action when they submit the form.

[^1]: https://github.com/guardian/source/pull/1275

Our unit tests for MainForm are modified to check that this new behaviour works as expected. The mocked and e2e tests for the change password page are also changed so we have coverage as this is the only place where the new behaviour is enabled.

## Behaviour demonstration
https://user-images.githubusercontent.com/1771189/155137420-34e964d1-d4ae-408f-91f4-770739d1bae4.mov

- [x] Tested on CODE
- [x] Tested on IE11
- [x] Tested on Firefox
- [x] Tested on Chrome
- [x] Tested on Safari
- [x] Tested on Samsung Internet


ob6160:

cypress/integration/mocked/change_password.spec.ts
```diff
@@ -193,14 +194,14 @@ describe('Password change flow', () => {
     });
 
     it('shows password change success screen, with default redirect button, and users email', () => {
-      cy.mockNext(200, fakeValidationRespone());
+      cy.mockNext(200, fakeValidationResponse());
```
ob6160, line 197: renamed this due to typo

src/client/__tests__/MainForm.test.tsx
```diff
@@ -76,7 +76,7 @@ test('shows sets an error message when the reCAPTCHA check is unsuccessful', asy
 test('calls the form submit override method if defined', async () => {
   const mockedSubmitOverride = jest.fn();
   const { findByText } = setup({
-    onSubmitOverride: mockedSubmitOverride,
+    onSubmit: mockedSubmitOverride,
```
ob6160, line 79: renamed this to more accurately describe what it does

src/client/components/MainForm.tsx
```diff
@@ -66,8 +67,9 @@ export const MainForm = ({
   setRecaptchaErrorMessage,
   setRecaptchaErrorContext,
   hasGuardianTerms = false,
-  onSubmitOverride,
+  onSubmit,
```
ob6160, line 70: renamed to more accurately describe what it does

cypress/integration/mocked/change_password.spec.ts
```diff
@@ -159,14 +159,14 @@ describe('Password change flow', () => {
...
       cy.visit(`/reset-password/fake_token`);
-      cy.contains(fakeValidationRespone().user.primaryEmailAddress);
+      cy.contains(fakeValidationResponse().user.primaryEmailAddress);
     });
   });
 
   context('Valid password entered', () => {
-    it('shows password change success screen, with a default redirect button.', () => {
+    it.only('shows password change success screen, with a default redirect button.', () => {
```
liamness, line 169: I imagine this was just for testing?

liamness (commented):
Just a question about this test usage, looks good more generally

cypress/integration/mocked/change_password.spec.ts
ob6160: ah nice one, I had in a commit but forgot to push :)

src/client/components/MainForm.tsx
coldlink: I just thought that `onSubmit` suggests that it should be the default behaviour, rather than the no javascript behaviour of the HTML form.

ob6160: I would argue that because it's an extension on top of some base JS functionality we have integrated in MainForm it makes sense to be `onSubmit` semantically as it is a callback that happens in addition to the submit behaviour of MainForm. 

With override added to it, my first thought is that it instead replaces the custom behaviour that MainForm applies and is passed directly to the underlying `form`

coldlink (approved):
approving with a small comment

src/client/layouts/Main.tsx
```diff
@@ -107,6 +107,13 @@ export const buttonStyles = ({ hasTerms = false, halfWidth = false }) => css`
   justify-content: center;
   width: 100%;
 
+  // Set the cursor when disabled temporarily so we have a visual indication
+  // until we have a loading spinner.
+  // PR for loading spinner in Source: https://github.com/guardian/source/pull/1275
```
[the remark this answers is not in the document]
ob6160: good spot, I'll change it to stay true to CSS standards

ob6160:
> approving with a small comment

thanks you! one change I've added since that needs a look if you get a moment:

In cases where we display the breach password error (or other errors resulting from the higher `onSubmit` handler) the button didn't leave the `disabled` state to let the user retry their submission.

I've changed it now to return whether an error was displayed back down to the `MainForm` so that we can re-enable the button for the user to try again once they've corrected their breached/incorrect password.

Here is the core logic: https://github.com/guardian/gateway/pull/1506/files#diff-6b2012d0cb55fc2f05610053970775217e250df05c01989f885eaab985f8c34cR99

coldlink (approved):

PR #1508, opened 2022-02-23
Standardise image sizes and remove unused images

coldlink:
## What does this change?

We noticed that there was a huge image being included in the production bundle.

```
1365668ad2a80e87d858149a2ced090d.jpg (681 KiB)
```

Which is the `morning-briefing-au.jpg`, the image resolution was `4096x2458` and the raw size was over 9MB, being optimised to 681KB by webpack.

Other images were generally standardised to `800x480` so were much smaller.

We also discussed possible future solutions around images, but decided to save these for a future discussion.

For now we've standardised all images to `800x480`, as well as removing images that we're not using.


ob6160 (approved):
interesting how we've got a mixture of pngs and jpgs too, is that fine or do we want to standardise on one maybe

liamness (approved):

coldlink:
> interesting how we've got a mixture of pngs and jpgs too, is that fine or do we want to standardise on one maybe

good point! one for the future discussion too

PR #1509, opened 2022-02-23
Add Focus Manager Decorator global to Storybook

vlbee:
## What does this change?
Adds the Source Focus Style Manager to the global Storybook environment so that developing components in Storybook accurately reflect how their styling behaves in production. 

## Accessibility

### DIFF
--- a/.storybook/preview.js
+++ b/.storybook/preview.js
@@ -3,10 +3,13 @@ import { Breakpoints } from '@/client/models/Style';
 import React from 'react';
 import { FocusStyleManager } from '@guardian/source-foundations';
 
+/* Source provides a global utility that manages the appearance of focus styles. When enabled,
+ * focus styles will be hidden while the user interacts using the mouse.
+ * They will appear when the tab key is pressed to begin keyboard navigation. */
 export const FocusManagerDecorator = (storyFn) => {
   React.useEffect(() => {
     FocusStyleManager.onlyShowFocusOnTabs();
-  });
+  }, []);
 
   return <>{storyFn()}</>;
 };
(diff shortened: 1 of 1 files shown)


.storybook/preview.js
```diff
@@ -1,5 +1,15 @@
 import { INITIAL_VIEWPORTS } from '@storybook/addon-viewport';
 import { Breakpoints } from '@/client/models/Style';
+import React from 'react';
+import { FocusStyleManager } from '@guardian/source-foundations';
+
+export const FocusManagerDecorator = (storyFn) => {
+  React.useEffect(() => {
+    FocusStyleManager.onlyShowFocusOnTabs();
+  });
```
ob6160, line 9: as this doesn't have a dependency array, this will apply on every re-render, do we want that to happen, or do we just want it to run once/when the dependencies change?

.storybook/preview.js
```diff
@@ -1,5 +1,15 @@
 import { INITIAL_VIEWPORTS } from '@storybook/addon-viewport';
 import { Breakpoints } from '@/client/models/Style';
+import React from 'react';
+import { FocusStyleManager } from '@guardian/source-foundations';
+
+export const FocusManagerDecorator = (storyFn) => {
```
ob6160, line 6: might be worth adding the same explanatory comment that we have in `sourceAccessibility.tsx`

.storybook/preview.js
coldlink:
```js
  React.useEffect(() => {
    FocusStyleManager.onlyShowFocusOnTabs();
  }, []);
```

surely this would stop the re-render issue then?

coldlink (approved):

PR #1510, opened 2022-02-24
Implemented onboarding flow progress indicator as per design

liamness:
## What does this change?

Fixes appearance. ~~Draft for now, checking I haven't broken anything in Chromatic.~~

## How to test

Testing via Storybook / Chromatic should be sufficient.

✨

### Mobile

image

### Tablet

image

### Desktop

image

## Accessibility

### DIFF
--- a/src/client/components/ConsentsSubHeader.tsx
+++ b/src/client/components/ConsentsSubHeader.tsx
@@ -165,6 +165,7 @@ const li = (index: number, status: PageStatus) => {
 
 const progressWrapper = css`
   padding: 0 ${CIRCLE_RADIUS}px;
+  overflow: hidden;
 `;
 
 const hiddenText = css`
--- a/src/client/components/ConsentsSubHeader.tsx
+++ b/src/client/components/ConsentsSubHeader.tsx
@@ -61,15 +61,15 @@ const ol = (active: number) => {
   return css`
     list-style: none;
     height: ${CIRCLE_DIAMETER}px;
-    margin: 14px 0 44px;
+    margin: 14px 0 40px;
     position: relative;
 
     ${from.tablet} {
-      margin: 18px 0 32px;
+      margin: 18px 0 28px;
     }
 
     ${from.desktop} {
-      margin: 26px 0 26px;
+      margin: 26px 0 22px;
     }
 
     &:before,
@@ -141,6 +141,8 @@ const li = (index: number, status: PageStatus) => {
       height: ${CIRCLE_DIAMETER}px;
       position: absolute;
       left: -${CIRCLE_RADIUS}px;
+      /* TODO: apply this to psuedo elements globally */
+      box-sizing: border-box;
       ${status === 'active' &&
       `
         background: ${neutral[100]};
--- a/src/client/components/ConsentsSubHeader.tsx
+++ b/src/client/components/ConsentsSubHeader.tsx
@@ -5,6 +5,7 @@ import {
   neutral,
   space,
   from,
+  until,
   textSans,
   headline,
 } from '@guardian/source-foundations';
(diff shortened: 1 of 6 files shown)


src/client/components/ConsentsSubHeader.tsx
```diff
@@ -40,224 +48,177 @@ const h1ResponsiveText = css`
...
+export const ConsentsSubHeader = ({ autoRow, title, current }: Props) => {
+  const active = current
+    ? (CONSENTS_PAGES_ARR as string[]).indexOf(current)
+    : 0;
+
+  const pageProgression = current && (
+    <div css={[progressWrapper, autoRow()]}>
+      <ol css={ol(active)}>
+        {CONSENTS_PAGES_ARR.map((page, index) => {
+          const status: PageStatus = (() => {
```
ob6160, line 188: thoughts on splitting out this unnamed function into a named one?

src/client/components/ConsentsSubHeader.tsx
```diff
@@ -40,224 +48,177 @@ const h1ResponsiveText = css`
...
+            if (index === active) {
+              return 'active';
+            }
+            if (index < active) {
+              return 'complete';
+            }
+            return 'pending';
+          })();
+          const screenReaderText = {
```
ob6160, line 197: is there a case for adding `pending` to this list?

src/client/components/ConsentsSubHeader.tsx
```diff
@@ -40,224 +48,177 @@ const h1ResponsiveText = css`
...
+            active: 'Current: ',
+            complete: 'Completed: ',
+          }[status as string];
+          return (
+            <li key={page} css={li(index, status)}>
+              <span>
+                {screenReaderText && (
+                  <span css={hiddenText}>{screenReaderText}</span>
```
ob6160, line 205: great work with the screen reader consideration!

ob6160 (approved):
this is great work, it's so hard to avoid a "hacky" solution to implementing designs like this within time constraints and I think this is entirely fit for purpose, so I've avoided commenting on the bulk of the styling implementation–I'm happy if it renders alright :) great work with the accessibility affordance too! it would be a great extra to have a demo in the description see/hear how it'd appear to a user who uses a screen reader

src/client/components/ConsentsSubHeader.tsx
```diff
@@ -40,224 +48,177 @@ const h1ResponsiveText = css`
...
+
+const progressWrapper = css`
+  padding: 0 ${CIRCLE_RADIUS}px;
+  overflow: hidden;
 `;
 
-const maxContentOverride = (isLastPage: boolean) => css`
-  width: ${isLastPage ? 'max-content' : 'initial'};
+const hiddenText = css`
+  display: none;
```
ob6160, line 176: just leaving as a note here – as discussed, this should probably be `visibility: hidden;` / using the Source visually hidden rules

src/client/components/ConsentsSubHeader.tsx
liamness: I was working from this example: https://www.w3.org/WAI/tutorials/forms/multi-page/

I think it should be implied from the order of the list items and the status of the preceding items?

src/client/components/ConsentsSubHeader.tsx
liamness: Have adjusted this now.

src/client/components/ConsentsSubHeader.tsx
liamness:
```diff
diff --git a/src/client/components/ConsentsSubHeader.tsx b/src/client/components/ConsentsSubHeader.tsx
index 9f882d95..8909b6be 100644
--- a/src/client/components/ConsentsSubHeader.tsx
+++ b/src/client/components/ConsentsSubHeader.tsx
@@ -177,6 +177,16 @@ const screenReaderTextStyles = css`
   ${visuallyHidden}
 `;
 
+const getPageStatus = (index: number, active: number): PageStatus => {
+  if (index === active) {
+    return 'active';
+  }
+  if (index < active) {
+    return 'complete';
+  }
+  return 'pending';
+};
+
 export const ConsentsSubHeader = ({ autoRow, title, current }: Props) => {
   const active = current
     ? (CONSENTS_PAGES_ARR as string[]).indexOf(current)
@@ -186,15 +196,7 @@ export const ConsentsSubHeader = ({ autoRow, title, current }: Props) => {
     <div css={[progressWrapper, autoRow()]}>
       <ol css={ol(active)}>
         {CONSENTS_PAGES_ARR.map((page, index) => {
-          const status: PageStatus = (() => {
-            if (index === active) {
-              return 'active';
-            }
-            if (index < active) {
-              return 'complete';
-            }
-            return 'pending';
-          })();
+          const status = getPageStatus(index, active);
           const screenReaderText = {
             active: 'Current: ',
             complete: 'Completed: ',
```

Something like this? I like IIFEs sometimes when I need to assign the result of something slightly complex, that can't just be a single line expression, but also isn't really used anywhere else so assigning a name to the function seems a bit redundant.

Nested ternaries could be a solution for this sort of thing, if Prettier formatted them in a way where they could be digested easily at a glance. But sadly that is not the case.

ob6160: yeah it's a tricky one as there are good arguments for both sides, having it as an iife helps when I'm reading code sometimes so I'm not constantly scrolling around. i'm cool with keeping it like it is here

src/client/components/ConsentsSubHeader.tsx
ob6160: That makes sense, that's a really helpful doc, thanks for linking

liamness:
[Onboarding flow redesign follow up](https://trello.com/c/ncPxVzLu/3810-onboarding-flow-redesign-follow-up)

PR #1511, opened 2022-02-24
Add invertOptOutConsent util 

vlbee:
## What does this change?

Adds a utility function to invert in both directions the boolean `consented` value of IDAPI opt out consents, eg. `profiling_optout`

First step to implement **option B** outlined below

This utility will be used [here](https://github.com/guardian/gateway/blob/main/src/server/lib/idapi/consents.ts) on [`update`](https://github.com/guardian/gateway/blob/964793dff8d47493ea9c1d06102ad54b8bbed4da/src/server/lib/idapi/consents.ts#L51) function and [`getUserConsentsForPage`](https://github.com/guardian/gateway/blob/964793dff8d47493ea9c1d06102ad54b8bbed4da/src/server/lib/idapi/consents.ts#L72) function

## Why?
The four legitimate interest opt out consents on the data model are being re-worded into opt ins from the user's perspective.
See [IDAPI PR 2079](https://github.com/guardian/identity/pull/2079)

This means we need to invert the boolean value of the consent on the client. There are two options where to do this: 

**Option A:** As close to the user as possible, when the value is displayed (eg, when the toggle switch is rendered)

* Tricky because Gateway uses uncontrolled form handling, so where data is passed into the component and where it is posted back is spread across different files. Modifying the data at the point of display is simple, but modifying it back before form submission would require clientside JS and potentially incorrectly capture user consents if JS is not enabled. CSS rendering of the toggle switch is also based on the form's input value so while it could be inverted here, it would only work for sighted users and be super hacky. 

**Option B:** As close to the API as possible, when data is fetched/posted to IDAPI

* Allows us to keep the inversion located in a single file so easier to document and test, but requires remodeling the Gateway consent as an optin. 

**EXISTING:**
| Consent ID | Value | User sees  | Client Input Value |
| ----------- | ----------- | ----------- | ----------- |
| `profiling_optout`   | `false`     | `I do NOT want`  | `false` - default |
| `profiling_optout`  | `true`       | `I do NOT want`      |`true`     |

**DESIRED:**
| Consent ID | Value | User sees  | Client Input Value |
| ----------- | ----------- | ----------- | ----------- |
| `profiling_optout`   | `false`     | `Allow`  | `true` - default |
| `profiling_optout`  | `true`       | `Allow`      | `false`     |

**PROPOSED SOLUTION:**
| Consent ID | IDAPI Value | Gateway ID | Server Value | User sees  | Client Input Value |
| ----------- | ----------- | ----------- | ----------- | ----------- | ----------- |
| `profiling_optout`   | `false`  | `profiling_optin`   | `true`     | `Allow`  | `true` - default |
| `profiling_optout`   | `true`  | `profiling_optin`  | `false`       | `Allow`      | `false`     |

## How to test

- [x] Unit test 

## Accessibility


src/server/lib/idapi/invertOptOutConsents.ts
```diff
@@ -0,0 +1,39 @@
...
+        } else {
+          return invertedConsent;
+        }
+      } else {
+        return consent;
+      }
+    });
+
+    return invertedConsents;
+  };
```
liamness, line 36: Use of the early return pattern could remove a significant amount of nesting here.

liamness: Here's what I mean (not tested):

```js
const invertConsents =
  (input: ConversionString, output: ConversionString) =>
  (pageConsents: Consent[]) => {
    const invertedConsents = pageConsents.map((consent) => {
      if (!consent.id.includes(input)) {
        return consent;
      }

      const invertedConsent = {
        id: consent.id.replace(input, output),
        description: consent.description,
        name: consent.name,
      } as Consent;

      if (!consent.hasOwnProperty('consented')) {
        return invertedConsent;
      }
      
      return {
        consented: !consent.consented,
        ...invertedConsent,
      } as Consent;
    });

    return invertedConsents;
  };
```

src/server/lib/idapi/invertOptOutConsents.ts
```diff
@@ -0,0 +1,29 @@
...
+ * See PR https://github.com/guardian/identity/pull/2079 for more details why this is needed.
+ * This is horrible but necessary due to business requirements because we cannot change
+ * the consent IDs "eg. Profiling_Optout" on the server data model or downstream integrations
+ */
+
+type ConversionString = '_optout' | '_optin';
+
+const invertConsents =
+  (input: ConversionString, output: ConversionString) =>
+  (pageConsents: Consent[]) => {
```
dominickendrick, line 13: A minor point, is it worth using curly braces and an explicit `return` here? I has to double check it was a function returning a function ...

vlbee:
I've just realised the outgoing payload is modelled on `UserConsent` not `Consent` so need to add a very similar second function

Is there a way to use TS generics to make this more dry? The `consented` property is optional on one not other.  
```
const invertUserConsents = (
  input: ConversionString,
  output: ConversionString,
) => {
  return (pageConsents: UserConsent[]) => {
    return pageConsents.map((consent) => {
      if (!consent.id.includes(input)) {
        return consent;
      }

      return {
        id: consent.id.replace(input, output),
        consented: !consent.consented,
      } as UserConsent;
    });
  };
};
```

dominickendrick:
> I've just realised the outgoing payload is modelled on `UserConsent` not `Consent` so need to add a very similar second function
> 
> Is there a way to use TS generics to make this more dry? The `consented` property is optional on one not other.
> 
> ```
> const invertUserConsents = (
>   input: ConversionString,
>   output: ConversionString,
> ) => {
>   return (pageConsents: UserConsent[]) => {
>     return pageConsents.map((consent) => {
>       if (!consent.id.includes(input)) {
>         return consent;
>       }
> 
>       return {
>         id: consent.id.replace(input, output),
>         consented: !consent.consented,
>       } as UserConsent;
>     });
>   };
> };
> ```
```
export interface Consent {
  id: string;
  name: string;
  description: string;
  consented?: boolean;
}

export interface UserConsent {
  id: string;
  consented: boolean;
}

type ConversionString = '_optout' | '_optin';

const invertUserConsents = <UserType extends UserConsent | Consent>(
  input: ConversionString,
  output: ConversionString,
): (pageConsents: UserType[]) => UserType[] => {
  return (pageConsents: UserType[]) => {
    return pageConsents.map((consent) => {
      if (!consent.id.includes(input)) {
        return consent;
      }

      return {
        id: consent.id.replace(input, output),
        consented: !consent.consented, 
      } as UserType;
    });
  };
};
```

https://www.typescriptlang.org/play?ssl=31&ssc=3&pln=1&pc=1#code/KYDwDg9gTgLgBASwHY2FAZgQwMbDgYQiQGdgU4BvAWACg5EATALjmJimQHMBuW+pTAFtgLNhyQ8+cBsGLYOYGAiKj2XXnTjYipFMAYB+FgCMIEADbBMSDQF9atUJFiI9GHHgCqpKIRJl4ak0EZlY1CQ16bX89UNMLKxtaexpaGABPMDw-ADc0YmUkAGVwzjgAXjgAcgB9CEUIAFcYKrgAH2q6xWQqjVpotlc82G80P10YYgq4AB5RqAAVTLxQVCQGKfnxgPaCHQCAPgAKKWQwZpZc-MKS8U4AGikmmHOYS6JhgqJbrkeaAEoWEcwJhOMBtihiCx5kssgBtAC6-wqBzgMOWiJRlCkUGAMEaUCQcGBoPB+0h0J8sOAiOR5VRQXo9Fx+MJcBBYIhkwAdIJMGAjkcBgE6QypEzEOhiQBCYUobkhBVIbDmRoyYhHM7Nf7IxkSpksglEuUwSL6lLig14o3YzT6xgsE0Khjc3FgcweTVIV73ODPV7-P72qLk1ChWWh7km-S+gD0sbg6SacCQwH0cBgEDgAHdoABrP3NHMAC0w8Ez0izCHLxbwcjImA4WeztdxiHgCCmjXWwHQyHT0st9FscEwmypyzNw-+Ztsdm4QA

coldlink (approved):

PR #1410, opened 2022-01-20
Okta | Sign In with Email and Password

coldlink:
## What does this change?

Replaces #638 with a cleaned up commit history and updated with existing Okta changes.

The same comments from that PR apply and have been addressed.

Each commit is mostly a clean change and should be possible to follow along that way.

The main focus of this PR is for users to sign in with email and password and set the Okta session cookie along with the IDAPI cookies.

In general the steps are sumarised as follows, assuming okta is enabled, this does not detail all technical requirements, just the main high level:

- User navigates to `/signin`
  - Check `EncryptedState` cookie for `signInRedirect` value
    - if `true` render the sign in page, as we've checked for an existing session
    - if `false` or `undefined` we need to check for an existing session 
      - Check for existing session by performing auth code flow by redirecting user to `/authorize`
      - Okta will redirect to callback endpoint `/oauth/authorization-code/callback`
        - if okta session exists, complete the auth code flow and refreshing identity cookies, redirect to original `returnUrl`
        - if no okta session, okta returns `login_required` error, intercept this
          - set the `EncryptedState` cookie for `signInRedirect` value to `true`
          - redirect to sign in, now that `signInRedirect` is `true`, the page will render
  - If sign in page shown
    - user enters email and password, makes request to gateway `POST /signin`
    - Use okta authenticate endpoint with the email and password `/api/v1/authn`
      - If email/password wrong, show `email or password was incorrect` error
      - If correct, okta returns a single use `sessionToken` which can be exchanged for an session cookie
        - to exchange this for a session cookie, we perform the auth code flow, pass the `sessionToken` to `/authorize` and redirect to okta `/authorize`
          - Okta will redirect to callback endpoint `/oauth/authorization-code/callback`
          - if all good, a session cookie will be set on okta domain, complete the auth code flow and refreshing identity cookies
          - redirect to original `returnUrl`

I've added a number of comments where appropriate to add reasoning to why things are in a particular way.

## Other change
I refactored the `crypto` library that we used to encrypt/decrypt the `EncryptedState` cookie (https://github.com/guardian/gateway/pull/1410/commits/45b202b8c896e58ade2cf970396c9a8bea127628). This removes the requirement of `getConfiguration` being needed by those methods, in turn meaning we're able to use the module outside of the server, specifically for cypress where we're now able to test and set the encrypted state from cypress tests (https://github.com/guardian/gateway/pull/1410/commits/72f12649e5ec4b830d93b2b7fc3f04eb85a7f43c).

#### Authentication API
[...]

PR #1512, opened 2022-02-25
Okta documentation | Sign in, Sign out, OAuth 2.0, and OpenID Connect

coldlink:
## What does this change?

Adds a whole bunch of documentation around OAuth 2.0, OpenId Connect, the sign in with email and password with Okta (much of which is a cleaned up version of this PR description: https://github.com/guardian/gateway/pull/1410), and sign out with Okta (RFC).

We also refactor the `performAuthorizationCodeFlow` method into it's own file as it was no longer just being used by the sign in flow.


dominickendrick (approved):
Great start for the documentation, I'm sure things will change over time, but its worth merging this so we have something to base future updates on!

PR #1513, opened 2022-02-28
Islets - Gateway Islands 🏝️

coldlink:
## Accessibility


coldlink:
This likely won't be merged in until after the release MVP3 of the Identity migration, at which point it will be safer to make architectural changes like this PR, and we'll have more time to work on this.

Until that point I'll keep this PR up to date with `main`, and start migrating functionality over in their own branches tracked against this PR.

mxdvl:
I know this was opened a while back, but I’ll still nominate this as one of May’s best PRs:

https://github.com/guardian/playground/issues/14

src/client/static/islet.ts
```diff
@@ -0,0 +1,140 @@
+import { createElement } from 'react';
+import { hydrate } from 'react-dom';
+
+/**
+ * use `whenIdle` to execute the given callback when the browser is 'idle'
+ *
+ * @param callback Fired when `requestIdleCallback` runs, when event loop is idle. If not executed within the `timeout`, the callback is added to the event loop. If `requestIdleCallback` is not supported, use `setTimeout` fallback.
+ */
+const whenIdle = (callback: () => void) => {
+  if ('requestIdleCallback' in window) {
```
liamness, line 10: I wonder if you could do `window.requestIdleCallback?.()`?

src/client/static/islet.ts
```diff
@@ -0,0 +1,140 @@
...
+        module[name](props);
+      }
+
+      element.setAttribute('data-gu-ready', 'true');
+    })
+    .catch((error) => {
+      if (name && error.message.includes(name)) {
+        console.error(
+          `🚨 Error importing ${name}. Components must live in the root of /components and follow the [MyComponent].importable.(m?)(j|t)s(x?) naming convention 🚨`,
+        );
```
liamness, line 88: Do we want these errors  and logs to appear in non-development / staging environments?

src/client/static/index.tsx
```diff
@@ -43,7 +64,27 @@ if (window.Cypress) {
...
+  if (params.has('useIslets')) {
+    console.log('init islets');
+    import('./islet').then(({ init }) => {
+      init();
+    });
+  } else {
+    import('./hydration').then(({ hydrateApp }) => {
+      hydrateApp(config);
+    });
+  }
```
liamness, line 89: I wonder if it would be best to load `./islet` and `./hydration` as part of the main bundle instead of loading them asynchronously using `import()`. I'm just concious that this would mean at least two bundles would need to be loaded and ran before the islet bundles are requested.

src/client/static/islet.ts
mxdvl: If you want logging in prod, I would recommend using `@guardian/libs`’s [logger](https://github.com/guardian/libs/tree/main/src/logger)

src/client/static/index.tsx
coldlink: That's not what's happening. It only loads the `./islet` or `./hydration` depending on the environment it's running in, i.e whether we should use islets, or whether we should hydrate the whole bundle (as what's happening now). Once Islets are live we won't have the need for the `./hydration` file, so can move the `./islet` into the `main` bundle then.

src/client/lib/isletStateStore.ts
```diff
@@ -0,0 +1,23 @@
+import create from 'zustand/vanilla';
+
+/**
+ * @name IsletState
+ * @description Add properties to this interface to share between Islet components.
+ */
+interface IsletState {
+  setIsletState: (state: Partial<IsletState>) => void;
```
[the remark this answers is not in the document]
coldlink: Updated the type to `Partial<Omit<IsletState, 'setIsletState'>>` so that it's not longer circular/recursive.

PR #1552, opened 2022-03-08
Handle case where email address is quite long on mobile

liamness:
## What does this change?

Stops long emails from causing horizontal scrolling to be possible on mobile.

Not the fanciest of solutions (I've just hidden overflowing content) but this is a bit of an edge case and all alternative solutions I could think of have drawbacks.

## How to test

Go to the Welcome component's stories in Storybook and enter a really long email.

### Before

Screenshot 2022-03-08 at 17-22-39 Pages Welcome - with email ⋅ Storybook

### After

Screenshot 2022-03-08 at 17-32-47 Pages Welcome - with email ⋅ Storybook

### DIFF
--- a/src/client/pages/Welcome.stories.tsx
+++ b/src/client/pages/Welcome.stories.tsx
@@ -1,5 +1,5 @@
 import React from 'react';
-import { Meta } from '@storybook/react';
+import { ComponentMeta, Story } from '@storybook/react';
 
 import { Welcome } from './Welcome';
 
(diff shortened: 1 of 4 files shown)


src/client/pages/Welcome.stories.tsx
```diff
@@ -1,50 +1,43 @@
...
-);
-PasswordAlreadySet.story = {
-  name: 'with password already set',
+export const PasswordAlreadySet = Template.bind({});
+PasswordAlreadySet.args = {
+  email: 'example@theguardian.com',
+  submitUrl: '',
+  passwordSet: true,
 };
+PasswordAlreadySet.storyName = 'with password already set';
```
liamness, line 43: Just amending this story so you can easily change the parameters via the action tab in storybook.

src/client/pages/Welcome.tsx
```diff
@@ -44,11 +48,11 @@ export const Welcome = ({
...
       showContinueButton={false}
     >
-      <p css={[text, greyBorderTop, autoRow()]}>
+      <p css={[text, greyBorderTop, passwordInstructions, autoRow()]}>
         {passwordSet
           ? 'Password already set for '
-          : 'Please create a password for '}{' '}
-        {<span css={emailSpan}>{email}</span> || 'your new account'}
+          : 'Please create a password for '}
+        {email ? <span css={emailSpan}>{email}</span> : 'your new account'}
```
liamness, line 55: The "your new account" text would never show! 😬

ob6160: good spot, maybe worth extending/adding an assertion in one of our tests to cover it along with this fix

there's already an assertion checking the `Password already set for` message here: https://github.com/guardian/gateway/blob/main/cypress%2Fintegration%2Fmocked%2Fwelcome.spec.ts#L135

ob6160 (approved):
looks good, I'm just wondering — was `text-overflow: ellipsis;` considered?

liamness:
@ob6160 I don't believe `text-overflow: scroll;` is a thing. Perhaps `overflow: auto;` might've been an option on the paragraph element, however I scrollable elements mid-page (particularly horizontally scrolling ones) can be confusing. I don't believe this is essential information for the user to see in any case (it is their email, they should already know it, or at least recognise it from the part they can see).

An ellipsis is one of the alternative solutions I considered, but as I said, it does have drawbacks, would require tweaks to the styling that I would rather avoid without good reason.

Good point on adding a new assertion, I'll go do that.

ob6160:
> @ob6160 I don't believe `text-overflow: scroll;` is a thing. Perhaps `overflow: auto;` might've been an option on the paragraph element, however I scrollable elements mid-page (particularly horizontally scrolling ones) can be confusing. I don't believe this is essential information for the user to see in any case (it is their email, they should already know it, or at least recognise it from the part they can see).
> 
> An ellipsis is one of the alternative solutions I considered, but as I said, it does have drawbacks, would require tweaks to the styling that I would rather avoid without good reason.
> 
> Good point on adding a new assertion, I'll go do that.

that reasoning sounds good to me. might be worth seeing if ux/design have any alternative opinions on this behaviour (if you haven't already that is)

liamness:
Ah yes, will check with Dom as he's the one who originally raised this issue.

liamness:
New solution (after chatting with Dom):

# Long email at 320px

image

# Short email at 375px
image

Will push momentarily

ob6160:
> New solution (after chatting with Dom):
> 
> # Long email at 320px
> image
> 
> # Short email at 375px
> image
> 
> Will push momentarily

looks great :)

PR #1555, opened 2022-03-09
Fix spacing issue on registration page

liamness:
## What does this change?

Spacing below "create password" button on registration page

### Before

image

### After

image

## How to test

Checking Storybook should suffice.

### DIFF
--- a/src/client/components/PasswordForm.tsx
+++ b/src/client/components/PasswordForm.tsx
@@ -28,7 +28,7 @@ import {
 import AwesomeDebouncePromise from 'awesome-debounce-promise';
 import { AutoRow } from '@/client/styles/Grid';
 import { MainForm } from '@/client/components/MainForm';
-import { passwordButton } from '../styles/Consents';
+import { passwordButton, controls } from '../styles/Consents';
 import { trackFormFocusBlur, trackFormSubmit } from '@/client/lib/ophan';
 import { logger } from '../lib/clientSideLogger';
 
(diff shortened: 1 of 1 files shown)


src/client/components/PasswordForm.tsx
```diff
@@ -179,24 +175,22 @@ const ValidationMessage = ({
...
 }: {
   isWeak: boolean;
   isTooShort: boolean;
   isTooLong: boolean;
   isChecking: boolean;
-  noMargin?: boolean;
 }) => {
   if (isTooShort) {
-    return <TooShort noMargin={noMargin} />;
+    return <TooShort />;
```
ob6160, line 185: was the `noMargin` on these here for a reason? I'm guessing so we could match the designs exactly? from looking at the chromatic diffs there's a small difference between the gap between the input and the button when the lower message is displaying and when it isn't, did my best to capture the difference with a floaty ruler app I found for Mac

Screenshot 2022-03-09 at 17 23 27
Screenshot 2022-03-09 at 17 24 07

ob6160: I don't think it really needs changing if it looks good on mobile/across other viewports, might be worth a quick look at the original Figma designs to make sure though

liamness: I think it was to account for the fact that the password form needed to reside in two different layouts. The one used in `/welcome` previously needed an extra margin here. Although tbh the styling wasn't very consistent even so, e.g. when there was an error these validation messages didn't show, so the margin they gave was lost and it looked strange (Dom has raised this before).

ob6160: that makes sense, can do some changes for consistency later on down the road then

ob6160 (approved):

PR #1558, opened 2022-03-11
Setup profile.thegulocal.com with Okta in DEV

coldlink:
## What does this change?

We recently changed the Okta url to sit on the profile subdomain, while the routing for this is easy to control in PROD and CODE thanks to vcl, it's a bit harder to do locally/in the DEV environment as the OAuth/OIDC tokens won't verify thanks to the `issuer` (`iss`) being different.

In https://github.com/guardian/identity-platform/pull/505 we set up the local nginx configuration for profile.thegulocal.com to point okta to the correct development environment. This is done through the use of the `gu_okta_env` cookie which can be set to `CODE` or `DEV` depending on which environment of okta you want to point to locally. This returns a "X-GU-Okta-Env" header which we can use to control this `issuer` field.

So when `stage` is `DEV`, and the `X-GU-Okta-Env` exists, we set up a new `OpenIdClient` with the correct `issuer` set so that the tokens we get back from okta match the `issuer` checks.

This allows issuers to develop locally with Okta tokens and an Okta session set locally too.

## How to test

Successfully go through the Okta web sign in flow on your local dev environment when using nginx.


docs/setup.md
```diff
@@ -45,6 +41,12 @@ This ensures that you don't get any certificate errors when running gateway and
 FetchError: request to https://idapi.thegulocal.com/auth?format=cookies failed, reason: unable to verify the first certificate\n
 ```
 
+### Environment File
+
+Populate `.env` and `cypress-ete.env` files by using the examples from [`.env.example`](../.env.example) and [`cypress-ete.env.example`](../cypress-ete.env.example.example). These `.env` files should **never** be committed as they store secrets.
```
liamness, line 46: I corrected an issue with this [here](https://github.com/guardian/gateway/pull/1569/files#diff-4ca53116333428d7d34e6d2a9a64069655c8c37d5d36b856320ac1a0c4cb9b23R22), as it doesn't actually link to a file.

coldlink: ah i didn't notice that, i'll bring that into this PR too

liamness (approved):
Looked through the changes here and in `identity-platform`, happy to give a 👍

PR #1569, opened 2022-03-14
Testing impact of removing image on opt in prompt

liamness:
## What does this change?

Creates two variants of the post sign in prompt, one with an image and one without, to test the impact on opt in rates

## How to test

Set `GU_mvt_id` to 0 to see the image, and to 1 to remove it.

Consider effect on opt in rates after test has run for one week.

### Mobile, variant with image

image

### Mobile, variant without image

image

### Tablet, variant with image

image

### Tablet, variant without image

image

## Accessibility


liamness:
[Test removing the image for the marketing consent in sign in flow (S)](https://trello.com/c/6NqD5kjm/3853-test-removing-the-image-for-the-marketing-consent-in-sign-in-flow-s)

liamness:
Rolling out the post sign in prompt has broken some E2E tests. Think I know how to fix this for the tests not using Okta, but the Okta ones may require a bit more investigation. Also raises the question of whether there should be coverage for the prompt in the E2E tests (I am thinking it should really be one or the other).

PR #1570, opened 2022-03-15
Set validate email and password set securely flags in Okta

rebecca-thompson:
## What does this change?
When a user changes their password via the welcome or set password flow, or resets their password, we set the `emailValidated`, `lastEmailValidatedTimestamp`, `passwordSetSecurely` and `lastPasswordSetSecurelyTimestamp` values in Okta. These fields are then used to determine whether a user has access to certain pages or if they need to perform an additional security measure.

## How to test
This has been tested locally and in CODE.


rebecca-thompson:
[Write emailValidated & passwordSetSecurely fields to Okta in gateway](https://trello.com/c/yXnv4d2q/3843-write-emailvalidated-passwordsetsecurely-fields-to-okta-in-gateway)

PR #1576, opened 2022-03-17
Clean up opt in prompt a/b test

liamness:
## What does this change?

Adding better error handling (at the suggestion of @ob6160) around the fetching of consents, and seeing if this at least allows e2e tests to pass.

I should probably also change the e2e tests so that they don't rely on a network error being caught to pass. This would likely involve creating specific test cases where the user will be sent to the prompt, and also where they won't. This PR is in draft until that is added.

## How to test

N/A, this should not result in any user facing changes.

### DIFF
--- a/src/client/components/GlobalError.tsx
+++ b/src/client/components/GlobalError.tsx
@@ -75,7 +75,7 @@ export const GlobalError = ({ error, link, left }: GlobalErrorProps) => {
           <div>
             {error}
             &nbsp;
-            <ExternalLink href={link.link} css={errorLink} subdued={true}>
+            <ExternalLink href={link.link} css={errorLink} target="_blank">
               {link.linkText}
             </ExternalLink>
           </div>
--- a/src/server/routes/signIn.ts
+++ b/src/server/routes/signIn.ts
@@ -233,35 +233,48 @@ const oktaSignInController = async (
   }
 };
 
+const optInPromptController = async (
+  req: Request,
+  res: ResponseWithRequestState,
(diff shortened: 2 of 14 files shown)


src/server/routes/signIn.ts
```diff
@@ -238,21 +238,29 @@ router.get(
...
-        },
-      }),
-      pageTitle: 'Signed in',
-    });
-    res.type('html').send(html);
+    const { returnUrl } = state.pageData;
+    const { defaultReturnUri } = getConfiguration();
+    const redirectUrl = returnUrl || defaultReturnUri;
+
+    try {
```
ob6160, line 245: might be worth adding a comment to the e2e test about how these handle the case where the request to IDAPI was failing for an unknown reason

ob6160 (approved):

src/client/components/GlobalError.tsx
```diff
@@ -75,7 +75,7 @@ export const GlobalError = ({ error, link, left }: GlobalErrorProps) => {
           <div>
             {error}
             &nbsp;
-            <ExternalLink href={link.link} css={errorLink} subdued={true}>
+            <ExternalLink href={link.link} css={errorLink} target="_blank">
```
liamness, line 78: Amended the appearance of this as per gchat with Dom. Added `target="_blank"` for good measure as I don't think users will want to be navigated away just to report this issue?

coldlink (approved):
nice changes 👍🏾

PR #1579, opened 2022-03-21
Add e2e test coverage for opt in prompt redirect

liamness:
## What does this change?

Just adds an extra test to the idapi / okta e2e tests, covering the case where the redirect to `/signin/success` happens.


cypress/integration/ete/sign_out/sign_out.spec.ts
```diff
@@ -8,6 +10,11 @@ describe('Sign out flow', () => {
 
   context('Signs a user out', () => {
     it('Removes IDAPI log in cookies and dotcom cookies when signing out', () => {
+      // Disable redirect to /signin/success by default
+      cy.setCookie(
+        'GU_ran_experiments',
+        stringify({ OptInPromptPostSignIn: Date.now() }),
+      );
```
liamness, line 17: I don't really understand how this was passing previously...

.storybook/preview-head.html
```diff
@@ -23,161 +30,122 @@
   @font-face {
     font-family: GT Guardian Titlepiece;
     font-weight: 700;
-    src: url(https://assets.guim.co.uk/static/frontend/fonts/guardian-titlepiece/noalts-not-hinted/GTGuardianTitlepiece-Bold.woff2)
-        format('woff2'),
-      url(https://assets.guim.co.uk/static/frontend/fonts/guardian-titlepiece/noalts-not-hinted/GTGuardianTitlepiece-Bold.woff)
-        format('woff'),
-      url(https://assets.guim.co.uk/static/frontend/fonts/guardian-titlepiece/noalts-not-hinted/GTGuardianTitlepiece-Bold.ttf)
-        format('ttf');
+    src: url(/GTGuardianTitlepiece-Bold.woff2) format('woff2'),
```
coldlink, line 33: any reason why we're self hosting the fonts now?

src/client/app.tsx
```diff
@@ -30,19 +30,25 @@ export const App = (props: ClientState) => {
...
             display: flex;
             flex-direction: column;
           }
+          *,
+          *:before,
+          *:after {
+            box-sizing: inherit;
+          }
           // Badge is hidden for Gateway, because we're using
           // the legal text to do this job
```
coldlink, line 53: I know this wasn't introduced here but though it's worth changing to valid css comments

```suggestion
          /* Badge is hidden for Gateway, because we're using
           the legal text to do this job */
```

.storybook/preview-head.html
coldlink: I don't see a benefit over using the CDN? As it just bloats the project slightly. Unless I'm missing something

liamness: It was a potential cause of the font rendering issue, suggested by Chromatic support. I will revert this change, we can revisit if issues continue to crop up.

coldlink (approved):

PR #1587, opened 2022-03-24
Handle attempts to register when a user is already signed in

raphaelkabo:
## What does this change?

A couple of separate behaviours and approaches to be implemented:

- Given a user has an Okta cookie set, when they `POST` to `/register`:
  - [ ] If the Okta cookie is valid, then they are redirected to Manage Profile
  - [ ] If the Okta cookie is invalid, then the cookie is removed and they are redirected to `/register` to try again without a cookie
- Given a user has an Okta cookie set, when they visit the `/register` page:
  - [ ] If the Okta cookie is valid, then they are redirected to Manage Profile
  - [ ] If the Okta cookie is invalid, then the cookie is removed and they are redirected to `/register`
- Given a user has IDAPI cookies set, when they `POST` to `/register`:
  - [ ] Then this is logged (so we want to see how many IDAPI-authed users do this)
- Given a user has IDAPI cookies set, when they visit the `/register` page:
  - [ ] ?


cypress/integration/mocked/okta_register.spec.ts
```diff
@@ -0,0 +1,205 @@
...
+      cy.getCookie('SC_GU_LA').should('not.exist');
+      cy.getCookie('SC_GU_U').should('not.exist');
+    });
+  });
+
+  context('Signed in user posts to /register', () => {
+    beforeEach(() => {
+      cy.mockPurge();
+      cy.clearCookies();
+      //we visit the healthcheck page to make sure the cookies are clearer from the browser
```
liamness, line 70:
```suggestion
      // we visit the healthcheck page to make sure the cookies are cleared from the browser
```

cypress/integration/mocked/okta_register.spec.ts
```diff
@@ -0,0 +1,205 @@
...
+
+      cy.intercept('https://manage.theguardian.com').as('manage');
+
+      cy.mockPattern(204, {}, '/api/v1/users/userId/sessions');
+
+      cy.visit('/register?useOkta=true');
+
+      cy.setCookie('sid', `the_sid_cookie`);
+
+      cy.get('input[name="email"]').type('example@example.com');
```
liamness, line 102: This is a suggestion not a requirement, but we have `@cypress/testing-library` now and could use queries that are more akin to how a user would navigate (either visually, or using a screen reader) the page. I would suggest `cy.findByLabel()` here for instance instead of querying the DOM directly.

cypress/integration/mocked/okta_register.spec.ts
```diff
@@ -0,0 +1,205 @@
...
+      cy.mockNext(200, {
+        userType: 'new',
+      });
+      cy.mockNext(200, {
+        status: 'success',
+        errors: [],
+      });
+      cy.get('[data-cy=register-button]').click();
```
liamness, line 110: Similarly this could be `cy.findByRole()`.

cypress/integration/mocked/okta_register.spec.ts
```diff
@@ -0,0 +1,205 @@
...
+        status: 'success',
+        errors: [],
+      });
+
+      cy.get('[data-cy=register-button]').click();
+
+      cy.intercept('/register*').as('register');
+
+      //we redirect to /register and unset the sid cookie, so we need to wait till the redirect has happened
+      cy.wait('@register');
```
liamness, line 142: IIRC, `cy.url().should()` already has retry logic in my experience, so you shouldn't need to wait for a redirect. Any reason you particularly need to use `cy.wait()` here, maybe `cy.location()` doesn't have the retry logic (but I think cypress assertions generally do).

liamness: Actually I'm looking at the docs and this aliasing seems pretty cool, maybe we should be doing that instead of just expecting cypress to keep retrying the assertion. I'm easy either way.

https://docs.cypress.io/api/commands/intercept#Aliasing-an-intercepted-route

cypress/integration/mocked/okta_register.spec.ts
```diff
@@ -0,0 +1,205 @@
...
+          idp: {
+            id: '01a2bcdef3GHIJKLMNOP',
+            type: 'OKTA',
+          },
+          mfaActive: true,
+        },
+        '/api/v1/sessions/the_sid_cookie',
+      );
+
+      cy.intercept('https://manage.theguardian.com').as('manage');
```
liamness, line 175: I don't think this alias is used?

liamness (approved):
Couple of minor comments, overall looks good

cypress/integration/mocked/okta_register.spec.ts
dominickendrick: Yeah good suggestion, there are 37 instances of this call in the tests, so I can address this in a seperate PR

cypress/integration/mocked/okta_register.spec.ts
dominickendrick: Yeah good suggestion, there are 37 instances of this call in the tests, so I can address this in a seperate PR

cypress/integration/mocked/okta_register.spec.ts
dominickendrick: Good catch

PR #1589, opened 2022-03-25
Add personalised advertising permission prompt in onboarding flow

vlbee:
## What does this change?

This PR adds thd `personalised_advertising` permission to the `ConsentsData` page in the onboarding flow. Originally we were going to AB test but this requirement was was dropped.

The advertising consent defaults to false - it is an opt in consent. 

The advertising consent will not show on the `ConsentsDataPage` unless the use has opted into all Cmp banner cookies and the user does not have an ad free cookie set for their session (because they have just purchased a digital subscription and are completing the registration process)

**Requirements**

- [x] cmp consent check - this happens clientside implemented [here](https://github.com/guardian/gateway/pull/1589/commits/fface6463a548d6b4af50f3cd444e1a7713a271c)
- [x] **new requirment** - exclude users with active digital subscriptions
- [x] Ensure consent confirmation page displays correctly
- [x] cypress tests

## How to test

**Styling checked in:**
- [x] Firefox
- [x] Safari
- [x] Chrome
- [x] IE11

**Tested in Code**

- [x] Does not appear for user that has rejected cmp
- [x] Does not appear for a user that has accepted cmp but has ad free cookie
- [x] Appears for user that has accepted cmp - default to off
- [x] Appears for user that has accepted cmp - turned on if user has previously accepted
- [x] Confirmation page appears correctly in all cases
- [x] Consent is only set upon hitting continue.

## Images

[Figma designs](https://www.figma.com/file/z74mPkMGzcATz3ZxrHetrs/Onboarding-UX?node-id=3449%3A51332)

**Your Data Page**
image
**Confirmation Page**
image

## Accessibility


src/client/lib/hooks/useCmpConsent.ts
```diff
@@ -0,0 +1,55 @@
...
+
+      if (state.aus) {
+        return resolve(state.aus.personalisedAdvertising);
+      }
+
+      // this shouldn't ever be hit, but this is here as safety
+      return resolve(false);
+    });
+  });
+};
```
vlbee, line 37: This logic has been lifted from DCR. I will add a comment and check for any tests we can port over.

src/client/pages/ConsentsDataPage.tsx
```diff
@@ -12,13 +15,22 @@ export const ConsentsDataPage = () => {
...
-      id={profiling?.id || Consents.PROFILING}
-    />
+    <>
+      {shouldPersonalisedAdvertisingPermissionRender ? (
+        <ConsentsData profiling={profiling} advertising={advertising} />
+      ) : (
+        <ConsentsData profiling={profiling} />
+      )}
+    </>
   );
```
liamness, line 35: Potentially nitpicky / code golf, but could avoid the use of a fragment here:

```tsx
return shouldPersonalisedAdvertisingPermissionRender ? (
  <ConsentsData profiling={profiling} advertising={advertising} />
) : (
  <ConsentsData profiling={profiling} />
);
```

Or you could even conditionally add the advertising prop somehow instead of having two separate JSX expressions. For example:

```tsx
return (
  <ConsentsData
    profiling={profiling}
    {...(shouldPersonalisedAdvertisingPermissionRender && { advertising })}
  />
);
```

(spreading `false` in an object doesn't do anything)

Or you could even do

```tsx
let props: ComponentProps<typeof ConsentsData> = { profiling };

if (hasCmpConsent && !isDigitalSubscriber) props = { ...props, advertising };

return <ConsentsData {...props} />;
```

Lots of ways to skin this particular cat.

src/client/styles/Consents.ts
```diff
@@ -23,6 +23,12 @@ export const text = css`
   max-width: 640px;
 `;
 
+export const subText = css`
+  margin: 0;
+  ${textSans.small()}
+  max-width: 640px;
```
liamness, line 29: I can't tell what the `max-width` is for here, in the usages I can see of `subText` the grid layout is what constrains the text width instead.

src/client/lib/hooks/useCmpConsent.ts
```diff
@@ -0,0 +1,60 @@
...
+
+const hasRequiredCmpConsents: () => Promise<boolean> = async () => {
+  return await hasRequiredConsents();
+};
```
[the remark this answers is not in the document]
vlbee: I think it might also be worthwhile to suggest adding this or similar to the CMP library itself as .hasConsentedToAll().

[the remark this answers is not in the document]
vlbee: This was just added to CMP lib last week so will utilize and update sign in gate code as well : https://github.com/guardian/consent-management-platform#consentstatecantarget

src/client/pages/ConsentsConfirmationPage.tsx
```diff
@@ -19,8 +19,17 @@ export const ConsentsConfirmationPage = () => {
     (consent) => consent.id === Consents.PROFILING && consent.consented,
   );
 
+  const optedIntoPersonalisedAdvertising = !!consents.find(
+    // If consent option present and consented === true, this means the user has expressed a
+    // preference to NOT be contacted - eg. They opted OUT
```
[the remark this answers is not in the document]
vlbee: I've updated all the comments as they were old from before we converted the `profiling_optout` to `profiling_optin`

Adding a note to all occurrences

src/client/pages/ConsentsDataPage.tsx
vlbee: I like the middle way, keeping the logic where it renders so you don't have to scroll :) 
But I do not endorse skinning cats and no animals were harmed writing this code.

src/client/lib/hooks/useAdFreeCookie.ts
```diff
@@ -17,5 +17,6 @@ export const useAdFreeCookie: () => boolean = () => {
     setIsDigitalSubscriber(digitalSubscriberCookieExists);
   }, []);
 
+  console.log(`has ad free: ${hasAdFree}`);
```
liamness, line 20: Stray console log

src/client/lib/hooks/useCmpConsent.ts
```diff
@@ -3,55 +3,23 @@ import { ConsentState } from '@guardian/consent-management-platform/dist/types';
...
-      }
-
-      if (state.aus) {
-        return resolve(state.aus.personalisedAdvertising);
-      }
-
-      // this shouldn't ever be hit, but this is here as safety
-      return resolve(false);
+    onConsentChange((state: ConsentState) => {
+      return resolve(state.canTarget);
```
liamness, line 12: This isn't going to cause any problems, but I thought I'd mention that you don't really need to return from within a Promise, I expect the previous code was just returning as a way to break out of the function.

src/client/lib/hooks/useAdFreeCookie.ts
```diff
@@ -0,0 +1,21 @@
...
+import { getCookie } from '@guardian/libs';
+
+// User Attributes API cookies are dropped on subscription on support.theguardian.com for use by frontend
+// User Cookie Information: github.com/guardian/members-data-api/blob/main/README.md
+export const useAdFreeCookie = (): boolean => {
+  const [hasAdFree, setHasAdFree] = useState(false);
+  const [isDigitalSubscriber, setIsDigitalSubscriber] = useState(false);
+
+  useEffect(() => {
+    const adfreeCookieExists = !!getCookie({ name: 'GU_AF1' }); // Ad Free session cookie valid 48hrs
```
vlbee, line 11: This is a session cookie set by Support. Expiry is in the value which is a Tz but not sure it's worth checking the expiry here.

src/client/lib/hooks/useCmpConsent.ts
```diff
@@ -0,0 +1,28 @@
+import { useState, useEffect } from 'react';
+import { ConsentState } from '@guardian/consent-management-platform/dist/types';
+import { onConsentChange } from '@guardian/consent-management-platform';
+
+/**
+ * ConsentState.canTarget checks if user has consented to all consents
+ * https://github.com/guardian/consent-management-platform#consentstatecantarget
+ */
+const hasRequiredCmpConsents = (): Promise<boolean> => {
+  return new Promise((resolve) => {
```
arelra, line 10: Just fyi if you don't need to listen for subsequent consent changes then the new method `onConsent` method will give you a promise with the consent state:
https://github.com/guardian/consent-management-platform#onconsent

vlbee: Thanks. I think onConsent will work for us.

src/client/lib/hooks/useAdFreeCookie.ts
vlbee: I'm removed the check on the subscriber cookie. This will show base only on the Ad free cookie now. 

https://github.com/guardian/gateway/pull/1589/commits/d54080ceec4a87c2bbd83ae4873cae9ac60451d6

cypress/support/commands/manageCmp.ts
```diff
@@ -0,0 +1,53 @@
...
+      enableCMP: typeof enableCMP;
+      acceptCMP: typeof acceptCMP;
+      declineCMP: typeof declineCMP;
+    }
+  }
+}
+
+export const disableCMP = () => {
+  return cy.setCookie('gu-cmp-disabled', 'true', {
+    log: true,
```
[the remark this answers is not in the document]
vlbee: fixed

PR #1590, opened 2022-03-28
Add a frontend route to render email templates as HTML and plain text

raphaelkabo:
## What does this change?

1. Add a new route, `/email/:template`, which takes in a valid email template name (`account-exists`, `account-without-password-exists`, `create-password`, `no-account`, `reset-password`, `verify`, or `accidental-email`). The route returns a JSON object containing `plain` and `html` versions of the email template as JSONified strings. As the template content is static for the lifetime of the server, the HTML is rendered by MJML from an `mjml-react` template when the server starts, and is then cached in memory to prevent unnecessary re-rendering.
2. Add tests for the email rendering function.
3. Create a template for the Okta 'accidental email' email, which is a fallback email that might be sent by Okta when something goes wrong, or for debugging.
4. As part of recreating the accidental email template so that it would function exactly as the old static `accidentalEmail.html` file in IDAPI functioned, it was necessary to pass through a `cssClass` prop to the `Text` email component. This is explained in this commit message: https://github.com/guardian/gateway/pull/1590/commits/81af75129bc87a4c774e98115c0b8f2b575742aa
5. As part of editing the reset password template so that it would function exactly as the old static `resetPassword.html` file in IDAPI functioned, I added a new prop, `hiddenText`, to the email `Footer` component, which allows us to pass a hidden div through to the final template containing a text string. More details in this commit message: https://github.com/guardian/gateway/pull/1590/commits/477f5d8df28ae2e88b17b8917d3b32c7a700941e
6. MJML did not like me using `display: {{CTA_DISPLAY}}` as a CSS rule (this value gets replaced by Terraform). I've changed it to `[[CTA_DISPLAY]]`.

The route only renders a whitelisted set of email template names, so this should be safe to live on the frontend. The rendered HTML is cached at server startup, so multiple callls to the endpoint won't tax server resources.

## Future work

- When we start using Gateway to send emails, we could make use of the cached email rendering function and fetch those pre-generated templates rather than rendering the template anew each time, as the `send...` modules do currently.


src/server/routes/index.ts
```diff
@@ -64,4 +65,7 @@ if (okta.enabled) {
 
 router.use(uncachedRoutes);
 
+// email template routes
+uncachedRoutes.use(emailTemplates);
```
dominickendrick, line 69: Should this go above `router.use(uncachedRoute)` ?

raphaelkabo: Oh yes. 👍

src/server/routes/emailTemplates.ts
```diff
@@ -0,0 +1,52 @@
...
+import { ResetPasswordText } from '../../email/templates/ResetPassword/ResetPasswordText';
+import { Verify } from '../../email/templates/Verify/Verify';
+import { VerifyText } from '../../email/templates/Verify/VerifyText';
+
+const router = Router();
+
+const emailTemplateTypes = ['ResetPassword', 'Verify'] as const;
+type EMAIL_TEMPLATE_TYPE = typeof emailTemplateTypes[number];
+
+type EMAIL_RENDER_RESULT = {
```
dominickendrick, line 15: I think we only use ALL_CAPS for string constants, rather than types. I would suggest `EmailRenderResult` and `EmailTemplateType` naming for the above types

src/server/routes/emailTemplates.ts
```diff
@@ -0,0 +1,52 @@
...
+  plain: string;
+  html: string;
+};
+
+const renderEmailTemplate = (template: EMAIL_TEMPLATE_TYPE) => {
```
dominickendrick, line 20: Can we add the return type `EmailRenderResult | undefined`

src/server/routes/emailTemplates.ts
```diff
@@ -0,0 +1,52 @@
...
+import { render } from 'mjml-react';
+
+import { ResetPassword } from '../../email/templates/ResetPassword/ResetPassword';
+import { ResetPasswordText } from '../../email/templates/ResetPassword/ResetPasswordText';
+import { Verify } from '../../email/templates/Verify/Verify';
+import { VerifyText } from '../../email/templates/Verify/VerifyText';
+
+const router = Router();
+
+const emailTemplateTypes = ['ResetPassword', 'Verify'] as const;
```
dominickendrick, line 12: urls should be considered case-sensitive 
```URLs in general are case-sensitive (with the exception of machine names). There may be URLs, or parts of URLs, where case doesn't matter, but identifying these may not be easy. Users should always consider that URLs are case-sensitive```
https://www.w3.org/TR/WD-html40-970708/htmlweb.html

I would make these types lower case if they are also being used to match on the inbound url.

**Comment by raphaelkabo** [comment]
Quick question, @dominickendrick: we pass Okta this 'accidentalEmail' template to avoid emailing errors. AFAIK the accidentalEmail template only exists in IDAPI as pure HTML - there isn't an MJML template for it in Gateway. Should I create one, and then add `accidentalEmail` to this new `/email/...` route, too?

**Comment by raphaelkabo** [comment]
Accidental email template in Storybook:
Screenshot 2022-03-29 at 15 53 50

File: src/server/routes/emailTemplates.ts [hunk 5]
```diff
@@ -0,0 +1,65 @@
...
+};
+
+const renderEmailTemplate = (
+  template: EmailTemplateType,
+): EmailRenderResult | undefined => {
+  switch (template) {
+    case 'reset-password':
+      return {
+        plain: ResetPasswordText(),
+        html: render(ResetPassword()).html,
```
coldlink, line 33: rather than rendering the HTML on each call to the route, an optimisation would be render the HTML earlier and return the html string. Calling the render method on each request is expensive, and a route to a possible DoS attack.

The HTML is static too, so doesn't need to be rendered over and over again. 

I suggest either:
a) render the html at app startup and return the HTML string for that email
b) memoize the output of the render method

src/server/routes/emailTemplates.ts
```diff
@@ -0,0 +1,65 @@
...
+ * Returns 406 for invalid template names. */
+router.get(
+  '/email/:template',
+  (req: Request, res: ResponseWithRequestState) => {
+    const template = req.params.template as EmailTemplateType;
+    const templateIsValid = emailTemplateTypes.includes(template);
+
+    return templateIsValid
+      ? res.json(renderEmailTemplate(template))
+      : res.send(406); // 406: Not Acceptable
```
coldlink, line 61: I'd suggest `400` over `406` here.

As according to [MDN](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/406) `406`  client error response code indicates that the server cannot produce a response matching the list of acceptable values defined in the request's proactive [content negotiation](https://developer.mozilla.org/en-US/docs/Web/HTTP/Content_negotiation) headers. In this case we're not relying on those headers, but rather malformed client request, which is what [`400`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/400) is used for.

However I think `404` makes the most sense, if an template doesn't exist, then it should be a "Not Found" case is for, as the request itself isn't malformed in any obvious way.

I also suggest [`res.sendStatus`](https://expressjs.com/en/4x/api.html#res.sendStatus) or [`res.status`](https://expressjs.com/en/4x/api.html#res.status) if you just need to return a status code, as (i think) `res.send(406)` will literally just send 406 in the body, with an okay response, but I'm not sure about that.

```suggestion
      : res.sendStatus(404);
```

src/server/routes/index.ts
```diff
@@ -62,6 +63,9 @@ if (okta.enabled) {
   uncachedRoutes.use(oauth);
 }
 
+// email template routes
+uncachedRoutes.use(emailTemplates);
```
coldlink, line 67: As mentioned previously, these emails don't need to be uncached as they're all static html. So you can just `router.use`

```suggestion
router.use(emailTemplates);
```

coldlink (changes requested):

src/server/routes/emailTemplates.ts
raphaelkabo: This has now been done by rendering the HTML at app startup!

src/server/routes/emailTemplates.ts
```diff
@@ -0,0 +1,72 @@
...
+      // We don't want to do anything for invalid template names
+      return undefined;
+  }
+};
+
+/* GET a valid email template name.
+ * Returns a JSON object with the fields 'plain' and 'html' which contain the
+ * plain-text and HTML versions of the requested email template.
+ * Returns 404 for invalid template names. */
+router.get(
```
ob6160, line 60: can we add the rate limiter middleware to protect this endpoint please

cypress/integration/ete/email/template.spec.ts
```diff
@@ -0,0 +1,44 @@
+describe('Email template generation', () => {
```
ob6160, line 1: just noticed that some more email tpls were added to the switch-case below, do they need a test in-place as well?

raphaelkabo: Haha, d'oh!

ob6160 (approved):

PR #1536, opened 2022-03-03
Sign out with Okta

coldlink:
## What does this change?
Adds Okta sign out functionality. When a user visits the `/signout` url, they will now also have:

1.  Their Okta sessions invalidated
2.  Their Okta `sid` cookie removed

This is in addition to current functionality from https://github.com/guardian/gateway/pull/1476

This PR 

- checks the users current session (via the `sid` cookie) against the okta sessions API (https://developer.okta.com/docs/reference/api/sessions/#get-session)
- removes the users current Okta session via the users API (https://developer.okta.com/docs/reference/api/users/#clear-user-sessions)
- refactors some wrapper functions to okta api calls to take simpler types.

## How to test

This has been tested manually in code.
And mocked tests have been added

Automated end to end tests are awaiting changes to CI so will follow in another PR

PR #1601, opened 2022-03-29
Refactor sign in to use Okta session cookie check

dominickendrick:
## What does this change?

This PR changes the way we check for existing sessions when a user visits the signin page.

If a user already has an Okta session, and tries to sign in with another account, Okta will refresh the old session, rather than adding the new one. For this reason we want to check for an existing session before we try and sign the user in.

We previously did not have access to the `sid` Okta session cookie as it was set on `auth.theguardian.com` domain name, and gateway is on the `profile.theguardian.com` domain. Now we have switch the domain of the cookie to be `profile.theguardian.com` we can read the session id from the cookie, and check if the session is valid via the Okta sessions API [`/api/v1/sessions/:sessionId`](https://developer.okta.com/docs/reference/api/sessions/#get-session) via the `getSession` method added in https://github.com/guardian/gateway/pull/1536

The changes in the PR:

- Adds the session check via the `getSession` call.
- Keeps the functionality to redirects the user to the `returnUrl` parameter if they have an existing session. Note that `redirectUrl` is parsed so only allowed values can be used!
- Moves the sign in mock tests to their own file
- Remove the redirect from `/signin` that sent the user through the authorization code flow to check their session.
- Removes the `signInRedirect` parameter from the `encryptedStateCookie`.
- Updates the documentation to reflect the new flow.                  

## How to test

- Sign in on `profile.theguardian.com/signin?useOkta=true`
- After successful sign in revist `profile.theguardian.com/signin?useOkta=true`
- You should be redirected to `theguardian.com/uk` and your current session should still exist
- You should not have been prompted to sign in


cypress/integration/mocked/okta_sign_in.spec.ts
```diff
@@ -0,0 +1,175 @@
...
+
+      // we can't actually check the authorization code flow
+      // so intercept the request and redirect to the guardian about page
+
+      cy.visit(
+        '/signin?returnUrl=https%3A%2F%2Fwww.theguardian.com%2Fabout&useOkta=true',
+      );
+
+      Cypress.on('uncaught:exception', (err, runnable) => {
+        return false;
```
liamness, line 42: Do we mind warnings? If we do, maybe we should run eslint on the CI with `--max-warnings=0`.

cypress/integration/mocked/okta_sign_in.spec.ts
```diff
@@ -0,0 +1,175 @@
...
+        },
+      ).as('authRedirect');
+
+      cy.get('[data-cy=sign-in-button]').click();
+      cy.waitFor('@authRedirect').then(() => {
+        cy.url().contains('https://theguardian.com/about');
+      });
+    });
+  });
+});
```
liamness, line 175: Note to self as much as anything, might make sense to having a equivalent to the tests that start with

```js
context('Redirect to post sign-in prompt', () => {
```

in the IDAPI sign in tests, now these aren't skipped.

liamness (commented):
Almost approving but I think we should make a decision on the warnings first.

cypress/integration/mocked/okta_sign_in.spec.ts
dominickendrick: yeah this was actually to catch exceptions when visiting `theguardian.com`. I'd be happy to discuss being stricter on warnings, but maybe outside this PR ?

liamness: I think the only action needed on this PR is to delete the unused vars `err` and `runnable`. There are no warnings from the linter output currently.

dominickendrick: ah sorry, I see : )

liamness (approved):

PR #1602, opened 2022-03-29
Updating url validator to also expect a hostname for app redirects

dominickendrick:
## What does this change?

This PR updates the returnUrl config for apps. 

Currently android require both `scheme` and `host` attributes when you handle a `deeplink` in android 

```
android:path
android:pathPrefix
android:pathPattern

These attributes are meaningful only if the `scheme` and `host` attributes are also specified for the filter.
```
https://developer.android.com/guide/topics/manifest/data-element#path

For this reason we are adding a hostname into the returnUrl for apps, and have chosen `profile.theguardian.com`.
This means the new returnUrl will be:

`com.guardian://profile.theguardian.com/authentication/callback`

The changes made were:

- Changed the android package name to com.guardian
- Added a check for the hostname
- Returned the `profile.theguardian.com` hostname when building the android returnUrl
- Added a guard for invalid protocol as only `https` 

## How to test
Confirmed this works whilst pairing with @mohammad-haque in CODE

### DIFF
--- a/src/server/lib/__tests__/validateUrl.test.ts
+++ b/src/server/lib/__tests__/validateUrl.test.ts
@@ -40,7 +40,8 @@ describe('validateReturnUrl', () => {
   });
 
   test('it should handle an native app redirect returnUrl - android live', () => {
-    const input = 'com.theguardian:/authentication/callback';
+    const input =
+      'com.guardian://profile.theguardian.com/authentication/callback';
 
     const output = validateReturnUrl(input);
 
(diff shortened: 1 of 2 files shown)


ob6160 (approved):

coldlink:
Just my two cents. Using `profile.theguardian.com` as the hostname doesn't make sense to me.

It reads to me as if we're redirecting back to our own domain rather than the application itself.

PR #1620, opened 2022-04-05
Remove .ratelimit.json

ob6160:
## What does this change?
This was accidentally tracked in my original PR #1571 — Rate limiting isn't turned on yet and we will use a different set of configuration values when we deploy it fully so this hasn't leaked any sensitive information.


raphaelkabo (approved):

jest.config.js
```diff
@@ -13,6 +13,7 @@ module.exports = {
     '<rootDir>/node_modules/',
     'utils',
   ],
+  testTimeout: 15000,
```
liamness, line 16: We still don't want this, right?

ob6160: yep, just a temporary solution until I can work out why they were timing out in GH Actions occasionally

PR #1621, opened 2022-04-05
Retry aws cp of rate limiter configuration

ob6160:
## What does this change?
Retries the `aws cp` of the rate limiter configuration as the copy was failing on instance startup sometimes.

### DIFF
--- a/cloudformation.yaml
+++ b/cloudformation.yaml
@@ -193,7 +193,14 @@ Resources:
             unzip -o /etc/gu/${App}.zip -d /etc/gu
 
             # Get Rate Limiter configuration file
-            aws s3 cp s3://identity-private-config/${Stage}/identity-gateway/.ratelimit.json /etc/gu/.ratelimit.json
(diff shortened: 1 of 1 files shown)


cloudformation.yaml
```diff
@@ -193,7 +193,14 @@ Resources:
...
             # Get Rate Limiter configuration file
-            aws s3 cp s3://identity-private-config/${Stage}/identity-gateway/.ratelimit.json /etc/gu/.ratelimit.json
+            # Try multiple times to the config file. The s3 cp command can fail when called immediately on instance startup.
+            while true; do
+              if \
+                aws s3 cp s3://identity-private-config/${Stage}/identity-gateway/.ratelimit.json /etc/gu/.ratelimit.json
+                then break
+              fi
+              sleep 1
+            done
```
liamness, line 203: I feel this should probably be limited in how many times it will try to fetch the file, but I can see this pattern is already used below. Could loop back and look into that, not a blocker considering you're troubleshooting.

liamness (approved):

PR #1623, opened 2022-04-06
Fix: Sign out issue

coldlink:
## What does this change?

There is an accidental `,` here, which should be a `;`, this may cause issues with signout. So changing this to fix that.


raphaelkabo (approved):

ob6160 (approved):
is there some kind of linter rule we can use for this kind of thing?

coldlink:
> is there some kind of linter rule we can use for this kind of thing?

yeah i'll look into it, i only noticed this when i was working out why i couldn't get sign out tests to work

coldlink:
https://eslint.org/docs/rules/no-sequences we'll use this

PR #1626, opened 2022-04-06
WIP Refactor how rate limiter middleware is applied

ob6160:
## Accessibility


src/server/lib/__tests__/rate-limit/rateLimitMiddleware.test.ts
```diff
@@ -114,7 +114,7 @@ describe('rate limiter middleware', () => {
   };
 
   it('should rate limit all clients when the global bucket is empty', async () => {
-    const rateLimiterConfig = {
+    const rateLimiterConfig: RateLimiterConfiguration = {
```
liamness, line 117: I'm wondering why you're using the normal type annotation, and also `as Type` here. I feel like one or the other would work here. Or maybe even just using neither any relying of structural typing to catch any issues would be fine?

liamness (approved):
Looks good, just one super minor comment / question which imo doesn't need to block release.

src/server/lib/__tests__/rate-limit/rateLimitMiddleware.test.ts
ob6160: Ah I missed this :) I'll switch to use one as it was just an oversight. I added the type to these so they're easier to modify with the type hints

PR #1630, opened 2022-04-07
Remove minification on MJML renderer

raphaelkabo:
## What does this change?

The standard `mjml-react` `render()` function gives us the following error when run:

```
"minify" option is deprecated in mjml-core and only available in mjml cli.
```

As per [this issue](https://github.com/wix-incubator/mjml-react/issues/58), the simple fix is to disable minification, which is what is done in this PR. There appears to be a PR in the works, but it's been blocked due to some disagreement between the PR author and the package maintainers. In the meantime, if need be, we can minify the HTML output ourselves with our minifier of choice.


ob6160 (approved):
looks good!

PR #1632, opened 2022-04-08
Rate limiter debug functionality

ob6160:
## What does this change?

This introduces new debug and monitoring features to help us to see how our configuration changes are affecting users of the website as we roll this out in production.

* Logging every 5 seconds of the current global bucket capacity for all routes
  * This is enabled by setting `trackBucketCapacity` to `true`.
* Refactors general configuration into an optional extra object called `settings`
  * Adds a new `logOnly` setting. If it is set to `true`, we will run the rate limiter but never return a HTTP 429 (rate limited exceeded) response; even if the user has reached their limit.
    * This will let us trial the rate limiter in production and iron out any problems in our configuration.
* Improved logging for when the rate limit is hit so that we can more accurately monitor whether we are rate limiting clients that we shouldn't.
* Renamed the rate limiter metric so that we know to be Gateway specific
* Extend the server side logger to allow for extra parameters to be passed in. This is so that we can log the bucket capacity remaining in a way that we can retrieve easily in Kibana

### DIFF
--- a/src/server/lib/__tests__/rate-limit/rateLimitMiddleware.test.ts
+++ b/src/server/lib/__tests__/rate-limit/rateLimitMiddleware.test.ts
@@ -116,6 +116,7 @@ describe('rate limiter middleware', () => {
   it('should rate limit all clients when the global bucket is empty', async () => {
     const rateLimiterConfig: RateLimiterConfiguration = {
       enabled: true,
+      debug: false,
       defaultBuckets: {
         globalBucket: { capacity: 2, addTokenMs: 500 },
       },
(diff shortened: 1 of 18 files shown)


ob6160:
[Add debug features to rate limiter](https://trello.com/c/nueakoNB/3899-add-debug-features-to-rate-limiter)

src/server/lib/__tests__/rate-limit/rateLimitMiddleware.test.ts
```diff
@@ -242,9 +254,43 @@ describe('rate limiter middleware', () => {
...
+    try {
+      // Confirm that we are not rate limited when logOnly is set to true.
+      await request(server).get('/signin').expect(200);
+      await request(server).get('/signin').expect(200);
+
+      await teardownServer();
+    } catch (error) {
+      await teardownServer();
+      throw error;
+    }
```
liamness, line 284: Could use a `finally` block for the teardown.

liamness (approved):
Looks good to me, just one super minor suggestion

src/server/lib/__tests__/rate-limit/rateLimitMiddleware.test.ts
ob6160: great idea, I'll action this 🔨

PR #1633, opened 2022-04-11
Add `componentEventParam` tracking through query params

coldlink:
## What does this change?

During the migration of the sign in and register pages from [identity-frontend](https://github.com/guardian/identity-frontend) to gateway, the tracking of events on frontend and DCR were missed from the migration, specifically using component event params to track how users would sign in and register from those systems.

This commit readds the `componentEventParams` query parameter, and makes sure that it is passed through to identity-api through the sign in and register endpoints.

We also update some types and function calls to clean up the tracking functionality too.


ob6160 (approved):
looks good -nice work

PR #1644, opened 2022-04-12
April 12th 2022 | Dependency upgrade

coldlink:
## What does this change?

* Brings our dependencies up to the latest versions.
* Keeps `ioredis` at v4 because `ioredis-mock` has not been upgraded to support ioredis v5 yet.
* Keeps `@storybook/*` packages pinned at alpha 34 until the full `6.5.0` release

Also fixes issues with onboarding flow test checkboxes.

## Tested

- [x] CODE
- [x] IE11


ob6160 (approved):
the list is growing 😅

PR #1651, opened 2022-04-13
Add .well-known static directory

coldlink:
## What does this change?

Adds a new static directory (`.well-known`) to the project to store Android and iOS validation files on the `profile` subdomain.

Currently:
`apple-app-association-file` (for iOS)
`assetlinks.json` (for Android)

## Other Changes

https://github.com/guardian/identity-platform/pull/513

### DIFF
--- a/src/server/routes/core.ts
+++ b/src/server/routes/core.ts
@@ -16,6 +16,16 @@ router.use(
   }),
 );
 
+router.use(
+  '/.well-known',
+  express.static(path.resolve(__dirname, '.well-known'), {
+    cacheControl: true,
+    // setting to 30 mins for development, so we don't have to invalidate
+    // the cache on every update to a .well-known file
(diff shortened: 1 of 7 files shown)


src/server/routes/core.ts
```diff
@@ -16,6 +16,14 @@ router.use(
   }),
 );
 
+router.use(
+  '/.well-known',
+  express.static(path.resolve(__dirname, '.well-known'), {
+    cacheControl: true,
+    maxAge: ms('1y'),
```
dominickendrick, line 23: hmm, what if they needed to make changes to this file ? It's probably worth setting this to 30 mins so we don't need to flush the cache in fastly if there is a prod change

coldlink: updated it to 30m and left a comment about this

dominickendrick (approved):

PR #1653, opened 2022-04-13
Persist query params in encryptedStateCookie for email based flows

dominickendrick:
## What does this change?

This PR adds the functionality to include `PersistableQueryParams: returnUrl, useOkta, and clientId` in `encryptedStateCookie` for the registration and password set/reset flow in Okta. 

We cannot pass these values through the Okta email system, due to Okta system limitations of us not being able to customise the url link sent in emails.

These parameters are added to the `encryptedStateCookie` as part of the registration flow, set and reset password flows.

We only remove these parameters from the `encryptedStateCookie` once a user has done a successful action (ie set their password), in case they reload the page etc.

## Testing
We have added new end to end tests for Okta to test this.
The apps teams will also test this in the `CODE` environment.

Scenarios:

- [x] - Register a new user with a returnUrl in Okta
- [x] - Register a new user without a returnUrl in Okta
- [x] - Register a new user with a returnUrl in Okta, and we change the url in the email link before visiting (to test what happens if the cookie and url return Url don't match - the url parameter should take precedence)


src/server/controllers/checkPasswordToken.ts
```diff
@@ -100,6 +106,34 @@ const checkTokenInIDAPI = async (
...
+  encryptedStateQueryParams: PersistableQueryParams,
+): string => {
+  // check that the returnUrl in requestState is not the defaultReturnUri
+  // as this suggests that it would have been modified, such as the native apps
+  // setting the return url on email link intercept
+  if (requestState.queryParams.returnUrl !== defaultReturnUri) {
+    return requestState.queryParams.returnUrl;
+  }
+  // otherwise check the encrypted state cookie for a returnUrl
+  // We always want to validate the returl url value, just in case its been incorrectly set though developer error
```
raphaelkabo, line 129:
```suggestion
  // We always want to validate the returl url value, just in case it's been incorrectly set through developer error
```

src/server/controllers/checkPasswordToken.ts
```diff
@@ -118,15 +152,34 @@ export const checkTokenInOkta = async (
...
 
-    setEncryptedStateCookie(res, { email, stateToken });
+    updateEncryptedStateCookie(req, res, { email, stateToken });
 
     trackMetric('OktaValidatePasswordToken::Success');
 
+    // since we can't pass query parameters through okta emails, we set the encryptedStateCookie
+    // as the email was sent to the user containing the query params at that time
+    // so we read them here, if the user comes back on the same browser
+    const encryptedState = readEncryptedStateCookie(req);
```
raphaelkabo, line 162: Are these functions only ever going to be used for this specific cookie `GU_GATEWAY_STATE`? Are we using the cookie more generally, or only for this specific flow, during registration?

coldlink: The encrypted state cookie (`GU_GATEWAY_STATE`) is already being used is multiple flows, not just for this use case. We thought we could extend the functionality for this use case.

This particular use case it for okta registration and okta password reset methods.

coldlink (approved):

PR #1654, opened 2022-04-14
Remove `copy-node-modules`

coldlink:
## What does this change?

[`copy-node-modules`](https://github.com/arloliu/copy-node-modules) doesn't seem to be supported anymore, so removing this dependency, and adding functionality in a different way. 

## Tested

- [x] CODE


ob6160 (approved):
nice nice nice

PR #1655, opened 2022-04-14
Actually remove `copy-node-modules` dependency

coldlink:
## What does this change?

Removed the usage of `copy-node-modules` in https://github.com/guardian/gateway/pull/1654 but forgot to remove the dependency itself. 🙈


ob6160 (approved):
:shipit:

PR #1656, opened 2022-04-14
Add timeout to Cypress Github Actions

coldlink:
## What does this change?

- Adds a timeout to the Cypress GitHub Actions to fail after 15 mins rather than the default 6 hours
  - While this doesn't solve flakiness, it should help to spot tests that are stuck
  - On success, most tests take less than 10 mins, so adding 5 mins for leeway
- Reduce `numTestsKeptInMemory` to `1` to help flakiness possibly
  - By reducing the number of tests in memory to 1, cypress only keeps 1 snapshot of tests in it's memory, this is useful for CI as we don't/can't inspect previous snapshots, and we just want to know the result.
  - This has the added benefit of reducing memory usage, which might reduce flakiness in cypress tests running in GitHub actions, especially those that seem to hang/freeze rather than fail.

### DIFF
--- a/.github/workflows/cypress-ete-okta.yml
+++ b/.github/workflows/cypress-ete-okta.yml
@@ -94,7 +94,7 @@ jobs:
           wait-on-timeout: 60
           browser: chrome
           spec: cypress/integration/ete-okta/**/*.ts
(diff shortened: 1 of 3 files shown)


coldlink:
image

it works 😛

liamness (approved):
fail fast, very agile of you

PR #1657, opened 2022-04-14
Rate limiter tweaks

ob6160:
## What does this change?

* Truncates the access token that we log in the rate limiter middleware when the limit is hit. This is so that we don't exceed the maximum message size in Kibana
* Checks that the global bucket key array isn't empty.
* Extends middleware test to check that the correct rate limit exceeded messages are being logged.
* Removes some unnecessary newlines

### DIFF
--- a/src/server/lib/middleware/rateLimit.ts
+++ b/src/server/lib/middleware/rateLimit.ts
@@ -66,9 +66,10 @@ export const rateLimiterMiddleware = async (
 
   if (ratelimitBucketTypeIfHit) {
     // Cut down the size of the access token so we don't make the log line too long.
-    const truncatedAccessToken = rateLimitData.accessToken
-      ? rateLimitData.accessToken.substring(0, 6)
-      : '';
+    const truncatedAccessToken =
+      typeof rateLimitData.accessToken === 'string'
+        ? rateLimitData.accessToken.substring(0, 6)
+        : '';
 
     // We append `-Gateway` so that we can differentiate between IDAPI rate limit log entries.
     logger.info(
--- a/src/server/lib/middleware/rateLimit.ts
+++ b/src/server/lib/middleware/rateLimit.ts
@@ -65,6 +65,7 @@ export const rateLimiterMiddleware = async (
   });
 
   if (ratelimitBucketTypeIfHit) {
+    // Cut down the size of the access token so we don't make the log line too long.
     const truncatedAccessToken = rateLimitData.accessToken
       ? rateLimitData.accessToken.substring(0, 6)
       : '';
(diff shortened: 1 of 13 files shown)


src/server/lib/middleware/rateLimit.ts
```diff
@@ -65,9 +65,14 @@ export const rateLimiterMiddleware = async (
   });
 
   if (ratelimitBucketTypeIfHit) {
+    // Cut down the size of the access token so we don't make the log line too long.
+    const truncatedAccessToken = rateLimitData.accessToken
+      ? rateLimitData.accessToken.substring(0, 6)
+      : '';
```
liamness, line 71: Could do `rateLimitData.accessToken?.substring(0, 6)` here maybe

src/server/routes/verifyEmail.ts
```diff
@@ -146,7 +144,6 @@ router.post(
 
 router.get(
   '/verify-email/:token',
-
```
liamness, line 149: Do we know what was up with this formatting before? Some kind of mass find + replace operation maybe?

I see that `yarn test` doesn't run prettier. Was that why this wasn't caught?

liamness (approved):
Looks good to me, only minor comments.

src/server/routes/verifyEmail.ts
ob6160: I think it was from when I moved from having the `rateLimitMiddleware` included at this level to moving it up to the router so we didn't have to add it to every single route. Odd that the linter didn' catch it

src/server/lib/middleware/rateLimit.ts
ob6160: good idea!

src/server/routes/verifyEmail.ts
liamness: I think only ESLint runs on the CI, not Prettier.

src/server/lib/middleware/rateLimit.ts
```diff
@@ -66,9 +66,10 @@ export const rateLimiterMiddleware = async (
 
   if (ratelimitBucketTypeIfHit) {
     // Cut down the size of the access token so we don't make the log line too long.
-    const truncatedAccessToken = rateLimitData.accessToken
-      ? rateLimitData.accessToken.substring(0, 6)
-      : '';
+    const truncatedAccessToken =
+      typeof rateLimitData.accessToken === 'string'
+        ? rateLimitData.accessToken?.substring(0, 6)
```
liamness, line 71: There won't be any need for the optional chaining syntax here as the type is known.

src/server/routes/verifyEmail.ts
ob6160: 👍

src/server/lib/middleware/rateLimit.ts
ob6160: 👍

PR #1670, opened 2022-04-22
Disable reCAPTCHA middleware on DEV stage

raphaelkabo:
## What does this change?

When running the handleRecaptcha middleware locally, we consistently saw a bug where the route handler would be called multiple times, causing 'Cannot set headers after they are sent to the client' errors and running everything in the handler multiple times. Our guesses for why this happens are:

1. The testing keys for reCAPTCHA are somehow broken;
2. `express-recaptcha` is somehow broken; 
3. The way we integrate it is somehow broken.

To solve it, I've set it up so we automatically pass `next()` instead of running the middleware in the `DEV` stage. The middleware still runs as before in other stages, as the bug doesn't seem to happen in production.

It would be best if at some point we actually got to the bottom of why the middleware was triggering the route handler multiple times, and fixed the bug at source.

### DIFF
--- a/src/server/lib/recaptcha.ts
+++ b/src/server/lib/recaptcha.ts
@@ -5,6 +5,7 @@ import { getConfiguration } from '@/server/lib/getConfiguration';
 import { logger } from '@/server/lib/serverSideLogger';
 import { trackMetric } from '@/server/lib/trackMetric';
 import { HttpError } from '@/server/models/Error';
+import { featureSwitches } from '@/shared/lib/featureSwitches';
 
 const {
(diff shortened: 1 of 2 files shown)


ob6160 (commented):
Looks good -just dropping a quick thought - maybe this could be a separate entry in the configuration object set using a `RECAPTCHA_ENABLED` env var (defaulted to on)? 

Suggesting because as-is I wouldn't know it's disabled on dev without opening the file and reading the comment.

coldlink (approved):

PR #1682, opened 2022-04-28
Social Sign In and Registration with Okta

coldlink:
## What does this change?

Adds the Okta Social Identity Providers to Gateway, allowing for the sign in and registration of social users directly within Okta.

### Old Identity Land

In our old Identity world, social sign-in for web is handled by Federation API and Identity API. Federation kicks off an OAuth flow with the social provider and then calls IDAPI to retrieve session cookies and determine whether the user is blocked from social sign-in.
Social sign-in for mobile does not use Federation API and is handled solely by the apps and Identity API. The mobile apps complete OAuth flows with the social providers directly (presumably using the providers’ native SDKs) and send either the access token or the ID token to IDAPI. IDAPI checks the token validity with the social provider, checks if the user is blocked from social sign-in and returns either an error or an IDAPI user access token.

### Okta

Social sign in with Okta works in a similar way, except we use the Okta OAuth/OIDC flows with the `idp` parameter on the `/authorize` endpoint.
This first kicks an OIDC flow with Okta from Gateway (the same as how it currently works for sign in with email and password), then Okta reads the `idp` parameter, determines the social provider, and starts an OIDC flow with that provider.
The user then authenticates (if required) and authorizes the application on the social provider, the social provider does a callback to Okta authorizing the user. Okta will then check the state of the user, e.g. if they're a new or existing user, and take the appropriate action. Okta will then authenticate the user and set a session cookie in that users browser.

Once the user is authenticated with the session cookies, Okta will complete the authorization code flow that was started in Gateway, and does a callback to the `/oauth/authorization-code/callback` endpoint, the same as how it works for email + password. We then exchange the auth code for the access tokens, get the old identity cookies if required, and redirect the user back to the return url or the native app.

#### Existing users

For existing users in Okta, we automatically link their account. We also restrict automatic account linking to users within the `GuardianUser-EmailValidated` group to mitigate account squatting, as we know that users in the email validated group have validated they own the account through access in their email inbox.

By restricting account linking, the user is blocked from authentication, and we receive a generic ‘User linking was denied because the user is not in any of the specified groups’ error without any user details. In this case we use this error to display a custom error message on the sign-in page but we won’t be able to auto-fill the email address because we won’t know what it is. This is roughly equivalent to what we already have with a slightly degraded user experience as the email is not pre-populated. It does not perform any manual account linking.

#### New users

If a user doesn't exist and tries to sign-in with social, we create them automatically using JIT (just in time) provisioning which creates a new user record. When a user is created by JIT, we can automatically assign them to a group, in our case we assign them to the `GuardianUser-All` and `GuardianUser-EmailValidated` groups, which match similarly to what we have in old Identity land.

We're unable to set the `user.emailValidated` field in the Okta user profile for social users when they are created, but we are able to put them in the `GuardianUser-EmailValidated` group.

So this is a workaround to set the `emailValidated` field to true in the Okta user profile if the user is in the `GuardianUser-EmailValidated` group, based on custom claims we have set up on the `id_token`. Specifically the `user_groups` and `email_validated` claim. The `user_groups` claim returns an array of groups that the user is in, and the `email_validated` flag returns the status of the `user.emailValidated` profile field. If the user is in the `GuardianUser-EmailValidated` group, but the `emailValidated` field is falsy, then we set the `emailValidated` field to `true` in the Okta user profile by manually updating the user object.

A new user that registers with a social account becomes `ACTIVE` in Okta, with the `SOCIAL` provider and no password set. This creates some issues regarding reset password.

When a `SOCIAL` user tries to reset their password using a reset/forgot password form, the request will fail with an error mentioning they do not have permissions to take that action. To be able to send a password reset email to the user we have to get them from the `SOCIAL` provider to the `OKTA` provider with the `ACTIVE` state and a placeholder/temporary password set.

#### Reset password for new users
This takes a few steps to do to get a user into the correct state with a placeholder password set.

Firstly we listen out for the error when a user attempts to reset their password [[1](https://github.com/guardian/gateway/blob/585b4d0f7b65e2acec156a2acd48b4a1cad14d55/src/server/controllers/sendChangePasswordEmail.ts#L109-L111)].
When we see this error we run a few checks to make sure the user is in the particular state without a password set. We first get the `user` object by using their email. This returns a `credentials` object. When a user does not have a password set the `user.credentials.password`  object will not exist, so we can use this to perform logic.
When a password isn't set, we HAVE to use the users api reset password endpoint ([method](https://github.com/guardian/gateway/blob/585b4d0f7b65e2acec156a2acd48b4a1cad14d55/src/server/lib/okta/api/users.ts#L177)/[api doc](https://developer.okta.com/docs/reference/api/users/#reset-password)), this forces the user from the `SOCIAL` provider to the `OKTA` provider, and into the `RECOVERY` state. No other endpoint does this, so we have to use this endpoint.
This returns a one time recovery token, which we don't send to the user. We use this token to get the authentication `stateToken`.
We then use this `stateToken` to set a randomly generated unreadable password, and this forces the user into the `ACTIVE` state with a password set.
Now the user is in this state we can send a user an email to be able to reset their password.

## Changes
- [x] Add environment variables for the Okta Social IDPs
- [x] Updates the Authorization Code flow to allow for the use of the Okta `idp` flag
- [x] Adds a `/signin/:social` route to handle social sign in for both Identity Federation API and Okta
- [x] Refactor social buttons to use this new route
- [x] Set placeholder passwords for social users that request to reset their password

## Checks
- New users
  - [x] For new users (no existing accounts) we should create a new social account in the active state and add them to the required groups (`GuardianUser-All`, `GuardianUser-EmailValidated`)
- Existing users
  - [x] Restrict account linking to the specified group (`GuardianUser-EmailValidated`)
  - [x] Return error for users outside this group

### New Account created in `GuardianUser-All`, `GuardianUser-EmailValidated` groups
- [x] New user registers with Facebook
- [x] New user registers with Google
- [x] New user registers with Apple

### Account linked automatically, users in `GuardianUser-EmailValidated` group
- [x] Existing user with password sign in with Facebook
- [x] Existing user with password sign in with Google
- [x] Existing user with password sign in with Apple

### Account linking error shown, users without `GuardianUser-EmailValidated` group
- [x] Existing user without password sign in with Facebook
- [x] Existing user without password sign in with Google
- [x] Existing user without password sign in with Apple 

### New Account created in `GuardianUser-All`, `GuardianUser-EmailValidated` groups
- [x] New user registers with Facebook
  - ~Registration hook not setting email validated flag for new users~
  - Fixed in https://github.com/guardian/gateway/pull/1682/commits/503d4f7b1eb469ad2667b4a5e9dd19d414c78c09
- [x] New user registers with Google
- [x] New user registers with Apple

### Account linked automatically, users in `GuardianUser-EmailValidated` group
- [x] Existing user with email validated sign in with Facebook
- [x] Existing user with email validated sign in with Google
- [x] Existing user with email validated sign in with Apple

### Account linking error shown, users without `GuardianUser-EmailValidated` group
- [x] Existing user without email validated sign in with Facebook
- [x] Existing user without email validated sign in with Google
- [x] Existing user without email validated sign in with Apple 

## Related PR

https://github.com/guardian/identity-platform/pull/519


src/server/lib/okta/api/users.ts
```diff
@@ -174,44 +174,42 @@ export const reactivateUser = async (
...
+ * https://developer.okta.com/docs/reference/api/users/#reset-password
+ *
+ * @param id Okta user Id
+ * @returns Promise<string>
  */
-export const generateResetPasswordToken = async (
-  id: string,
-  sendEmail = false,
-): Promise<TokenResponse | void> => {
+export const resetPassword = async (id: string): Promise<string> => {
```
ob6160, line 203: could the method name reflect the fact that this is generally a dangerous method to use if you haven't read the included comment? something along the lines of: `resetPasswordAndDangerouslySetState` or something?

src/server/routes/oauth.ts
```diff
@@ -127,6 +149,28 @@ router.get(
...
+      // set up on the id_token.
+      if (tokenSet.id_token) {
+        // extracting the custom claims from the id_token and the sub (user id)
+        const { user_groups, email_validated, sub } =
+          tokenSet.claims() as CustomClaims;
+
+        // if the user is in the GuardianUser-EmailValidated group, but the emailValidated field is falsy
+        // then we set the emailValidated field to true in the Okta user profile by manually updating the user
+        if (
+          !email_validated &&
```
ob6160, line 166: are these variables snake case because of the shape of the Okta API response?

src/server/routes/signIn.ts
```diff
@@ -325,24 +356,31 @@ router.post(
...
+type SocialProvider = 'google' | 'facebook' | 'apple';
+
+router.get('/signin/:social', (req: Request, res: ResponseWithRequestState) => {
+  // todo can we use the login middleware here to stopped logged in users from accessing this endpoint?
+  const { useOkta, returnUrl } = res.locals.queryParams;
+  const socialIdp = req.params.social as SocialProvider;
+
+  // validate social is one of facebook, google, or apple
+  // if not, redirect to signin
+  if (!['facebook', 'google', 'apple'].includes(socialIdp)) {
```
ob6160, line 368: what d'you think about having a var called `isValidSocialNetwork` or similar? I think we could avoid the small two line comment above with a name like that

ob6160 (approved):
great work-approved! so cool to see this come to life in the demo today as well. left a few comments, but nothing blocking

src/server/lib/okta/api/users.ts
coldlink: changed the name to `dangerouslyResetPassword`

src/server/routes/signIn.ts
coldlink: created a `isValidSocialProvider` helper method

ob6160 (approved):

PR #1684, opened 2022-04-29
Run cypress tests in dev mode !

dominickendrick:
## What does this change?
This makes a change to our cypress test runner scripts for `mocked` and `ete` (end to end) tests.

We now support a development mode that allows you to make code changes to the server without having the restart cypress to check your changes against the tests.

This should significantly speed up local development when using cypress. 🐎 

This is achieved by starting the server in `watch` mode rather than the production mode.

We support 2 modes as we still want the tests to run in a production like environment in CI.

To run the cypress tests in dev mode, use the following commands:

`make cypress-mocked-dev`

`make cypress-ete-dev`

`make cypress-ete-okta-dev`

### DIFF
--- a/cypress-ete.sh
+++ b/cypress-ete.sh
@@ -1,6 +1,16 @@
 #!/bin/bash
 # Quickly fire up Cypress using the CI settings for interactive local usage.
 
+# Set DEV_MODE=true to run the mocked tests with the development server.
+# We default to false here.
+: "${DEV_MODE:=false}"
+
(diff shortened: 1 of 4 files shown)


liamness (approved):

coldlink:
Updating the testing documentation with this info would be useful! https://github.com/guardian/gateway/blob/main/docs/setup.md#testing

ob6160 (approved):

PR #1701, opened 2022-05-04
Jobs: GRS group consent flow & mandatory fields 

ob6160:
## What does this change?
Prior to completing the migration of Jobs users to Okta we first need to reinstate the mandatory first and last name fields and reimplement the Jobs terms and conditions acceptance flow (`/agree/GRS`) to retain the functionality provided by `identity-frontend`. This PR covers both of these requirements.

We need to record names as part of the registration process for Madgex. We've not been doing this since January. For  users who have signed up since January, the terms and conditions page will be shown to them again when they next sign in to Madgex. This page will only ask them for their name so that, over time, we ensure that we start to collect this information.

Other changes include:

* Field validation functionality for the `<EmailInput/>` field was abstracted into a new hook: `useInputValidityState`. This returns event handlers for: ` onBlur`, `onInput` and `onInvalid` that, when passed to an `<input/>`, will transition the state of the returned `inputFieldState` variable.
  * This hook means that the input validation logic can now shared between the email, first and last name input fields.
* A new `<NameInputField />` component provides a First/Last name input field with field validation.
* A new `<JobsLogo />` was added and included in the terms and conditions accept flow header under an optional prop flag: `useJobsHeader`

## Key user journeys covered by this PR:
_(note: these are all covered by the included cypress test suite)_

### User with Guardian account and First/Last name not set and with no account in Madgex (Jobs) 

* Given the user clicks "sign in"
* When they are redirected to the `/agree/GRS` page
* Then they see the page with Jobs terms + first/last name field
* Given they fill in their First/Last name
* When they click "Continue" 
* Then they are redirected to Jobs; their Madgex account is created; and their name is set.

### User with Guardian account and First/Last name set and with no account in Madgex (Jobs) 

* Given the user clicks "sign in"
* When they are redirected to the `/agree/GRS` page
* Then they see the Jobs terms + first/last name field prefilled with their existing name
* Given they edit their First/Last name
* When they click "Continue" 
* Then they are redirected to Jobs; their Madgex account is created; and their name is updated.

### User with Guardian and Madgex account and First/Last name not set

* Given the user clicks "sign in"
* When they are redirected to the `/agree/GRS` page
* Then they see just the first+last name field
* Given they set their first/last name
* When they click "Continue"
* Then they are redirected to Jobs and their name is updated.

### New user with no account at the Guardian signs in from Jobs

* Given the user clicks "sign in"
* When they are redirected to the Guardian `/signin` page
* And they click the "Register" link
* Then they have `clientId=jobs` param set and can see the Jobs terms and conditions
* Given they enter their email
* When they click the link sent to their inbox
* Then they should see the welcome flow with First/Last name fields and Jobs t&c's

## How was this tested?

Each user journey was manually tested against Gateway CODE and the Madgex Jobs CODE instance.

A cypress suite covers these journeys up to the point that users would be redirected to the Jobs site. We avoid the redirect so that we don't create a lot of burner Madgex accounts during our tests.

### Name input field
Screenshot 2022-05-18 at 16 54 16

### Name input field empty error behaviour
https://user-images.githubusercontent.com/1771189/169087637-c9cd59b5-f75b-44c5-a79d-ef85f574137f.mov


src/shared/model/Routes.ts
```diff
@@ -42,6 +42,7 @@ export const ValidRoutePathsArray = [
   '/welcome/email-sent',
   '/welcome/expired',
   '/welcome/resend',
+  '/agree/GRS',
```
coldlink, line 45: Does this need to be changed in identity-platform too, i,e point it to gateway instead of identity-frontend?

src/server/routes/agree.ts
```diff
@@ -0,0 +1,99 @@
...
+
+const { defaultReturnUri } = getConfiguration();
+
+router.get(
+  '/agree/GRS',
+  async (req: Request, res: ResponseWithRequestState) => {
+    const SC_GU_U = req.cookies.SC_GU_U;
+    const state = res.locals;
+    const { returnUrl } = state.pageData;
+    const { signInPageUrl } = getConfiguration();
```
coldlink, line 21: This should be in the global scope, i.e outside the router, you can add it on L13 where theres another getConfiguration call.

src/server/routes/agree.ts
```diff
@@ -0,0 +1,99 @@
...
+  },
+);
+
+router.post(
+  '/agree/GRS',
+  async (req: Request, res: ResponseWithRequestState) => {
+    const state = res.locals;
+    const { firstName, secondName } = req.body;
+    const { returnUrl } = state.pageData;
+    const redirectUrl = returnUrl || defaultReturnUri;
```
coldlink, line 86: if you get `returnUrl` from `state.queryParams` instead, it will always be defined, as that has a default set.

`pageData` is used for things to display on the page, which may not always be defined

```suggestion
    const { returnUrl } = state.queryParams;
    // use returnUrl instead of redirectUrl below
```

src/shared/model/ClientState.ts
```diff
@@ -41,6 +41,8 @@ export interface PageData {
   previousPage?: ConsentPath;
   // reset password token specific
   timeUntilTokenExpiry?: number;
+  firstName?: string;
+  secondName?: string;
```
coldlink, line 45: maybe a comment here explaining this is for jobs specifically e.g. 
```suggestion
  // jobs specific flow
  firstName?: string;
  secondName?: string;
```

src/server/routes/agree.ts
```diff
@@ -0,0 +1,99 @@
...
+import setupJobsUser from '../lib/jobs';
+
+const { defaultReturnUri } = getConfiguration();
+
+router.get(
+  '/agree/GRS',
+  async (req: Request, res: ResponseWithRequestState) => {
+    const SC_GU_U = req.cookies.SC_GU_U;
+    const state = res.locals;
+    const { returnUrl } = state.pageData;
```
coldlink, line 20: same as previous comment
```suggestion
    const { returnUrl } = state.queryParams;
```

src/server/lib/jobs.ts
```diff
@@ -0,0 +1,25 @@
...
+  //
+  // Once users belong to the GRS group, they aren't shown the /accept/GRS page when
+  // they try to sign in to the Jobs site for the first time.
+  //
+  // We can resolve both promises here because they are not dependent on each other.
+  return await Promise.all([
+    updateName(firstName, secondName, ip, sc_gu_u),
+    addToGroup(GroupCode.GRS, ip, sc_gu_u),
+  ]);
+};
```
coldlink, line 23: you don't actually need the `async/await` here, as where this method is already returning a promise, so is being used is already calling `await` at that level, there's other places in the project that has used this antipattern too, e.g. the okta api calls, so it's worth starting to address this at least here anyway

```suggestion
const setupJobsUser = (
  firstName: string,
  secondName: string,
  ip: string,
  sc_gu_u: string,
) => {
  if (firstName === '' || secondName === '') {
    throw new Error('Empty values not permitted for first or last name.');
  }
  // When a jobs user is registering, we add them to the GRS group and set their full name.
  // We also want to set their first and last name—these are required fields for all Jobs users.
  //
  // Once users belong to the GRS group, they aren't shown the /accept/GRS page when
  // they try to sign in to the Jobs site for the first time.
  //
  // We can resolve both promises here because they are not dependent on each other.
  return Promise.all([
    updateName(firstName, secondName, ip, sc_gu_u),
    addToGroup(GroupCode.GRS, ip, sc_gu_u),
  ]);
};
```

coldlink (commented):
Looking good! Just have a few suggestions.

src/shared/model/Routes.ts
ob6160: ah yes — I had it ready in a branch — raised the PR here: https://github.com/guardian/identity-platform/pull/529

src/server/lib/jobs.ts
ob6160: ah yes, good spot, much better this way

src/client/components/Terms.tsx
```diff
@@ -62,8 +62,27 @@ export const GuardianTerms = () => (
   </>
 );
 
+export const JobsStandaloneTerms = () => (
+  <div css={terms}>
+    <Text>
+      By proceeding, you agree to our{' '}
```
dominickendrick, line 68: maybe `{' '}` is the react way to insert a space, but you can also use `&nbsp;` html entity ...

coldlink (approved):

ob6160:
[Implement Jobs mandatory fields and Consent flow in Gateway](https://trello.com/c/T9wlY1aC/3880-implement-jobs-mandatory-fields-and-consent-flow-in-gateway)

PR #1702, opened 2022-05-09
Add `gu-cmp-disabled` cookie for redirect to manage/signin CODE

coldlink:
## What does this change?

Fixes an issue where the CMP banner would appear when redirecting to the manage and signin page on the CODE environment from Cypress.

This randomly started causing test failures in cypress with a `(uncaught exception)Error: no IAB consent framework found on the page` error.

Adding the ``gu-cmp-disabled=true` cookie on the redirect to the CODE environments fixes the issue as it stops the CMP from attempting to load.


raphaelkabo (approved):
👍  Nice one!

PR #1703, opened 2022-05-09
Rk/add ses role to GitHub actions

raphaelkabo:
## What does this change?

- Adds a new GH action in the Cypress ETE with Okta tests config file, which allows the Actions runner to assume a new AWS role
- In Cloudformation, adds a new `GithubActionsSESSendEmailsRole` role, which permits only the sending of emails via SES, and only applies this role in CODE
- Also adds the same permission to the regular role we use for Gateway
- Adds a `RUNNING_IN_CYPRESS` environment variable which we can make use of in Cypress tests on GHA


cloudformation.yaml
```diff
@@ -147,6 +148,36 @@ Resources:
...
+    Type: AWS::IAM::Role
+    Condition: NotIsProd
+    Properties:
+      RoleName: GithubActionsSESSendEmailsRole
+      AssumeRolePolicyDocument:
+        Statement:
+          - Effect: Allow
+            Action: sts:AssumeRoleWithWebIdentity
+            Principal:
+              Federated: arn:aws:iam::942464564246:oidc-provider/token.actions.githubusercontent.com
```
coldlink, line 168: should this ARN be public?

raphaelkabo: I believe ARNs can be public but we can make it a GitHub secret if we prefer?

coldlink: this is in cloudformation, so it can't go in github secrets, i'm just saying this as other ARNs we use in this file are pulled in as parameters rather than made public

coldlink: e.g. https://github.com/guardian/gateway/blob/19f03a7ba11810b2bc88540871dbeb447c30fa35/cloudformation.yaml#L29

coldlink: and then you add the value of the parameter in cloudformation console

coldlink: generally ARNs aren't made public as it may give away information about the infrastructure of our AWS account

coldlink: actually in this case you might just want to hide the account id

similar to what we do elsewhere in the file e.g. https://github.com/guardian/gateway/blob/19f03a7ba11810b2bc88540871dbeb447c30fa35/cloudformation.yaml#L134

```suggestion
              Federated: arn:aws:iam::${AWS::AccountId}:oidc-provider/token.actions.githubusercontent.com
```

dominickendrick (approved):

PR #1720, opened 2022-05-12
 Send emails to existing Okta users who try to register

raphaelkabo:
### Emails

1. Updates the Account Exists email to prioritise the sign in button and deprioritise the reset password button
2. Adds a new function and some logic to generate URLs for emails being sent through Gateway
3. Adds retries to the AWS email sender

### Users who are already registered trying to register

1. Adds logic to send the appropriate email when a user who is already registered tries to register with Okta, depending on their Okta status
2. Adds some new Okta helper methods for this to work

### Testing

1. Adds ETE and Jest tests for Okta for attempts to register and email resending


src/server/lib/retry.ts
```diff
@@ -0,0 +1,40 @@
...
+ * 500ms in between attempts.
+ * Adapted from https://gitlab.com/-/snippets/1775781
+ * @param {Function} params.fn - Returns a promise
+ * @param {Number} params.retriesLeft - Number of retries. If -1 will keep retrying
+ * @param {Number} params.interval - Milliseconds between retries. If exponential is
+ * set to true, will be doubled each retry
+ * @param {Boolean} params.exponential - Flag for exponential back-off mode
+ * @return {Promise<*>}
+ */
+export default async function retry<T>({
```
coldlink, line 20: You may not need this retry method! The AWS SES library seems to have a builtin retry mechanism available as a config option!

https://docs.aws.amazon.com/AWSJavaScriptSDK/latest/AWS/Config.html#maxRetries-property

src/server/lib/okta/register.ts
```diff
@@ -71,27 +151,79 @@ export const register = async (email: string): Promise<UserResponse> => {
...
+        } else {
+          throw new OktaError({
+            message: `Okta user activation failed: Failed to send email`,
+          });
+        }
+      } else {
+        throw new OktaError({
+          message: `Okta registration failed: missing re-activation token`,
+        });
+      }
```
coldlink, line 188: I'm generally more of a fan of guard clauses, as it makes errors and when you throw errors a bit easier to read

So something like this, instead of nesting if/else statements:

```suggestion
if (!tokenResponse?.token.length) {
  throw new OktaError({
    message: `Okta registration failed: missing re-activation token`,
  });
}

const emailIsSent = await sendAccountExistsEmail({
  to: user.profile.email,
  activationToken: tokenResponse.token,
});

if (!emailIsSent) {
  throw new OktaError({
    message: `Okta user activation failed: Failed to send email`,
  });
}

return user;
```

coldlink: can implement a similar thing below too for the other cases too above and below

src/server/lib/retry.ts
coldlink: https://docs.aws.amazon.com/AWSJavaScriptSDK/latest/AWS/Config.html#retryDelayOptions-property

https://docs.aws.amazon.com/AWSJavaScriptSDK/latest/AWS/Request.html#retry-event

src/email/lib/send.ts
```diff
@@ -1,4 +1,5 @@
 import * as AWS from 'aws-sdk';
+import retry from '@/server/lib/retry';
 import { awsConfig } from '@/server/lib/awsConfig';
 
 // We need to adjust the SES timeouts to be longer than the standard AWS
```
coldlink, line 5: Should be able to just add the retry settings to the `sesConfig` object below.

coldlink (commented):
Looking good, a few suggestions I have!

src/server/lib/okta/register.ts
raphaelkabo: Yeah, this is great, I prefer this too!

src/email/lib/send.ts
```diff
@@ -44,5 +50,10 @@ export const send = ({ html, plainText, subject, to }: Props) => {
     FromEmailAddress: 'registration-reply@theguardian.com',
   };
 
-  return SESV2.sendEmail(params).promise();
+  const result = await SESV2.sendEmail(params).promise();
+  console.log(result);
```
coldlink, line 54: CLOG
```suggestion
```

raphaelkabo: AUGH

coldlink (approved):

PR #1728, opened 2022-05-16
Fix register redirects

raphaelkabo:
## What does this change?

- Clientside:
  - Added a `displayRegisterTab` flag to the `SignInPage` component, so that we could selectively show the register tab depending on whether we load `/signin` (yes) or `/reauthenticate` (no).

- Serverside
  - Make the existing `registerMiddleware` middleware more general, renaming it to `loggedInGuard`
  - Attached the new middleware to the `/signin` and `/register` route handlers
  - Created a new route handler for `/reauthenticate`, which renders the `SignInPage` component but hides the register button from the navbar
  - Renamed the `domain` config value to `guardianDotcomDomain` for clarity

[Related PR for necessary Fastly changes](https://github.com/guardian/identity-platform/pull/539).


src/server/lib/middleware/register.ts
```diff
@@ -8,69 +8,79 @@ import { getConfiguration } from '../getConfiguration';
...
-  const { okta, accountManagementUrl } = getConfiguration();
+/**
+ * Middleware function which will redirect the client to a specified URL if
+ * a valid logged in session is found. Currently checks both IDAPI and Okta
+ * session validity.
+ *
+ * @param {string} loggedInRedirectUrl - The URL to which to redirect the client
+ * if a valid logged in session is found.
+ */
+export const loggedInGuard = (loggedInRedirectUrl: string) => {
```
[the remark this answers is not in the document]
raphaelkabo: I think this is fair - I briefly had it using a different route for one handler but I don't think we need that any more.

src/server/lib/middleware/register.ts
```diff
@@ -8,69 +8,79 @@ import { getConfiguration } from '../getConfiguration';
...
+      }
     } else {
-      //otherwise the user has invalid session cookie, so we take them back to the /register page and delete the Identity cookies
-      clearIDAPICookies(res);
-      return res.redirect(
-        addQueryParamsToPath('/register', res.locals.queryParams),
-      );
+      if (identitySessionCookie && identityLastAccessCookie) {
+        logger.info(
+          'User attempting to register had existing IDAPI cookies set.',
```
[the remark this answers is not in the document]
raphaelkabo: Good catch, I'll fix this!

src/server/lib/middleware/loggedInGuard.ts
```diff
@@ -0,0 +1,80 @@
+import { NextFunction, Request } from 'express';
```
dominickendrick, line 1: Could you do a `git-mv` on this file so we can see the changes more clearly. `git-mv` will rename the file without it looking like we've deleted the old one and just created a new one

src/server/lib/middleware/loggedInGuard.ts
```diff
@@ -0,0 +1,80 @@
...
+import { read } from '../idapi/auth';
+import { IDAPIAuthStatus } from '@/shared/model/IDAPIAuth';
+import { clearIDAPICookies } from '../idapi/IDAPICookies';
+const { okta, defaultReturnUri } = getConfiguration();
+
+/** Middleware function which will redirect the client to the base Guardian
+ * domain if a valid logged in session is found. Currently checks both IDAPI and
+ * Okta session validity.
+ */
+export const loggedInGuard = async (
```
dominickendrick, line 17: would `blockLoggedInUsers` be more descriptive ?

src/client/pages/SignInPage.tsx
```diff
@@ -23,6 +23,7 @@ export const SignInPage = () => {
       error={error}
       queryParams={queryParams}
       recaptchaSiteKey={recaptchaSiteKey}
+      displayRegisterTab={displayRegisterTab}
```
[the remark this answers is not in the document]
raphaelkabo: In this case, in that declaration, `true` is the default fallback value if `displayRegisterTab` is undefined. If it's undefined, the tab is shown. We don't want it to always be `true`!

src/server/lib/middleware/loggedInGuard.ts
raphaelkabo: Yes, that's good!

raphaelkabo: Or `redirectLoggedInUsers` might be even better.

dominickendrick: 👍

src/client/pages/SignInPage.tsx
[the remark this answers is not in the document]
dominickendrick: Cool

src/server/lib/middleware/loggedInGuard.ts
raphaelkabo: I didn't know about git mv, thank you!

src/server/lib/middleware/loggedInGuard.ts
raphaelkabo: I'm going with `redirectIfLoggedIn`!

src/client/pages/SignIn.tsx
```diff
@@ -174,25 +178,28 @@ export const SignIn = ({
     executeCaptcha();
   };
 
+  const tabs: TabType[] = [
```
ob6160, line 181: Suggestion: Instead of pushing to the tabs array at runtime, we could declare both possible tab list states more concretely with something like this:

```
const signInTab = {
  displayText: 'Sign in',
  queryParams: queryParams,
  linkTo: '/signin',
  isActive: true,
}

const registerTab = {
  displayText: 'Register',
  queryParams: queryParams,
  linkTo: '/register',
  isActive: false,
}

const tabs = displayRegisterTab ? [signInTab, registerTab] : [signInTab];
```

raphaelkabo: Will this be beneficial for memory reasons or something else?

ob6160: I was thinking more in terms of readability, both possible list states are clearly defined here, with named vars and no need to push

raphaelkabo:
[Register page gets into redirect loop with some signed in users](https://trello.com/c/iEfZvY83/3944-register-page-gets-into-redirect-loop-with-some-signed-in-users)

ob6160 (approved):

PR #1742, opened 2022-05-23
Okta native apps fallback option

coldlink:
## What does this change?

The fallback option for native apps sign in and registration is required in case okta change their systems to no longer support the `fromURI` parameter.

This is the backup option to https://github.com/guardian/gateway/pull/1699

If this ever happens we've programmed a fallback option in the applications for when a session is set within the browser.

Essentially if the `fromURI` parameter is no longer available, we redirect instead to a custom uri scheme/deeplink within the app based on the client id of the application, which we'll always be able to get. The app will listen on this link, and if it's ever fired to call the Okta SDK sign in method again with the `prompt=none` parameter to log them in.

This is essentially the same as the initial proposal we had for the apps.

Rather than hard coding the fallback link for all the applications into gateway, I decided to repurpose the `redirect_uris` option within the application config within Okta. Essentially we can provide multiple valid urls to this setting. So we can add our fallback link here for each application, and at runtime make an API call to read the applications from the Okta API based on the given client id, and then redirect to the fallback url.

## Changes
- Okta JS
  - Code defensively to avoid any possible runtime error and make sure we can get the client id in multiple ways
  - Code in ES5 for maximum compatibility 
  - Change from `clientId` to `appClientId` to differentiate between the other `clientId` parameter we already have in gateway
- Add Okta Apps API with a `getApp` method, which is memoized using a JS `Map` 
- Add the fallback option within the `/oauth/authorization-code/callback` route
- Updates the native apps integration document with the information
- Removed the `returnUrl` functionality for app redirects, as we no longer need this.

## Related
https://github.com/guardian/identity-platform/pull/531


docs/okta/native-apps-integration-guide.md
```diff
@@ -239,4 +239,44 @@ note over NativeLayer: the SDK manages the tokens, which can now be used to<br/>
 
 ## Fallback
 
-TODO
+While we are confident that the above approach is robust enough to work, we will also include a fallback option in case something does go wrong with the above approach.
+
+Specifically a fallback option will be used if we're unable to intercept the `fromURI` parameter from Okta.
+
+To handle this fallback option, a [private-use URI scheme](https://datatracker.ietf.org/doc/html/rfc8252#section-7.1) (referred to as "custom URL scheme") will have to be registered with the app. We suggest a custom URL scheme for this URI, which we can identify (`your-app://identity/fallback`) e.g. `com.theguardian.app://identity/fallback`.
```
dominickendrick, line 246: is it worth mentioning that the custom url also need to be added as a valid return url in the app okta config too ?

coldlink: Good idea, I'll add that in!

PR #1757, opened 2022-06-01
Upgrade to Cypress 10

ob6160:
## What does this change?
Upgrades Cypress from v9 to the new v10 release, following this migration guide: https://docs.cypress.io/guides/references/migration-guide

Cypress 10 brings a beta release of "Component Testing" as well as a refreshed UI, other changes are listed here: https://docs.cypress.io/guides/references/changelog

* New file extension defaults are for `*.cy.ts`, so they have been renamed from `*.spec.ts`. We keep the root directory called `integration` instead of the suggested `e2e`, because our tests include mocked suites: 
  * https://docs.cypress.io/guides/references/migration-guide#Updated-Test-File-Locations
* Configuration file: `cypress.json` has been migrated to the new typescript config file format: `cypress.config.ts`: 
  * https://docs.cypress.io/guides/references/migration-guide#Configuration-File-Changes
* The `testFiles` config option used in our shell scripts has removed and changed to: `e2e: { specPattern: *** }`:
  * https://docs.cypress.io/guides/references/migration-guide#testFiles-specPattern
* The plugins file is now unnecessary, it's handled in the new typescript configuration file. We weren't using ours anyway so it's been removed. The docs linked describe how to use plugins now in the config file:
  * https://docs.cypress.io/guides/references/migration-guide#Plugins-File-Removed
* Upgrades our Cypress GitHub actions to `cypress-io/github-action@v4` to support Cypress v10 and adds `specPattern` to each config specifying the specs to run.
* Adds `--e2e` and `--browser chrome` to the cypress launch on our local test runner scripts so the runner opens first instead of the browser picker.
* Moves out the shared sign-in tests to a `shared` directory


raphaelkabo (commented):
This is great work, thank you for doing it! I've run some Cypress tests and it's all working great, but I'll let someone else check it, too. 👍

coldlink (approved):
looks amazing to me!

PR #1769, opened 2022-06-10
Do not issue Okta sessions for unvalidated users 🚫🐙

dominickendrick:
This PR stops us issuing Okta sessions for unvalidated users. 
This is because we want to only have validated users being logged in to Okta, so we can reduce application complexity around checking for validation status.

Changes:
1. Add check for validation status in signin.ts and only sign user in via IDAPI in this case
2. Add emailValidated field when handleUserResponse from okta users api
3. Updated and added mocked ete tests to cover this use case
4. Checks users authentication status from Okta, and will block login on both Identity and Okta if the status is not `SUCCESS`

Co-authored-by: Patricio Vighi

### DIFF
--- a/src/server/routes/signIn.ts
+++ b/src/server/routes/signIn.ts
@@ -27,7 +27,7 @@ import postSignInController from '@/server/lib/postSignInController';
 import { performAuthorizationCodeFlow } from '@/server/lib/okta/oauth';
 import { getSession } from '../lib/okta/api/sessions';
 import { redirectIfLoggedIn } from '../lib/middleware/redirectIfLoggedIn';
-import { getUser } from '../lib/okta/api/users';
+import { getUserGroups } from '../lib/okta/api/users';
 
 const { okta, accountManagementUrl, oauthBaseUrl } = getConfiguration();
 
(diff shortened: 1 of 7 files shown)


src/server/routes/signIn.ts
```diff
@@ -247,6 +248,17 @@ const oktaSignInController = async (
     // we're authenticated track this metric
     trackMetric('OktaSignIn::Success');
 
+    // If the user was able to log in with Okta BUT NOT VALIDATED, we will still try and sign them in with Identity
+    // This is because we will initally block unvalidated users from getting Okta sessions, but still allow them
+    // Identity sessions.
+
+    if (response._embedded?.user.id) {
```
ob6160, line 255: is it possible for this to be falsy and for them not to be validated? I'm guessing if this is the case, they won't be getting signed in either way

ob6160: maybe we could cover this scenario with a test case

ob6160 (commented):
looks good! a small comment around the program flow

src/server/routes/signIn.ts
dominickendrick: yeah, as per the docs, https://developer.okta.com/docs/reference/api/authn/#request-parameters-for-primary-authentication the api endpoint for authentication can return different `_embedded` objects depending on the users current state https://developer.okta.com/docs/reference/api/authn/#transaction-state for example if a user required MFA you would get a user state returned, but also be required to take further steps to sign the user in successfully. In this case, we should also do a check for `status: 'SUCCESS'`

dominickendrick: I've added a new test and check for this

ob6160: thanks for the explanation — sounds good!

ob6160 (approved):

ob6160 (approved):

PR #1797, opened 2022-06-22
Chromatic "Turbosnap" (build/snapshot only changed components)

ob6160:
## What does this change?
Learnt today in our client side meet-up that Chromatic include a feature called "Turbosnap" (https://www.chromatic.com/docs/turbosnap) which allows you to reduce the time for our builds and the number of snapshots we make based on which stories have changed.

It works by checking whether the webpack dependency graph has changed and then attempts to work out which components have changed. Based on this info it can intelligently rebuild and snapshot only those stories that will have been affected.

We only turn it on for feature branches where we have the most need not to run a lot of unnecessary snapshots, not main where the reasoning was that we want to have a comprehensive Chromatic run.

_Note: according to their docs, turbosnap will only take effect with merge commits, not squash or rebase (https://www.chromatic.com/docs/turbosnap#squashrebase-merging) in this case it should fall back to a full build/snapshot_

### DIFF
--- a/.github/workflows/chromatic.yml
+++ b/.github/workflows/chromatic.yml
@@ -30,3 +30,7 @@ jobs:
           token: ${{ secrets.GITHUB_TOKEN }}
           # 👇 Chromatic projectToken, refer to the manage page to obtain it.
           projectToken: ${{ secrets.CHROMATIC_APP_CODE }}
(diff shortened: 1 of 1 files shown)


coldlink (approved):
🚀 TURBO 🚀

liamness (approved):
📷 SNAP 📸

PR #1803, opened 2022-06-27
fix hover style for eye part of password input field

pvighi:
## What does this change?
Made the focus style for the eye button consistent with the other fields

**before:**

https://user-images.githubusercontent.com/15324270/175987671-e5b580d1-b384-41e8-82db-6fe81a1cd21a.mov

**after**

https://user-images.githubusercontent.com/15324270/175987974-54897a7c-8508-4c28-9465-02211a776e34.mov


pvighi:
[Change outline around 'reveal password' eye icon in password fields to match the one around the input element](https://trello.com/c/0kKgYN8k/3878-change-outline-around-reveal-password-eye-icon-in-password-fields-to-match-the-one-around-the-input-element)

PR #1805, opened 2022-06-30
Add Jest test for security headers, improve Supertest setup

raphaelkabo:
## What does this change?

As part of our threat mitigation tasks, we should make sure that our [CSP](https://developer.mozilla.org/en-US/docs/Web/HTTP/CSP) setup is working as expected.

This PR adds a new Jest test (`headers.test.ts`) which checks every route listed in the array `ValidRoutePathsArray` in `@/shared/model/Routes` for the presence of a CSP header, and checks that all the parameters in it are set as we've defined them. It also adds a check for the [STS](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Strict-Transport-Security) header. As this is not an E2E test, every route in the array may not run as expected, especially since we mock a lot of the server infrastructure during the tests - but every route should come back with CSP and STS headers, which are tested for here.

We use [Supertest](https://www.npmjs.com/package/supertest) to test the server, and as part of improving the way it works and making the tests less flaky, I've made a few changes to the way the Express server is run:

- Instead of running both `express()` and `express.listen()` in one file, I've split the functionality into two files. `@/server/server.ts` exports a function, `createServer()`, which creates an Express server, sets up its middleware, and returns it. `@/server/index.ts` executes this function and begins the server listener. Although the standard server functionality is now in two files, no behaviour has changed in the way the app is started and run.
- As a result of this change, Supertest now imports and uses just the `createServer()` function, not the listener. This means that Supertest itself has full control over how ports are opened and closed during testing, which prevents [a lot of flaky behaviour](https://github.com/visionmedia/supertest/issues/568) involving ports not being closed properly. This allows us to continue running our tests in parallel and will make it easier to add more Supertest tests in future.
- I've also moved the function that @ob6160 created to spin up a testing server into a separate file, `sharedConfig.ts`, which allows us to use it across tests in future. I've made some changes to Olly's rate limiter tests as a result of all this refactoring.

### DIFF
--- a/src/server/lib/__tests__/headers.test.ts
+++ b/src/server/lib/__tests__/headers.test.ts
@@ -1,0 +1,123 @@
+import { ValidRoutePathsArray } from '@/shared/model/Routes';
+import Redis from 'ioredis-mock';
+import request from 'supertest';
+import {
+  defaultRateLimiterConfiguration,
+  getServerInstance,
+} from './sharedConfig';
+
+// Override the default 5s max timeout for these tests because Supertest takes some time to run.
+jest.setTimeout(10000);
+
+/**
+ * We mock a section of our app infrastruture so that the routes that Supertest
+ * runs complete without throwing early errors. We're only testing the headers
+ * we set in middleware here.
+ */
+jest.mock('@/server/lib/getAssets', () => ({
+  getAssets: () => ({
+    main: { js: 'mocked' },
+    vendors: { js: 'mocked' },
+    runtime: { js: 'mocked' },
+  }),
+}));
+jest.mock('@/server/lib/serverSideLogger', () => ({
+  logger: {
+    info: jest.fn(),
+    error: jest.fn(),
+  },
+}));
+jest.mock('aws-sdk');
+jest.mock('@/server/lib/getAssets', () => ({
+  getAssets: () => ({
+    main: { js: 'mocked' },
+    vendors: { js: 'mocked' },
+    runtime: { js: 'mocked' },
+  }),
+}));
+jest.mock('@/server/lib/serverSideLogger', () => ({
+  logger: {
+    info: jest.fn(),
+    error: jest.fn(),
+  },
+}));
+jest.mock('@/server/lib/redis/redisClient', () => new Redis());
+jest.mock('@/server/lib/middleware/login', () => ({
+  loginMiddleware: jest.fn((req, res, next) => next()),
+}));
+jest.mock('@/server/lib/IDAPIFetch');
+
+describe('Content Security Policy headers', () => {
+  test.each(ValidRoutePathsArray)(
+    'returns CSP headers for %s',
+    async (route) => {
+      const server = await getServerInstance(defaultRateLimiterConfiguration);
+      const response = await request(server).get(route);
+      // Dynamically import getConfiguration so the changes to the environment
+      // variables are reflected inside the getConfiguration() call
+      const { getConfiguration } = await import(
+        '@/server/lib/getConfiguration'
+      );
+      const { baseUri, apiDomain } = getConfiguration();
+      // Is the CSP header set?
+      expect(response.header).toHaveProperty('content-security-policy');
+      const splitCSPHeader =
+        response.header['content-security-policy'].split(';');
+      // Does the CSP header match what we expect it to be? If the CSP settings
+      // in Helmet are updated, we expect to need to update these tests.
+      expect(splitCSPHeader).toContain("base-uri 'none'");
+      expect(splitCSPHeader).toContain("default-src 'none'");
+      expect(splitCSPHeader).toContain(
+        [
+          'script-src',
+          baseUri,
+          "'sha256-411aj9j2RJj78RLGlCL/KDMK0fe6OEh8Vp6NzyYIkP4='",
+          'www.google-analytics.com',
+          'sourcepoint.theguardian.com',
+          'gdpr-tcfv2.sp-prod.net',
+          'ccpa.sp-prod.net',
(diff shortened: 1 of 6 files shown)


src/server/lib/__tests__/headers.test.ts
```diff
@@ -0,0 +1,71 @@
...
+    // Tests call this method on success and error to tear down the server.
+    const teardownServer = async () => {
+      // Remove the scheduled teardown because we're doing it now.
+      await new Promise((r) => server.close(r));
+    };
+
+    return { server, teardownServer };
+  };
+
+  it('should ', async () => {
```
ob6160, line 46: can we get a description in for this test case here please

src/server/lib/__tests__/headers.test.ts
```diff
@@ -0,0 +1,71 @@
...
+    const { server, teardownServer } = await getServerInstance();
+    // Check our base routes that have user input forms
+    ['/register', '/sign-in', '/reset-password'].forEach(async (route) => {
+      try {
+        const response = await request(server).get(route);
+        // Is the CSP header set?
+        expect(response.header).toHaveProperty('content-security-policy');
```
ob6160, line 53: I'm guessing this is just the first pass of testing as it's only checking these three routes. Do you plan to extend this base file further to cover all our routes?

ob6160 (commented):
nice work! just a few comments/queries

src/server/lib/__tests__/headers.test.ts
raphaelkabo: I was trying to work out the best way to do this - at the moment, these are the three core routes which have forms and where we definitely want our CSP to work, and most of the other routes we serve will redirect to /signin if the user doesn't log in. I could write the test in Cypress instead and do this more in-depth but as an initial pass I think this does the job - would love your feedback!

ob6160 (commented):
just checking — is this still open for review or a work in progress after chatting about it yesterday?

raphaelkabo:
@ob6160 This is now once again open for review!

src/server/lib/__tests__/headers.test.ts
```diff
@@ -0,0 +1,74 @@
...
+      expect(splitCSPHeader).toContain("default-src 'none'");
+      expect(splitCSPHeader).toContain(
+        `script-src ${baseUri} 'sha256-411aj9j2RJj78RLGlCL/KDMK0fe6OEh8Vp6NzyYIkP4=' www.google-analytics.com sourcepoint.theguardian.com gdpr-tcfv2.sp-prod.net ccpa.sp-prod.net ccpa-service.sp-prod.net ccpa-notice.sp-prod.net cdn.privacy-mgmt.com www.google.com www.gstatic.com assets.guim.co.uk 'unsafe-eval'`,
+      );
+      expect(splitCSPHeader).toContain(
+        `img-src ${baseUri} static.guim.co.uk ophan.theguardian.com www.google-analytics.com www.google.com`,
+      );
+      expect(splitCSPHeader).toContain('font-src assets.guim.co.uk');
+      expect(splitCSPHeader).toContain(
+        `connect-src www.google-analytics.com vendorlist.consensu.org consent-logs.${apiDomain} sourcepoint.theguardian.com gdpr-tcfv2.sp-prod.net ccpa.sp-prod.net ccpa-service.sp-prod.net ccpa-notice.sp-prod.net cdn.privacy-mgmt.com api.nextgen.guardianapps.co.uk https://api.pwnedpasswords.com localhost:1234 www.google.com o14302.ingest.sentry.io`,
```
ob6160, line 59: i wonder if the two longer strings (this one and the one on line 52) could be constructed from arrays that are then joined into the final string?

suggesting because some of them are sourced from our mocked env object and would be good to assert against the values stored there so when the env object is changed this changes too.. also it might make it a more pleasant experience to read/modify them in the future :)

src/server/lib/__tests__/headers.test.ts
```diff
@@ -0,0 +1,74 @@
...
+    jest.mock('@/server/lib/serverSideLogger', () => ({
+      logger: {
+        info: jest.fn(),
+        error: jest.fn(),
+      },
+    }));
+
+    jest.mock('@/server/lib/redis/redisClient', () => new Redis());
+
+    jest.mock('@/server/lib/middleware/login', () => ({
```
ob6160, line 32: a really small explanatory comment around why we're mocking these might be helpful

src/server/lib/__tests__/headers.test.ts
```diff
@@ -0,0 +1,74 @@
...
+      loginMiddleware: jest.fn((req, res, next) => next()),
+    }));
+    jest.mock('@/server/lib/IDAPIFetch');
+  });
+
+  it('should return CSP headers for routes', async () => {
```
ob6160, line 38: could be worth explicitly saying this covers all routes in the test description here

ob6160 (commented):
looks good, thanks for making those changes — just a few more comments

src/server/lib/__tests__/headers.test.ts
raphaelkabo: Done!

src/server/lib/__tests__/rate-limit/rateLimitMiddleware.test.ts
```diff
@@ -150,8 +98,6 @@ describe('rate limiter middleware', () => {
       await request(server).get('/register').expect(200);
     } catch (error) {
       throw error;
-    } finally {
-      await teardownServer();
```
ob6160, line 154: what's happened to cause these to be removed?

PR #1806, opened 2022-06-30
Support creation of Jobs users directly through Okta

ob6160:
## What does this change?
Now that we have positioned Okta as the arbiter of truth for the "GRS" group using the `isJobsUser` field, we are able to implement an Okta variant of the Jobs user registration and terms agreement flow.

The `GRS` keyword floats around still (of note in the URL: `/agree/GRS`). This is because Madgex still link to the `/agree/GRS` route, and this is not something we can change until after they've completed their integration with Okta.

The existing e2e tests for the Jobs agreement flow were ported into a separate Okta suite which has the same coverage as the IDAPI variant. Some e2e helper methods also didn't have the "retry on error" property set, this was set to hopefully reduce flakiness on the IDAPI side of the tests.

### DIFF
--- a/cypress/integration/ete-okta/jobs_terms.cy.ts
+++ b/cypress/integration/ete-okta/jobs_terms.cy.ts
@@ -1,0 +1,183 @@
+describe('Jobs terms and conditions flow in Okta', () => {
+  context('Accepts Jobs terms and conditions and sets their name', () => {
+    it('should redirect users with an invalid session cookie to reauthenticate', () => {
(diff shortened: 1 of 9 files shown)


dominickendrick (approved):

ob6160:
[Guardian Jobs migrate GRS group to Okta. (S)](https://trello.com/c/WCTKBvxO/3008-guardian-jobs-migrate-grs-group-to-okta-s)

PR #1820, opened 2022-07-07
Add identity fallback endpoint to apple site association

coldlink:
## What does this change?

- Update the apple site association file to add the fallback endpoint so they can intercept it


raphaelkabo (approved):
🍏

PR #1821, opened 2022-07-07
Close active Okta session on POST to /signin and /reauthenticate

coldlink:
## What does this change?

This resolves:

1. A problem with the `/reauthenticate` page:
    - Given **User A** is signed in with Okta
    - When **User B** POSTS from the `/reauthenticate` page
    - Then, after verifying their login details, Okta will [keep **User A** signed in despite **User B**'s details being entered](https://developer.okta.com/docs/guides/session-cookie/main/).
    This is becasue "Okta Classic orgs ignore the sessionToken in a request if there is already a session cookie set in the browser." For more on these flows and our new solution, see `docs/reauthenticate.md`.
3. A bug where a user who was already logged in would end up in an endless loop after POSTing from `/reauthenticate`, because a piece of code would see they had a valid `sid` cookie and send them straight back to `/reauthenticate`.

The solution: In the `performAuthorizationCodeFlow` function, we pass a parameter to close the existing session. If true, before attempting to negotiate with Okta for a new session, we call an Okta endpoint to close the existing session on the current browser. When Okta tries to set a new session, it will see that the `sid` cookie is now invalid, and overwrite it with a new one. This parameter is currently set to `true` for all calls to POST `/signin` and `/reauthenticate`.

## Caveats

Due to the new behaviour, if the user is signed in elsewhere on the Guardian on the same browser using [OpenID Connect access tokens](https://developer.okta.com/docs/concepts/oauth-openid/), these tokens will _not_ be destroyed during reauthenticate flow. We do not currently use these tokens, but may do in future. This can lead to a situation where **User A** is logged in on theguardian.com, but **User B** is logged into MMA, or similar.

### DIFF
--- a/src/server/lib/okta/api/sessions.ts
+++ b/src/server/lib/okta/api/sessions.ts
@@ -33,6 +33,29 @@ export const getSession = async (
   }).then(handleSessionResponse);
 };
 
+/**
+ * Close a User session by session Id
+ *
+ * https://developer.okta.com/docs/reference/api/sessions/#close-session
+ *
+ * The Okta API closes the users session on success or
+ * returns a 404 on invalid.
+ *
+ * @param sessionId Okta session ID
+ * @returns Promise<boolean>
+ */
+export const closeSession = async (sessionId: string): Promise<undefined> => {
+  const path = buildUrl('/api/v1/sessions/:sessionId', { sessionId });
+  const response = await fetch(joinUrl(okta.orgUrl, path), {
+    method: 'DELETE',
+    headers: { ...defaultHeaders, ...authorizationHeader() },
+  });
+
+  if (!(response.ok || response.status === 404)) {
+    return handleErrorResponse(response);
+  }
+};
+
 export const handleSessionResponse = async (
   response: Response,
 ): Promise<SessionResponse> => {
--- a/src/server/models/okta/User.ts
+++ b/src/server/models/okta/User.ts
@@ -101,3 +101,12 @@ export enum Status {
   PASSWORD_RESET = 'PASSWORD_RESET',
   SUSPENDED = 'SUSPENDED',
 }
+
+// https://developer.okta.com/docs/reference/api/sessions/#session-object
+export interface SessionResponse {
+  id: string;
+  login: string;
+  userId: string;
+  expiresAt: string;
+  status: string;
+}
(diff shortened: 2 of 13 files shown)


dominickendrick (dismissed):

src/server/models/okta/User.ts
```diff
@@ -101,3 +101,12 @@ export enum Status {
   PASSWORD_RESET = 'PASSWORD_RESET',
   SUSPENDED = 'SUSPENDED',
 }
+
+// https://developer.okta.com/docs/reference/api/sessions/#session-object
+export interface SessionResponse {
```
dominickendrick, line 106: Is this is already defined here https://github.com/guardian/gateway/blob/50c35f0a6a42129d68acd3f58a1b6d0d1db3f850/src/server/models/okta/Session.ts

src/server/lib/okta/api/sessions.ts
```diff
@@ -33,6 +33,29 @@ export const getSession = async (
...
+ */
+export const closeSession = async (sessionId: string): Promise<undefined> => {
+  const path = buildUrl('/api/v1/sessions/:sessionId', { sessionId });
+  const response = await fetch(joinUrl(okta.orgUrl, path), {
+    method: 'DELETE',
+    headers: { ...defaultHeaders, ...authorizationHeader() },
+  });
+
+  if (!(response.ok || response.status === 404)) {
+    return handleErrorResponse(response);
```
ob6160, line 55: the first part of this logic makes sense to me, as in "if response.ok is falsy, handle error response".

But the second part I think is saying "if response.status is not 404", which is confusing me.. what if the status is 200, why would that warrant handling an error response?

raphaelkabo: The endpoint will close the session ID passed to it, *and* if the session we asked it to close doesn't exist/is invalid, it will return 404. Otherwise it'll pass `204 No Content`. So 404 isn't an error code we need to know about - if the session doesn't exist, it doesn't need to be closed.

`response.ok` is returned `true` from `fetch()` on [a status in the range 200-299](https://developer.mozilla.org/en-US/docs/Web/API/Response/ok). So here we're saying "if the response is any code 0-199 and 300-infinity but not 404, return an error".

I can add some of this in a comment above this line!

src/server/models/okta/User.ts
raphaelkabo: Yes it is, I'll remove the unnecessary import!

src/server/lib/okta/api/sessions.ts
raphaelkabo: Could be clearer to write `if (response.ok === false && response.status != 404)`.

ob6160: it does look clearer, having the external logical NOT made it harder to read
yep— that comment idea would be great just because it wasn't immediately obvious

PR #1835, opened 2022-07-12
Upgrade babel-preset-react

ob6160:
## What does this change?
We've held back upgrading `@babel/preset-css-prop` for a while because it broke the `css` prop in Storybook.

This PR upgrades the package and filters it out of the loaded presets as we already add it here: 

https://github.com/guardian/gateway/blob/b334494844e022c0cd3ff2e0313b1ea58d48542d/.storybook/main.js#L12 

This caused a problem where the `css` prop was not evaluated as expected.

## Testing

- [x] Tested on CODE
- [x] Tested on IE11


coldlink (approved):
I could cry. 🥲

PR #1836, opened 2022-07-13
Add Okta session refresh endpoint

raphaelkabo:
## What does this change?
It adds a new endpoint that we can redirect users to when their Idapi session cookies need refreshing so that we keep Okta and Idapi cookies in sync as far as possible.  This will mitigate against any potentially strange behaviour where cookies are in different states.

## How to test
There are Cypress tests included but these are the test scenarios for manual testing:
### User with a valid Okta session
1. Sign in as user
2. Go to /signin/refresh?useOkta=true&returnUrl=...
=> Redirected through authorisation code flow and ultimately back to `returnUrl`. 
Cookie state should be refreshed `sid` cookie.
### User with an invalid Okta session
1. Sign in as user
2. Go to /signin/refresh?useOkta=true&returnUrl=...
=> Redirected to `returnUrl`.
Cookie state should be no `sid` cookie.
### User without an Okta session but with an Idapi session
1. Sign in as user
2. Go to /signin/refresh?returnUrl=...
=> Redirected to `returnUrl`.
Cookie states should be unchanged.
### User with neither an Okta session or an Idapi session
1. Sign in as user
2. Go to /signin/refresh?returnUrl=...
=> Redirected to sign-in page.
Cookie states should be unchanged.

Potential risk of being sent around in a loop appears to be low.


src/server/routes/signIn.ts
```diff
@@ -405,6 +406,48 @@ router.post(
...
+ */
+router.get(
+  '/signin/refresh',
+  handleAsyncErrors(async (req: Request, res: ResponseWithRequestState) => {
+    const { useOkta, returnUrl } = res.locals.queryParams;
+    const oktaSessionCookieId: string | undefined = req.cookies.sid;
+    const identitySessionCookie = req.cookies.SC_GU_U;
+    const identityLastAccessCookie = req.cookies.SC_GU_LA;
+
+    const { defaultReturnUri } = getConfiguration();
```
coldlink, line 423: This is a constant, so you can move this to the top of the file (L33).

src/server/routes/signIn.ts
```diff
@@ -405,6 +406,48 @@ router.post(
...
+        // the SID cookie value to Okta.
+        await getSession(oktaSessionCookieId);
+        return performAuthorizationCodeFlow(req, res);
+      } catch {
+        //if the cookie exists, but the session is invalid, we remove the cookie
+        //and return them to the URL they came from
+        clearOktaCookies(res);
+        return res.redirect(redirectUrl);
+      }
+    } else if (identitySessionCookie && identityLastAccessCookie) {
```
coldlink, line 439:
```suggestion
    } else if (identitySessionCookie) {
```

we shouldn't check for the `SC_GU_LA` cookie in this flow, as that cookie is only ever set for 30 mins, so most users won't have this cookie anyway when logged in

src/server/routes/signIn.ts
```diff
@@ -405,6 +406,48 @@ router.post(
...
+    const { defaultReturnUri } = getConfiguration();
+    const redirectUrl = returnUrl || defaultReturnUri;
+
+    // Check if the user has an existing Okta session.
+    if (okta.enabled && useOkta && oktaSessionCookieId) {
+      try {
+        // If the user session is valid, we re-authenticate them, supplying
+        // the SID cookie value to Okta.
+        await getSession(oktaSessionCookieId);
+        return performAuthorizationCodeFlow(req, res);
```
coldlink, line 432: Something I completely forgot about during the planning!

On the OAuth callback endpoint (https://github.com/guardian/gateway/blob/main/src/server/routes/oauth.ts#L174-L185), all 3 identity cookies are actually refreshed/set. However only the `GU_U` and `SC_GU_U` cookies should be set from this call when a refresh happens, not the `SC_GU_LA`.

The best way to do this would be to add an optional flag to the `AuthorizationState` (https://github.com/guardian/gateway/blob/main/src/server/lib/okta/openid-connect.ts#L24), e.g. something like `noRefreshSCGULA?: boolean`.

Set this flag as an option in the `performAuthrozationCodeFlow` method (`performAuthorizationCodeFlow(req, res, { noRefreshSCGULA: true });`).

Then refactor the callback endpoint and the `setIDAPICookies` method to also utilize this flag.

It may sound a bit complicated, so give me a shout if this doesn't make sense!

coldlink (commented):
Looking good so far! Just some comments. Give me a shout if you need to go through them.

src/server/routes/signIn.ts
kelvin-chappell: @coldlink The fix is in 41d358884957fc6636041e31fcadc97dc482a064 and the test for it in 20e2ce17131e4f52ee402227f62948e1468e42b2, if you don't mind having another look.

coldlink: Great, looks good to me!

coldlink (approved):

PR #1837, opened 2022-07-13
feat: add retry logic to the setup jobs user method because the call …

ob6160:
…to IDAPI can fail sometimes

## Accessibility

### DIFF
--- a/src/server/lib/jobs.ts
+++ b/src/server/lib/jobs.ts
@@ -27,11 +27,12 @@ export const setupJobsUserInOkta = (
   });
 };
 
-export const setupJobsUserInIDAPI = (
+export const setupJobsUserInIDAPI = async (
   firstName: string,
   secondName: string,
   ip: string,
   sc_gu_u: string,
+  retries: number,
 ) => {
   if (firstName === '' || secondName === '') {
     throw new Error('Empty values not permitted for first or last name.');
@@ -43,8 +44,21 @@ export const setupJobsUserInIDAPI = (
   // they try to sign in to the Jobs site for the first time.
   //
   // We can resolve both promises here because they are not dependent on each other.
-  return Promise.allSettled([
-    updateName(firstName, secondName, ip, sc_gu_u),
-    addToGroup(GroupCode.GRS, ip, sc_gu_u),
-  ]);
+  //
+  // We retry this operation up to the number of retries specified because the requests
+  // to IDAPI can intermittently fail.
+  try {
+    return await Promise.allSettled([
+      updateName(firstName, secondName, ip, sc_gu_u),
+      addToGroup(GroupCode.GRS, ip, sc_gu_u),
+    ]);
+  } catch (error) {
+    if (retries > 0) {
+      setTimeout(() => {
+        setupJobsUserInIDAPI(firstName, secondName, ip, sc_gu_u, retries - 1);
+      }, 100);
+    } else {
+      throw error;
+    }
+  }
 };
--- a/src/server/controllers/changePassword.ts
+++ b/src/server/controllers/changePassword.ts
@@ -80,6 +80,7 @@ const changePasswordInIDAPI = async (
           secondName,
           req.ip,
           SC_GU_U.value,
+          3,
         );
         trackMetric('JobsGRSGroupAgree::Success');
       }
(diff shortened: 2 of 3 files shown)


src/server/lib/jobs.ts
```diff
@@ -43,8 +44,21 @@ export const setupJobsUserInIDAPI = (
...
+  } catch (error) {
+    if (retries > 0) {
+      setTimeout(() => {
+        setupJobsUserInIDAPI(firstName, secondName, ip, sc_gu_u, retries - 1);
+      }, 100);
+    } else {
+      throw error;
+    }
+  }
 };
```
coldlink, line 64: i'm wondering if this could cause some funky issues due to the way javascript will return from the setupJobUsersInIDAPI method as i expect an implicit `return undefined` from the method before setTimeout executes


this is the way I'm reading the code (not a suggestion)
```ts
  } catch (error) {
    if (retries > 0) {
      // set timeout gets called here
      setTimeout(() => {
        setupJobsUserInIDAPI(firstName, secondName, ip, sc_gu_u, retries - 1);
      }, 100);
     // before the 100ms is up the code will continue executing here
    // theres nothing else here so it will go out of this block
    } else {
      throw error;
    }
   // it will then end up here, it sees the end of the method, will do the implicit `return undefined` before this the setTimeout callback method is called
  // could this cause potential issues?
  }
};
```

coldlink: something like this perhaps for a promise based approach?

```ts
// set timeout to wait a specified amount of time as a promise
const wait = (ms: number) => new Promise((resolve) => {
  setTimeout(() => resolve(), ms);
});
```

then something like

```suggestion
  } catch (error) {
    if (retries > 0) {
      await wait(100)
      return  setupJobsUserInIDAPI(firstName, secondName, ip, sc_gu_u, retries - 1);
    } else {
      throw error;
    }
  }
};
```

coldlink: something like this maybe?

```ts
const retryWithTimeout = async (
fn: Promise<unknown>
retries: number,
delay: number
) => {
  try {
    await fn();
  } catch (err) {
    if (retries <= 0) {
      return Promise.reject(err);
    }

    await wait(delay);
    
    return retryWithTimeout(fn, (retries - 1), delay);
  }
}
```

ob6160: all good suggestions — that's very much along the lines of my thinking. will see what i can cook up

coldlink: maybe for now do the simple promise timeout `wait` method, and if we need it in other places we can always introduce it.

ob6160: agreed, we can abstract it out later

ob6160:
[Fix Jobs First/Last name change flakiness](https://trello.com/c/8g9rmDTF/4030-fix-jobs-first-last-name-change-flakiness)

src/server/lib/jobs.ts
```diff
@@ -43,8 +61,31 @@ export const setupJobsUserInIDAPI = (
...
+  //
+  // We retry this operation up to the number of retries specified because the requests
+  // to IDAPI can intermittently fail.
+  try {
+    const updateResult = await Promise.allSettled([
+      updateName(firstName, secondName, ip, sc_gu_u),
+      addToGroup(GroupCode.GRS, ip, sc_gu_u),
+    ]);
+
+    return updateResult;
```
coldlink, line 73: could this just become?

```suggestion
    return await Promise.allSettled([
      updateName(firstName, secondName, ip, sc_gu_u),
      addToGroup(GroupCode.GRS, ip, sc_gu_u),
    ]);
```

coldlink (commented):
one more suggestion

src/server/lib/jobs.ts
ob6160: 👍

coldlink (approved):
Approving, but would be nice to get a test for this if possible.

Best way might to do a cypress mocked test, so we can mock a failed API call to check the retry behaviour works?

ob6160:
> Approving, but would be nice to get a test for this if possible.
> 
> Best way might to do a cypress mocked test, so we can mock a failed API call to check the retry behaviour works?

good point — 

> Approving, but would be nice to get a test for this if possible.
> 
> Best way might to do a cypress mocked test, so we can mock a failed API call to check the retry behaviour works?

good point, I think I will put a test in here to have confidence it's working 👍

pvighi (approved):
looks good

PR #1838, opened 2022-07-18
Okta | Check for breached password when setting a new password

coldlink:
## What does this change?

IDAPI would check against the breached passwords api when setting a password on the server. This functionality was not added to gateway when we migrated the setting of passwords to Okta.

This helps to form a 3 line defense against the setting of weak password.

Firstly there is already an existing client side check in the browser against the breached passwords API which should catch most cases. However we should not rely purely on client side checks.

So we add this server side check to also check for breached passwords before we send the password to the Okta API to be set. This prevents users who bypass the client side check from setting a week password. this is what's set up in this PR.

Finally there is a 3rd check within Okta itself to check against it's own dictionary. This will be enabled by this PR: https://github.com/guardian/identity-platform/pull/549

### DIFF
--- a/src/server/lib/breachedPasswordCheck.ts
+++ b/src/server/lib/breachedPasswordCheck.ts
@@ -1,0 +1,75 @@
+import { createHash } from 'crypto';
+import { fetch } from '@/server/lib/fetch';
+import { logger } from '@/server/lib/serverSideLogger';
+import { trackMetric } from '@/server/lib/trackMetric';
+
+/**
+ * @name: isBreachedPassword
+ * @description: Checks if a password has previously been in a data breach by checking the Pwned Passwords API
+ *
+ * The documentation for the pwned password API is here for more detail:
+ * https://haveibeenpwned.com/API/v3#PwnedPasswords
+ *
+ * This method returns true only if the password has been in a data breach.
+ *
+ * We return false otherwise. This could be the case that the password has not been in a data breach, or if there is an error with the API/fetch call. This way we don't block the user from changing their password should there be an error.
+ *
+ * @param {string} password Password to check against the Pwned Passwords API
+ * @returns {boolean} True if the password has been in a data breach, false if it hasn't been breached or there was an API error
+ */
+export const isBreachedPassword = async (
+  password: string,
+): Promise<boolean> => {
+  try {
+    // the Pwned Passwords API uses SHA1 hashes of the password
+    // so we need to hash the password before sending it to the API
+    const hash = createHash('sha1').update(password).digest('hex');
+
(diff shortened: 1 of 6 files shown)


src/server/lib/okta/api/authentication.ts
```diff
@@ -129,6 +132,14 @@ export const resetPassword = async (body: {
   newPassword: string;
 }): Promise<AuthenticationTransaction> => {
   const path = buildUrl('/api/v1/authn/credentials/reset_password');
+
+  if (await isBreachedPassword(body.newPassword)) {
```
ob6160, line 136: I wonder if this should be so tightly coupled with our Okta API method here. If I called `resetPassword` I think i'd expect it to just send my request directly to Okta and not apply any extra checks.

I realise that in most, if not all of the cases we use this method we want to apply this check, but the pattern across our codebase appears to be that these methods are purely calls to the related API endpoint. Perhaps a version of this called `resetPasswordWithBreachedCheck`, or just a boolean argument to this method could be more explicit?

src/server/lib/breachedPasswordCheck.ts
```diff
@@ -0,0 +1,40 @@
...
+import { fetch } from '@/server/lib/fetch';
+import { logger } from '@/server/lib/serverSideLogger';
+import { trackMetric } from '@/server/lib/trackMetric';
+
+export const isBreachedPassword = async (
+  password: string,
+): Promise<boolean> => {
+  try {
+    const hash = createHash('sha1').update(password).digest('hex');
+    const first5 = hash.substring(0, 5);
```
ob6160, line 11: a doc comment for this method / explanatory comment about the method of checking whether a password is breached could really help to illustrate the logic behind how the breached check works here

ob6160 (changes requested):

src/server/lib/okta/api/authentication.ts
coldlink: Yeah I see what you're getting at, but in this case I think we want a tight coupling to the method, as in all cases where we update a password in Okta we should do a breached check first for our security compliance.

By separating them out into two different methods, future developers might accidently use the normal `resetPassword` method rather than the `resetPasswordWithBreachedCheck` method without realising.

I'll update the documentation so that this reasoning is clear.

src/server/lib/breachedPasswordCheck.ts
coldlink: Good idea! I'll add some additional comments.

ob6160 (approved):
looks good! Nice work 🔏🔐

PR #1849, opened 2022-07-18
Dependency upgrades (18th July 2022)

ob6160:
## What does this change?

Changes from previous #1832  :

Holds back the following:

## Tested

- [x] CODE
- [x] IE11


kelvin-chappell (approved):
👍

PR #1856, opened 2022-07-20
Okta | Update session expiry behaviour

raphaelkabo:
## What does this change?

This resolves unintended behaviour on the page where a user types in a new password (`/reset-password/:recovery-token`), reached from a link in the reset password email.

**Previous behaviour**

1. The client loads the page, supplying a **recovery token** with a 60-minute expiry.
2. The server checks that the **recovery token** is valid and exchanges it with Okta for a **session token**, a short-lived token with a fixed expiry time of 5 minutes 30 seconds.
3. The server uses the **session token's** expiry time (as it does not have access to the **recovery token's** expiry time) to set a client-side timeout. After 5:30, the page is refreshed and the client is shown a 'token expired' page, even though the **recovery token** may not have expired - only the **session token**.
4. The user is confused and sad.

**New behaviour**

1. The client loads the page, supplying a **recovery token** with a 60-minute expiry.
2. The server checks that the **recovery token** is valid. If it's valid, it shows the page to let the user enter a new password. If it's invalid, it will show them the 'token expired' page.
3. _When the user submits the form_, the server again checks that the **recovery token** is valid and, if so, exchanges it with Okta for a **session token**. If it's invalid, it will send them to the 'token expired' page.
4. The server immediately uses this **session token** to change the user's password with Okta.
5. The user is happy, at least as far as changing their Guardian password goes!

### DIFF
--- a/src/server/controllers/changePassword.ts
+++ b/src/server/controllers/changePassword.ts
@@ -12,10 +12,7 @@ import {
 } from '@/server/lib/idapi/changePassword';
 import { setIDAPICookies } from '@/server/lib/idapi/IDAPICookies';
 import { trackMetric } from '@/server/lib/trackMetric';
-import {
-  readEncryptedStateCookie,
-  updateEncryptedStateCookie,
-} from '@/server/lib/encryptedStateCookie';
+import { updateEncryptedStateCookie } from '@/server/lib/encryptedStateCookie';
 import { ApiError } from '@/server/models/Error';
 import { PasswordRoutePath, RoutePaths } from '@/shared/model/Routes';
 import { PasswordPageTitle } from '@/shared/model/PageTitle';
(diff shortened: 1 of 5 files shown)


src/server/controllers/changePassword.ts
```diff
@@ -172,13 +172,24 @@ const changePasswordInOkta = async (
   res: ResponseWithRequestState,
   successRedirectPath: RoutePaths,
 ) => {
-  const { stateToken } = readEncryptedStateCookie(req) ?? {};
```
ob6160, line 175: is `stateToken` being used anymore? Can we remove it from the encrypted state cookie as a result?

coldlink: Good point, have removed it from the encrypted state cookie

ob6160 (approved):

PR #1857, opened 2022-07-20
change link expiration message to 60 mins

pvighi:
## What does this change?
Change the pages and email templates to say link expiration is 60 minutes instead of 30

### DIFF
--- a/cypress/integration/ete/registration/register.cy.ts
+++ b/cypress/integration/ete/registration/register.cy.ts
@@ -209,7 +209,7 @@ describe('Registration flow', () => {
         expect(body).to.have.string('This account already exists');
         expect(body).to.have.string('Sign in');
         expect(body).to.have.string('Reset password');
-        expect(body).to.have.string('This link is only valid for 60 minutes.');
+        expect(body).to.have.string('This link is only valid for 30 minutes.');
 
         expect(links.length).to.eq(3);
 
@@ -250,7 +250,7 @@ describe('Registration flow', () => {
         expect(body).to.have.string(
           'To continue to your account please click below to create a password.',
         );
-        expect(body).to.have.string('This link is only valid for 60 minutes.');
+        expect(body).to.have.string('This link is only valid for 30 minutes.');
         expect(body).to.have.string('Create password');
 
         expect(links.length).to.eq(2);
--- a/cypress/integration/ete/registration/register.cy.ts
+++ b/cypress/integration/ete/registration/register.cy.ts
@@ -209,7 +209,7 @@ describe('Registration flow', () => {
         expect(body).to.have.string('This account already exists');
         expect(body).to.have.string('Sign in');
         expect(body).to.have.string('Reset password');
-        expect(body).to.have.string('This link is only valid for 30 minutes.');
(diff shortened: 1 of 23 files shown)


pvighi:
[update link expiration in gateway and email templates](https://trello.com/c/v4HZGSrN/4079-update-link-expiration-in-gateway-and-email-templates)

src/client/pages/EmailSent.tsx
```diff
@@ -56,7 +56,7 @@ export const EmailSent = ({
       )}
       <MainBodyText>Please follow the instructions in this email.</MainBodyText>
       <MainBodyText>
-        <b>The link is only valid for 30 minutes.</b>
+        <b>The link is only valid for 60 minutes.</b>
```
pvighi, line 59: this seems very repetitive. Maybe in the future we can extract this to some common location, either the expiration description or the amount of minutes (if we are sure we are never going to express it in hours or days)

cypress/integration/ete/registration/register.cy.ts
```diff
@@ -209,7 +209,7 @@ describe('Registration flow', () => {
         expect(body).to.have.string('This account already exists');
         expect(body).to.have.string('Sign in');
         expect(body).to.have.string('Reset password');
-        expect(body).to.have.string('This link is only valid for 30 minutes.');
+        expect(body).to.have.string('This link is only valid for 60 minutes.');
```
ob6160, line 212: following a discussion yesterday we decided to drop the "only" from the message copy now the time is longer than 30 mins, so it should now read:

"This link is valid for 60 minutes."

ob6160 (changes requested):

cypress/integration/ete/registration/register.cy.ts
pvighi: I'm not sure what is so wrong about saying that it's "only" valid for 60 minutes that would justify changing it everywhere but if the decision is already made I'll change it

src/client/pages/ResetPasswordSessionExpiredPage.tsx
```diff
@@ -32,7 +32,7 @@ export const ResetPasswordSessionExpiredPage = () => {
       formPageTrackingName="reset-password-session-expired"
     >
       <MainBodyText>
-        The link we sent you was valid for 30 minutes and it has now expired.
+        The link we sent you was valid for 60 minutes and it has now expired.
```
pvighi, line 35: a lot (maybe all) of these `*SessionExpiredPage` pages don't seem to be being used currently. I could test my change by going to the url manually but when the token actually is invalid it just stays in the regular page (password reset in this case) and display a different "session expired" message

cypress/integration/ete/registration/register.cy.ts
pvighi: I'm not really sure how the "changes requested" feature works...
I made the required changes, do you have to click something to make it official that your requested changes were made ?

ob6160: you can re-request a review, or like i'm about to do i can just put a new review through myself 👍

ob6160 (approved):
⏳ 🔢 📧

PR #1858, opened 2022-07-21
Okta | Fix registration emails for users without password trying to register

coldlink:
## What does this change?

- Fixes an issue where a user without a password tries to register (or resend an email after trying to register) where they would receive an email with a link to sign in (AccountExistsEmail), despite them not
having a password to do so.
  - This was fixed by sending the AccountExistsWithoutPasswordEmail instead, which doesn't have a sign in link.
- Refactors the registration with Okta code to share code between the register and resend functionality

### DIFF
--- a/src/server/lib/okta/register.ts
+++ b/src/server/lib/okta/register.ts
@@ -10,6 +10,124 @@ import { OktaError } from '@/server/models/okta/Error';
 import { causesInclude } from '@/server/lib/okta/api/errors';
 import { sendAccountExistsEmail } from '@/email/templates/AccountExists/sendAccountExistsEmail';
 import { sendResetPasswordEmail } from '@/email/templates/ResetPassword/sendResetPasswordEmail';
+import { sendAccountWithoutPasswordExistsEmail } from '@/email/templates/AccountWithoutPasswordExists/sendAccountWithoutPasswordExists';
+
+/**
+ * @name sendRegistrationEmailByUserState
+ *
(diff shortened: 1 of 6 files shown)


src/server/lib/okta/register.ts
```diff
@@ -211,14 +109,63 @@ export const resendRegistrationEmail = async (
...
+ * @method resendRegistrationEmail
+ *
+ * Used by the email sent page if we need to resend the registration email.
+ *
+ * We read the state of the user and send the appropriate email.
+ *
+ * @param {string} email
+ * @returns {Promise<UserResponse>} Promise that resolves to the user object
+ */
+export const resendRegistrationEmail = async (
```
ob6160, line 167: not sure i like the layer of indirection here with the wrapping method.. if both methods do the same job, is there a name that would be able to cover the intent for all use cases instead of having to wrap it like this?

ob6160 (commented):
one comment, otherwise looks good to me

src/server/lib/okta/register.ts
coldlink: Yeah good point. Naming things is hard. 🤔

coldlink: Changed it to `sendRegistrationEmailByUserState`

ob6160: nice 👍

ob6160 (approved):

PR #1872, opened 2022-07-25
Okta | Reset password flows for non-`ACTIVE` Okta user states

coldlink:
## What does this change?

During MVP3 testing it was discovered that the reset password flow was unable to handle non-`ACTIVE` Okta user states, and would either send a "Ignore this email" (Forgot password denied template in Okta), or cause an error to be thrown.

This PR adds the following functionality to the reset password flow for the following user states

- `STAGED` - Send create password email, with activation token, which will set a password and activate their account.
- `PROVISIONED` - Send create password email, with activation token, which will set a password and activate their account.
- `RECOVERY` - Send forgot password email, with recovery token, which will reset a password and put then into the ACTIVE state
- `PASSWORD_EXPIRED` - Send forgot password email, with recovery token, which will reset a password and put then into the ACTIVE state

Cypress mocked and end to end tests are added to cover these scenarios

## Other changes

- Add `RUNNING_IN_CYPRESS_MOCKED` flag to mocked tests, so we don't attempt to send an email through AWS SES, and just mock that it was sent successfully

### DIFF
--- a/cypress/integration/ete-okta/reset_password.cy.ts
+++ b/cypress/integration/ete-okta/reset_password.cy.ts
@@ -1,4 +1,12 @@
+import { Status } from '../../../src/server/models/okta/User';
 import { randomPassword } from '../../support/commands/testUser';
+
+const breachCheck = () => {
+  cy.intercept({
+    method: 'GET',
+    url: 'https://api.pwnedpasswords.com/range/*',
+  }).as('breachCheck');
+};
 
 describe('Password reset flow in Okta', () => {
   context('Account exists', () => {
@@ -15,11 +23,7 @@ describe('Password reset flow in Okta', () => {
       const appClientId = 'appClientId1';
       const fromURI = 'fromURI1';
 
-      cy.intercept({
-        method: 'GET',
-        url: 'https://api.pwnedpasswords.com/range/*',
-      }).as('breachCheck');
-
+      breachCheck();
       cy.createTestUser({
         isUserEmailValidated: true,
       })?.then(({ emailAddress }) => {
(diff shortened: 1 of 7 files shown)


cypress/integration/ete-okta/reset_password.cy.ts
```diff
@@ -1,5 +1,13 @@
+import { Status } from '../../../src/server/models/okta/User';
 import { randomPassword } from '../../support/commands/testUser';
 
+const breachCheck = () => {
```
ob6160, line 4: might be nice to make this a command, then we can benefit from this in all the other tests we have this `intercept` 🙂

cypress/integration/mocked/okta_send_reset_password.cy.ts
```diff
@@ -1,14 +1,172 @@
...
+        errorCauses: [],
+      });
+      cy.mockNext(200, {
+        resetPasswordUrl: `https://${Cypress.env(
+          'BASE_URI',
+        )}/reset_password/token_token_token_to`,
+      });
+      cy.mockNext(200, {
+        stateToken: 'stateToken',
+        expiresAt: new Date(Date.now() + 1800000 /* 30min */),
```
ob6160, line 137: should these be 60 min now we have the new timeout?

ob6160 (commented):
great work, two comments

cypress/integration/ete-okta/reset_password.cy.ts
coldlink: Good idea, but this should probably go into another PR as I don't want to pollute this PR with changes to that.

cypress/integration/mocked/okta_send_reset_password.cy.ts
coldlink: For mocked tests it doesn't actually matter what the expiry time is.

In this case it won't be 60 mins, or 30 mins. It's actually 5 mins 30 seconds as we discovered during the testing, which is what we're fixing in https://github.com/guardian/gateway/pull/1856

ob6160 (approved):

PR #1875, opened 2022-07-27
explicitly add users to the guardianUsers group on creation

pvighi:
## What does this change?

we do have a rule that does this automatically on the Okta side, but it looks like the Okta permissions system isn't smart enough to take that into account and it fails because we don't have permissions to create users outside the designated group.
By adding the group id explicitly to the user creation API we are allowed to execute it even though it's redundant and doesn't really change anything.

## Before merging
we need to add the group id to the configuration


src/server/lib/getConfiguration.ts
```diff
@@ -184,6 +184,12 @@ export const getConfiguration = (): Configuration => {
       process.env.OKTA_CLIENT_SECRET,
       'OKTA client secret missing',
     ),
+    groupIds: {
+      GuardianUserAll: getOrThrow(
+        process.env.OKTA_GUARDIAN_USERS_GROUP_ID,
+        'OKTA_GUARDIAN_USERS_GROUP_ID missing',
+      ),
+    },
```
pvighi, line 192: I't annoying to have to add this to the configuration. I guess the alternative would be to make an extra api call on startup to get the group id from its name. But I don't know if it's worth the trouble

pvighi:
[explicitly add users to GuardianUser-All group from gateway](https://trello.com/c/UACf9wDV/4090-explicitly-add-users-to-guardianuser-all-group-from-gateway)

src/server/models/okta/User.ts
```diff
@@ -78,6 +78,7 @@ export interface UserCreationRequest {
       'email' | 'login' | 'isGuardianUser' | 'registrationPlatform'
     >
   >;
+  groupIds?: Array<string>;
```
ob6160, line 81: In `getConfiguration` we use `getOrThrow` so the application won't start without this information. Does that mean we can make this type non optional?

ob6160 (commented):
looks good — one comment. might be worth letting the team know so they can update their local. did you also update the DEV configuration in s3 to include it?

src/server/models/okta/User.ts
coldlink: This is on the user creation request, technically it's possible to create a user without a groupId, which is why it's optional I think as it matches what the okta documentation says: https://developer.okta.com/docs/reference/api/users/#request-parameters

ob6160: sounds good — i had the wrong context at the time and mistook this for the general configuration type

src/server/lib/getConfiguration.ts
ob6160: agreed, I don't think it'd be worth the trouble in this case either. in my mind it would be more appealing if we started to need multiple group ids in the future (for example, the Jobs user group id could foreseeably be needed)

src/server/models/okta/User.ts
pvighi: we won't have permissions to execute the api call without setting this param, so we could also make it mandatory here and make it a build time error instead of runtime error ?

coldlink: Yeah, it might be a good idea to make it mandatory for that reason 👍🏾

pvighi: I made it mandatory now

coldlink (approved):

PR #1876, opened 2022-07-28
Okta | Ophan | Fire ComponentEvent using componentEventParams

coldlink:
## What does this change?

We rely on end-to-end tracking from frontend/dcr through to the moment that we sign users in/create an account. We use ophan component events for this tracking. For example we track a user click from the Sign in Gate through to whether they signed in or created an account.

Previously we sent the `compoentEventParams` query param through gateway to IDAPI so that IDAPI would handle the call to ophan. However when moving to Okta, gateway needs to be able to handle this instead.

This PR sets up the `ComponentEventParams` type using `zod`, has a method to generate this object from the query param, and additional methods to send this to ophan.

We also send the ophan component events in the closest places to where IDAPI was sending them, i.e on account creation (`CREATE_ACCOUNT`) or sign in/session creation (`SIGN_IN`).

To facilitate some of the changes in this PR, we updated the ophan types in `@guardian/libs`: https://github.com/guardian/libs/pull/375

## Other changes

- Updates `tsconfig` to use `ES2022` so we can use new features!
 - In this PR I use `replaceAll` (from ES2021)
 - We can see from https://node.green/#ES2022 that versions of node >16.10 to <= 17.9.1 support all ES2022 features.
   - We're currently using 16.13 (might be worth updating to latest 16 LTS version in another PR)

### DIFF
--- a/package.json
+++ b/package.json
@@ -128,7 +128,7 @@
     "@guardian/ab-core": "^2.0.0",
     "@guardian/ab-react": "^2.0.1",
     "@guardian/consent-management-platform": "^10.8.0",
-    "@guardian/libs": "^6.0.0",
+    "@guardian/libs": "^6.1.0",
     "@guardian/source-foundations": "^5.1.0",
     "@guardian/source-react-components": "^6.0.0",
     "@guardian/source-react-components-development-kitchen": "^2.0.0",
--- a/src/server/lib/ophan.ts
+++ b/src/server/lib/ophan.ts
@@ -2,17 +2,127 @@ import { OphanEvent, OphanInteraction } from '@/shared/model/ophan';
 import { fetch } from '@/server/lib/fetch';
 import { logger } from '@/server/lib/serverSideLogger';
 import timeoutSignal from 'timeout-signal';
-import { stringify } from 'query-string';
+import { stringify, parse } from 'query-string';
+import { z } from 'zod';
+import {
+  OphanAction,
+  OphanComponent,
+  OphanComponentEvent,
+  OphanComponentType,
+} from '@guardian/libs';
+import serialize from 'serialize-javascript';
 
 const ophanUrl = 'https://ophan.theguardian.com/img/2';
 
+// Component event params is the decoded version of the componentEventParams query parameter
+// defined in the TrackingQueryParams interface from src/shared/model/QueryParams.ts
+const componentEventParamsSchema = z
+  .object({
+    componentType: z.string(),
+    componentId: z.string(),
+    abTestName: z.string().optional(),
+    abTestVariant: z.string().optional(),
+    viewId: z.string(),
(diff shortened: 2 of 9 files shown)


src/server/lib/ophan.ts
```diff
@@ -41,7 +147,56 @@ export const record = (event: OphanEvent, config: OphanConfig = {}) => {
...
+/**
+ * @name sendOphanComponentEventFromQueryParamsServer
+ * @description Send a component event to Ophan from the server side using the componentEventParams query parameter passed in from the frontend
+ *
+ * @param componentEventParamsQuery The componentEventParams query parameter from the frontend
+ * @param action The OphanAction to send to ophan
+ * @param value The value of the component event to send to ophan
+ * @param consentUUID The user consent UUID from the consentUUID cookie
+ */
+export const sendOphanComponentEventFromQueryParamsServer = (
```
ob6160, line 171: it'd be great to get some unit test coverage in place for this method

package.json
```diff
@@ -128,7 +128,7 @@
     "@guardian/ab-core": "^2.0.0",
     "@guardian/ab-react": "^2.0.1",
     "@guardian/consent-management-platform": "^10.8.0",
-    "@guardian/libs": "^6.0.0",
+    "@guardian/libs": "^6.1.0",
```
ob6160, line 131: a link to your changes in guardian/libs in the PR description would be great, just for posterity

src/server/lib/ophan.ts
```diff
@@ -2,17 +2,123 @@ import { OphanEvent, OphanInteraction } from '@/shared/model/ophan';
...
+ * ComponentEventParams type using zod, return undefined if
+ * unable to parse the query string
+ */
+const parseComponentEventParams = (
+  componentEventParams: string,
+): ComponentEventParams | undefined => {
+  const parsedQuery = parse(componentEventParams.replace('undefined', ''));
+
+  const componentEventParamsParsed =
+    componentEventParamsSchema.safeParse(parsedQuery);
```
ob6160, line 55: I wonder if [`safeParseAsync`](https://github.com/colinhacks/zod#safeparseasync) might be a better method to use here so we don't block the main thread, e.g if someone malicious decides to hit us with a lot of data?

ob6160 (commented):
great work, nice use of zod! main thought is that it seems quite key to get at least some test coverage on this

src/server/lib/ophan.ts
coldlink: That's a very good point! Updated to use `safeParseAsync`.

src/server/lib/ophan.ts
coldlink: added in lots of tests!

ob6160 (approved):
nice work, thanks for getting those tests in 🧪

PR #1888, opened 2022-08-03
MVP2 | Add metrics for determining user state on sign in

coldlink:
## What does this change?

As part of MVP2 we want to know the number of users who specifically when signing in have validated emails, but weak passwords. This information will enable us to weigh in on whether the functionality to address this needs to be put in place now or whether we can carry on with weak passwords in Okta.

To help with this this PR looks at all 4 possible states the user can be in and adds an AWS metric on the state that the user is in.

The 4 states are:

1. User has Email Validated and has a strong password
2. User has Email Validated and has a weak password
3. User has not Email Validated and has a strong password
4. User has not Email Validated and has a weak password

This maps to the following 4 metrics respectively:

1. `User-EmailValidated-StrongPassword`
2. `User-EmailValidated-WeakPassword`
3. `User-EmailNotValidated-StrongPassword`
4. `User-EmailNotValidated-WeakPassword`

This PR only adds the metrics for these states, and doesn't take any actions based on the state.

For each sign in with Okta, we get the user groups, check if the user is in the `GuardianUser-EmailValidated` group, as well as checking their password against the breached passwords API.

We don't log/track the user's email or password, nor would we ever want to do this!

### DIFF
--- a/src/server/routes/signIn.ts
+++ b/src/server/routes/signIn.ts
@@ -30,6 +30,7 @@ import { redirectIfLoggedIn } from '../lib/middleware/redirectIfLoggedIn';
 import { getUserGroups } from '../lib/okta/api/users';
 import { clearOktaCookies } from '@/server/routes/signOut';
 import { sendOphanComponentEventFromQueryParamsServer } from '../lib/ophan';
+import { isBreachedPassword } from '../lib/breachedPasswordCheck';
 
 const { okta, accountManagementUrl, oauthBaseUrl, defaultReturnUri } =
   getConfiguration();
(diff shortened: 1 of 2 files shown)


raphaelkabo:
An idle thought: why do we namespace some of the metrics with `::`, and some with `-` like here?

coldlink:
> An idle thought: why do we namespace some of the metrics with `::`, and some with `-` like here?

There isn't really a reason for this. The convention in gateway is to add `Success` or `Failure` after the `::` (Conditional Metrics). And use `-` if it's not.

src/server/routes/signIn.ts
```diff
@@ -291,17 +292,38 @@ const oktaSignInController = async (
...
+      // retrieve the user groups
       const groups = await getUserGroups(response._embedded.user.id);
+
+      // check if the user has their email validated based on group membership
       const emailValidated = groups.some(
         (group) => group.profile.name === 'GuardianUser-EmailValidated',
       );
+
+      // check the user password strength
+      const hasWeakPassword = await isBreachedPassword(password);
```
pvighi, line 305: I don't how precise we need the metrics to be but in case of error I think this defaults to false

pvighi (approved):
looks good to me

src/server/routes/signIn.ts
coldlink: Since this is just for getting a comparison, and it's very rare that the breached password api to be down, this should be okay in this scenario. Long term it shouldn't affect our numbers.

ob6160 (approved):

PR #1890, opened 2022-08-03
Disable useOkta flag

raphaelkabo:
Some time ago, we introduced a param `useOkta`, which when present in a query would authenticate the request against Okta rather than the legacy Identity API.

This PR makes that param the default option, thus making the param redundant.  The logic will now follow the same path as if the `useOkta` param were present, by default.

It also introduces a new param, `useIdapi`, which when true authenticates against the Identity API instead of Okta.  We think this will be a useful way of rooting out any differences in the way the different systems operate.

Still to do:

- [ ] Remove the `useOkta` param from email queries

Co-authored-by: Kelvin Chappell <1722550+kelvin-chappell@users.noreply.github.com>


coldlink:
I suggest leaving this PR as a draft PR until we're nearly ready with it to go live!

src/server/routes/register.ts
```diff
@@ -68,102 +68,11 @@ router.post(
   '/register/email-sent/resend',
   handleRecaptcha,
   handleAsyncErrors(async (req: Request, res: ResponseWithRequestState) => {
-    const { useOkta } = res.locals.queryParams;
-    if (okta.enabled && useOkta) {
-      await OktaResendEmail(req, res);
+    const { useIdapi } = res.locals.queryParams;
+    if (useIdapi) {
```
coldlink, line 72: probably a bit early to review this pr, but it will be useful to keep the existing switch in `okta.enabled` so we can disable okta through a switch

so we should do something like
```suggestion
    if (!okta.enabled || useIdapi) {
```

kelvin-chappell: Good idea!  Done in 3b71c141d700091db6e04b7613c69a6e932e747b

cypress/integration/mocked/okta_change_password.2.cy.ts
```diff
@@ -133,7 +133,7 @@ describe('Change password in Okta', () => {
       mockPasswordResetInvalidStateTokenFailure();
       mockValidateRecoveryTokenFailure();
 
-      cy.visit(`/reset-password/token?useOkta=true`);
+      cy.visit('/reset-password/token?useIdapi=true');
```
coldlink, line 136: The tests in this file relate to okta, so they shouldn't have the `useIdapi=true` flag

coldlink, line 136: The tests in this file relate to okta, so they shouldn't have the `useIdapi=true` flag

coldlink:
```suggestion
      cy.visit('/reset-password/token');
```

coldlink (commented):
some slight changes to the mocked tests 😄

coldlink:
My only other comment is that there are 40+ commits in this PR, which is quite a lot! Would it be possible to interactively rebase this PR and squash some commits together so they become easier to follow?

See Step 4 in https://github.com/guardian/gateway#contributing

Give me a shout if you need hand in doing this. 😄

kelvin-chappell:
Thanks @coldlink Still going through the test flags so it's quite possible that I'm going flag-blind.  I'll squash it into some more meaningful commits once all tests are passing.

coldlink (approved):
Amazing work!

raphaelkabo:
Looks great to me, well done Kelvin!

dominickendrick:
🎉 !

PR #1891, opened 2022-08-04
feat: add a `request_id` with middleware and to logger

coldlink:
## Accessibility


src/shared/lib/baseLogger.ts
```diff
@@ -1,21 +1,28 @@
+/* eslint-disable @typescript-eslint/no-explicit-any */
 import { LogLevel, Logger } from '@/shared/model/Logger';
 
+export interface ExtraLogFields {
+  request_id?: string;
```
ob6160, line 5: is there anywhere we *don't* want to track the `request_id`? wondering if we don't make it required, future additions to the codebase could forget to include it

coldlink: There are some logs which happen outside the context of a request, so we can't add a request id there, which means we can't make this a required property.

Not sure of the best way to enforce this for a request though where we do want to include it 🤔

ob6160: the first thing that comes to mind is wrapping it in a method for use in request contexts i suppose, although i hesitate to suggest that level of abstraction

coldlink: Yeah, we could make abstract the server side logger further into a "RequestLogger" which lives on the request/request state object, but that may cause even more confusion on it's usage.

ob6160: My thoughts:
So the risk here is that new log entries added to requests in the future don't include the uuid mistakenly. this would mean that we lack the functionality which might be frustrating when debugging, but once noticed could be added back easily.

in my opinion that's an acceptable risk when considered against the potential cost of adding complexity

ob6160 (approved):

src/shared/lib/baseLogger.ts
coldlink: Yeah I agree with that, in most cases developers will look at prior use cases of the code and see the request id being used.

PR #1892, opened 2022-08-04
Okta | Fix reset password for user not found

coldlink:
## What does this change?

In #1872 we fixed a bunch of errors for users in the non ACTIVE state for users going through the reset password flow, but a regression was introduced for users without an account requesting password reset.

In this case an error would be shown, rather than the email sent page. We want to show the email sent page for users without accounts too who go through the password reset flow to prevent this page from being used for account enumeration (checking if an account exists).

This PR fixes this by handling this case explicitly, and adds a cypress mocked test to check this behaviour.

### DIFF
--- a/src/server/controllers/sendChangePasswordEmail.ts
+++ b/src/server/controllers/sendChangePasswordEmail.ts
@@ -252,6 +252,23 @@ const sendEmailInOkta = async (
       }),
     );
   } catch (error) {
+    // handle check for if the user is not found
+    // this is a special case because we don't want to send an email
+    // but we want to show the email sent page
+    // so the reset password page cannot be used for account enumeration
+    if (
+      isOktaError(error) &&
+      error.status === 404 &&
+      error.code === 'E0000007'
+    ) {
+      return res.redirect(
+        303,
+        addQueryParamsToPath('/reset-password/email-sent', state.queryParams, {
+          emailSentSuccess: true,
+        }),
+      );
+    }
+
     logger.error('Okta send reset password email failed', error, {
       request_id: res.locals.requestId,
     });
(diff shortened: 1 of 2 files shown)


src/server/controllers/sendChangePasswordEmail.ts
```diff
@@ -252,6 +252,23 @@ const sendEmailInOkta = async (
...
     );
   } catch (error) {
+    // handle check for if the user is not found
+    // this is a special case because we don't want to send an email
+    // but we want to show the email sent page
+    // so the reset password page cannot be used for account enumeration
+    if (
+      isOktaError(error) &&
+      error.status === 404 &&
+      error.code === 'E0000007'
```
ob6160, line 262: do we have these codes mapped to an enum (or similar somewhere?) would be nice to have them a little bit more human readable.. e.g: `OktaErrorCodes.ResourceNotFoundError` for this one

ob6160 (approved):
looks good - one comment

src/server/controllers/sendChangePasswordEmail.ts
coldlink: There is

https://github.com/guardian/gateway/blob/a3801031c60d067675d92de30e82a869a7d85a7a/src/server/models/okta/Error.ts#L29-L37

But it's not quite useful as a typescript enum, so maybe thats worth looking into for a future improvement PR!

ob6160: will make a ticket for the 'ol backlog

ob6160:
[Health/Enhancement: Put Okta error codes into an Enum so we can access them in a human readable manner](https://trello.com/c/etQgKvie/4113-health-enhancement-put-okta-error-codes-into-an-enum-so-we-can-access-them-in-a-human-readable-manner)

PR #1893, opened 2022-08-04
Generate JSON schema for rate limiter configuration file

ob6160:
## What does this change?
We use the `zod-to-json-schema` module to generate a JSON schema on the `postinstall` hook of the project. This will help us write the configuration; giving us intellisense code completion on `.ratelimit.json` and `.ratelimit.example.json` as well as increased confidence that we are writing a valid configuration file.

This file is called `.ratelimit.schema.json` and is not tracked by version control, because it will be liable to change as the rate limiter configuration schema is modified over time.

The command `yarn gen:ratelimit:schema` creates the schema file, by running the `zodToJsonSchema` method on the rate limiter configuration schema using `ts-node` to transpile the typescript files.

## Demonstration
https://user-images.githubusercontent.com/1771189/182916733-5653dc64-2f8d-4d3d-95b0-e93b85b5b1f0.mov

### Notes
* The import path to the Routes file was changed to: `../../../shared/model/Routes` from a `@/*` path alias because of the way that module resolution works in `ts-node`.

### DIFF
--- a/scripts/generate-rate-limit-schema.ts
+++ b/scripts/generate-rate-limit-schema.ts
@@ -1,11 +1,18 @@
+import { writeFileSync } from 'node:fs';
 import zodToJsonSchema from 'zod-to-json-schema';
 import { rateLimiterConfigurationSchema } from '../src/server/lib/rate-limit/configurationValidator';
 
+process.stdout.write('✨  Generating rate limit configuration schema  ✨\n');
+
+// Generate a JSON schema for the rate limiter configuration.
 const jsonSchema = zodToJsonSchema(
   rateLimiterConfigurationSchema,
   'rateLimiterConfigurationSchema',
 );
 
-// Output the JSON schema for the rate limiter configuration.
-// We use this for autocompletion when editing the rate limit configuration file.
-process.stdout.write(JSON.stringify(jsonSchema) + '\n');
+// Write the pretty printed JSON schema to .ratelimit.schema.json.
+// We use this for autocompletion when editing the rate limit configuration file (and example).
+writeFileSync(
+  '.ratelimit.schema.json',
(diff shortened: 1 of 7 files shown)


scripts/generate-rate-limit-schema.ts
```diff
@@ -0,0 +1,18 @@
...
+  rateLimiterConfigurationSchema,
+  'rateLimiterConfigurationSchema',
+);
+
+// Write the pretty printed JSON schema to .ratelimit.schema.json.
+// We use this for autocompletion when editing the rate limit configuration file (and example).
+writeFileSync(
+  '.ratelimit.schema.json',
+  JSON.stringify(jsonSchema, null, 2) + '\n',
+);
```
coldlink, line 18: there isn't any typescript specific stuff going on here, could this just be a `.js` file and we just use `node` rather than having to introduce `ts-node` as a dependency

i know `ts-node` is good, but maybe it's a bit overkill for this use case.

coldlink: if we were able to introduce ts-node to a wider part of this project, then i'd be all over it 😅

ob6160: i added `ts-node` because i'm importing `configurationValidator.ts` which has multiple typescript dependencies, couldn't think of a cleaner way to do it

ob6160: as it's just a dev dependency, it doesn't seem too bad to add it in - what d'you think?

coldlink: Yeah in agreement now, and we can move some of our other `js` scripts to `ts` by using `ts-node` :)

coldlink (approved):
👍🏾 
time to start moving our other scripts too to `ts-node`

PR #1894, opened 2022-08-04
Okta | Fix welcome resend and register resend for user not found

coldlink:
## What does this change?

If on the welcome link expired page (and the register email sent page resend functionality) a user entered an email for an account that doesn't exist in Okta the following error would occur:

https://user-images.githubusercontent.com/13315440/182950596-320cd3c1-1e7d-45bd-8399-acb6836f8de6.mov

This PR fixes this issue by utilising the same functionality as the register page for this form, i.e use the same controller as register on the welcome link expired page.

https://user-images.githubusercontent.com/13315440/183031991-a927366a-c5d9-4fbb-a235-02635379c90e.mov

Also added a bunch of tests for this welcome link expired page.

### DIFF
--- a/cypress/integration/ete-okta/registration.2.cy.ts
+++ b/cypress/integration/ete-okta/registration.2.cy.ts
@@ -680,4 +680,293 @@ describe('Registration flow', () => {
       });
     },
   );
+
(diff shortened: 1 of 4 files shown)


ob6160 (approved):
nice catch!

PR #1895, opened 2022-08-05
Refactor: Sign In and Register pages to use `MainLayout` and `MainForm`

coldlink:
## What does this change?

Refactors the `SignIn` and `Registration` pages to use the `MainLayout` and `MainForm` components used by many of the other pages, e.g. reset password.

This also refactors the `Nav` component so that the tabs are generated in that file. And the component itself is used on the `MainLayout`. We also add the `errorReportUrl` to the `ErrorSummary` component in the `MainLayout`.

I've also gone and removed unused exports, variables, and methods.

There are some Chromatic changes, due to the difference in spacing between the old components/pages and the `MainLayout` component, however the spacing is now more in line with everything else using the `MainLayout` component.

We also update all the Cypress tests that relied on the previous components too.

This PR will also help moving to the [`Islet`s](#1513) architecture in the future by cleaning up the implementation of client side code.

| - | Before | After |
| - | - | - |
| Register | Screenshot 2022-08-10 at 15 29 07 | capture (1) |
| Sign In | capture (3) | capture (2) |


ob6160 (commented):
code changes look great — nice work — it's a lot neater

two comments before approving:

* I think we should definitely get design sign-off on those chromatic diffs and how it behaves at different viewports. More than anything because the original pages were designed to a spec which is now changing, this may need to be reflected in Figma.
* Just checking that this has been tested in CODE and it all looks good there

raphaelkabo (approved):
I think this is great - a big difference to code structure and readability, and improved visual consistency.

PR #1906, opened 2022-08-09
Email validation in Okta

ob6160:
## Background
We require a way to validate users' emails through Okta without requiring them to also reset their password.

This PR adds documentation describing how we can achieve this using the [Okta Factors API](https://developer.okta.com/docs/reference/api/factors/#enroll-okta-email-factor) to validate against an email factor.


ob6160:
[Spike (aka preliminary research card): Enforce email validation for all users with strong password signing in (on web) (M)](https://trello.com/c/F6WppUpZ/4093-spike-aka-preliminary-research-card-enforce-email-validation-for-all-users-with-strong-password-signing-in-on-web-m)

coldlink:
After a discussion in the Identiteam chat this morning, we no longer need to do email validation, as the number of users this would affect was minuscule, and not worth the work to implement this. For unvalidated users we're instead going to put them down the path of "Reset Password" so that they're email gets validated while still getting a strong password set for all users.

PR #1913, opened 2022-08-10
Allow individual routes to be placed in "log only" mode.

ob6160:
## What does this change?
We'll be using the log only mode soon once we start to issue Okta sessions to signed in dotcom users.

This PR aims to ease that process. Instead of having to put the whole limiter into log only mode, it allows us to selectively enable / disable it for specific routes. So in this scenario, we would only need to put `/signin/refresh` into log only mode, allowing the rate limiter to still be effective for all of the other routes we want to protect.

As an example, this configuration would look like this:

```json
{
  "enabled": true,
  "settings": {
    "logOnly": false,
    "trackBucketCapacity": false
  },
  "defaultBuckets": {
    "globalBucket": { "capacity": 500, "addTokenMs": 50 },
    "ipBucket": { "capacity": 100, "addTokenMs": 50 },
    "emailBucket": { "capacity": 100, "addTokenMs": 50 },
    "oktaIdentifierBucket": { "capacity": 100, "addTokenMs": 50 },
    "accessTokenBucket": { "capacity": 100, "addTokenMs": 50 }
  },
  "routeBuckets": {
    "/signin/refresh": {
      "settings": {
        "logOnly": true
      },
      "globalBucket": {
        "addTokenMs": 10,
        "capacity": 100
      }
    }
  }
}
```

## How to test

- [ ] Deploy to CODE and put a single route into log only mode. Check to see that it is not rate limited and logs are still captured.

### DIFF
--- a/docs/rate-limit/configuration.md
+++ b/docs/rate-limit/configuration.md
@@ -40,7 +40,8 @@ If the Redis connection fails, the server refuse to serve any requests until the
 
 ## Entry: `settings`
 
-An optional object which lets you enable some extra debug functionality
+An optional object which lets you enable some extra debug functionality.
+The `settings` configuration object is also available for individual routes, so you can override the global settings selectively on a route-by-route basis.
 
 - When `logOnly` is set to `true`, the rate limiter logic will still run, but the result will not be enforced. This is useful for when you want to try out some new configuration values but you don't want to actually apply them yet.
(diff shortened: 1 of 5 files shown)


coldlink (approved):
📝

raphaelkabo:
😍

PR #1920, opened 2022-08-18
Add mocked Okta tests for account enumeration and form submission edge cases

raphaelkabo:
## What does this change?

This adds a set of new Cypress tests which run against mocked Okta endpoints. It also adds the infrastructure to allow these tests to run in GitHub Actions like our other Cypress tests.

The tests are in the `cypress/integration/mocked-okta` folder, and are separated into files by the general type of controller which handles the same kinds of requests in Gateway.

In each testing file, we loop through a set of [Okta user statuses](https://help.okta.com/en-us/Content/Topics/users-groups-profiles/usgp-end-user-states.htm?cshid=ext_end_user_states), testing each against form submission on a specific Gateway route. We mock the responses from the Okta API to control the inputs, increase speed, and prevent flakiness.

States to routes matrix: https://docs.google.com/spreadsheets/d/1Zy259LBeJCTzJ7QhckbO0fVTdbVoqTfyMsVZ0l-vcMs/edit#gid=0

**NB**: I've added a new environment variable, `CYPRESS_ENCRYPTION_SECRET_KEY`, to the GitHub Actions env file and the local env file for mocked Cypress testing. This is because `Cypress.env()` is only able to read environment variables prefixed by `CYPRESS_` if they're not set within Cypress via another method: https://docs.cypress.io/api/cypress-api/env. We need to read the encryption secret key value in the `setEncryptedStateCookie` Cypress function, as `Cypress.env('ENCRYPTION_SECRET_KEY')`.

**To do**

- [x] Documentation for the routes vs. Okta statuses matrix
- [x] Add tests for the `ACTIVE (social only)` status 
- [x] Add tests for the three `/set-password/*` routes


cypress/integration/mocked-okta/registerController.1.cy.ts
```diff
@@ -0,0 +1,131 @@
...
+                          const response = { ...userResponse.response, status };
+                          cy.mockNext(
+                            userExistsError.code,
+                            userExistsError.response,
+                          );
+                          cy.mockNext(userResponse.code, response);
+                          cy.mockNext(
+                            tokenResponse.code,
+                            tokenResponse.response,
+                          );
```
pvighi, line 80: Maybe it's because I'm not very familiar with the project but this type of mocking makes it kind of hard for me to follow the test.

As far as I understand we are just mocking all http requests and relying on the order of the requests to provide the correct mocked response, which is also pretty brittle as any change in the sequence of calls breaks the test and in some cases it can be hard to understand what each mocked response was originally meant for.

Could we use `mock_pattern` like [this](https://github.com/guardian/gateway/blob/main/cypress/integration/mocked/okta_register.3.cy.ts#L18-L36) instead ? It looks like you can provide at least the url you are mocking ( but not the http method as far as I can see) which makes more clear what is being mocked and it would make the test more robust.

If we cannot do that maybe a comment before each `mockNext` could make it easier to follow.

Of course there's a chance I misunderstood how all of this works and this whole comment doesn't make sense

raphaelkabo: I totally agree - as we discussed in the call, I think it would be worth doing the following:

- Add comments to all the intercepts explaining why each mock is being used, what URL is being called, and what function is triggering that URL request.
- Consider using `mockPattern` instead to make it even more clear which URLs are being mocked, although I thought that it was worthwhile continuing to intercept single requests in order as it would make it very clear if a change modified the order of requests in the actual code.

cypress/integration/mocked-okta/signInController.2.cy.ts
```diff
@@ -0,0 +1,135 @@
...
+        cy.visit(`/signin?useOkta=true`);
+        cy.get('input[name="email"]').type('example@example.com');
+        cy.get('input[name="password"]').type('password');
+      });
+      switch (state) {
+        case false:
+        case 'PROVISIONED':
+        case 'STAGED':
+        case 'RECOVERY':
+        case 'PASSWORD_EXPIRED':
```
pvighi, line 21: I think in a few places these tests are kind of misleading. 
For example, here we are generating 4 different tests that all are exactly the same and labelling them as:

>  Given that I am a **[nonexistent/provisioned/staged/recovery/password_expired]** user when I submit the form on /signin Then I should be shown an "Email or password incorrect" error ". 

Unless I'm misunderstanding something this test should only run a single time and be labelled something like 

> 'When I submit the form on /signin' I should be shown an "Email or password incorrect" error if Okta authentication fails' 

Nowhere in the test we are validating that non existent users or users in states other than ACTIVE actually cause Okta to return an authentication failed error.

I would make sure all the tests are labelled in a way that makes it very clear what it is we are verifying. We can add to the test description or comments when we expect to encounter these situations just to get some context.

src/server/lib/encryptedStateCookie.ts
```diff
@@ -35,9 +35,17 @@ export const setEncryptedStateCookie = (
 export const readEncryptedStateCookie = (
   req: Request,
 ): EncryptedState | undefined => {
-  const encryptedCookie = baseUri.includes('localhost')
-    ? req.cookies[encryptedStateCookieName]
-    : req.signedCookies[encryptedStateCookieName];
+  // TODO: We need to double check the security implications of using an env variable
+  // to read from the unsigned cookie storage.
```
pvighi, line 39: @coldlink this is the TODO I was talking about, also I noticed this is the only change in the PR that doesn't just affect documentation or tests so we probably want to make sure we are happy with this.

cypress/integration/mocked-okta/registerController.1.cy.ts
```diff
@@ -0,0 +1,91 @@
...
+            () => {
+              // Set the correct user status on the response
+              const response = { ...userResponse.response, status };
+              cy.mockNext(userResponse.code, response);
+              cy.get('button[type=submit]').click();
+              verifyInRegularEmailSentPage();
+            },
+          );
+          break;
+        case 'ACTIVE':
```
pvighi, line 35: I don't know what we would do here to test register with social users ? @coldlink  do we need to mock pushing the apple, facebook and google button ?

pvighi, line 35: I don't know what we would do here to test register with social users ? @coldlink  do we need to mock pushing the apple, facebook and google button instead of submitting the form ?

src/server/lib/encryptedStateCookie.ts
coldlink: Ah the environment variable it's referring to I think is the `RUNNING_IN_CYPRESS` one. Which is fine, as long as we make sure that isn't set on any production environment (which is isn't currently). The encryption secret keys are always stored in either github action or aws parameter store, and we don't use the same keys for any environment.

src/server/lib/encryptedStateCookie.ts
```diff
@@ -35,9 +35,17 @@ export const setEncryptedStateCookie = (
...
+  const encryptedCookie =
+    baseUri.includes('localhost') || process.env.RUNNING_IN_CYPRESS === 'true'
```
coldlink, line 41:
```suggestion
    process.env.RUNNING_IN_CYPRESS === 'true'
```

I don't think we need the `localhost` check here, as we never want to run gateway on localhost

src/server/lib/encryptedStateCookie.ts
coldlink: So I think this is fine.

cypress/integration/mocked-okta/registerController.1.cy.ts
coldlink: I think the social users case we're talking about here is if a social only user already exists in the okta user store in the ACTIVE state but without a password.

So we're talking about an existing user attempting to register with an email that already exists. In this case the ACTIVE case covers both the social and non-social existing accoubnt

pvighi: I could do what I did with other tests and add a second version of the test but use a social user json response. The only difference is the contents of the social part of the json that I don't think anything checks right now.

src/server/lib/encryptedStateCookie.ts
pvighi: ok we can remove that, but it isn't being added by this PR.
What about [encryptedStateCookieOptions](https://github.com/guardian/gateway/blob/4c574d435a5eef78fc6528547b9050ed6bc4289b/src/server/lib/encryptedStateCookie.ts#L13-L14) in this same file? it's also checking that baseUri includes localhost there.. 
Would this logic have to match the encryptedStateCookieOptions ? 
I'm a bit confused because otherwise this could have checked  `encryptedStateCookieOptions.secure` or `encryptedStateCookieOptions.signed` instead of look at the baseUri directly, right ?

cypress/integration/mocked-okta/registerController.1.cy.ts
pvighi: I ended up adding the social user case like I did everywhere else to complete the way the rest of the tests are written. I guess it validates that the code on our side wouldn't do anything different if the user response has different credentials

src/server/lib/encryptedStateCookie.ts
```diff
@@ -35,9 +35,17 @@ export const setEncryptedStateCookie = (
...
+      ? // If we're in testing or localhost, first try reading from signedCookies,
+        // and only then fall back to regular cookies.
+        Object.keys(req.signedCookies).includes(encryptedStateCookieName)
+        ? req.signedCookies[encryptedStateCookieName]
+        : req.cookies[encryptedStateCookieName]
+      : // If we're not in testing, always read from signedCookies.
+        req.signedCookies[encryptedStateCookieName];
```
pvighi, line 48: I find this ternary operator to be super confusing for anything other the most trivial cases. In particular here where there's 2 nested. Wouldn't it be better to just use regular if/else statements or is there a reason to do it this way ?

cypress/integration/mocked-okta/signInController.2.cy.ts
raphaelkabo: I think for these labels I went with a user-story approach rather than a test case approach - you're probably right in that the test case description would be more useful in this case. That means we could also collapse some identically performing tests down as @coldlink suggested.

src/server/lib/encryptedStateCookie.ts
raphaelkabo: I agree, the `localhost` check there doesn't seem to add any information. I think we should keep [the check in `encryptedStateCookieOptions`](https://github.com/guardian/gateway/blob/7f805dd1a982c98d8557be182edf6cd65391cfbc/src/server/lib/encryptedStateCookie.ts#L11-L16) though - don't Cypress mocked tests run Gateway on `localhost`? I suppose it also prevents the possibility of trying to set a secure cookie on an insecure connection. Then again, from the [cookie module docs](https://www.npmjs.com/package/cookies#readme):

> An error will be thrown if you try to send secure cookies over an insecure socket.

src/server/lib/encryptedStateCookie.ts
raphaelkabo: Happy for this to be broken into something more readable!

PR #1932, opened 2022-08-26
Okta | Fix `POST /set-password/resend` not handling Okta

coldlink:
## What does this change?

Fixes an issue where the resend functionality on the `/set-password`(Create password) pages when a link is expired would only send emails from IDAPI and not Okta.

This was because the controller used in the create password page, the `POST /set-password/resend` route, was only handling IDAPI, and not Okta.

This PR refactors this route so that is uses the  `sendChangePasswordEmailController/sendEmailInOkta` that `/reset-password` does for Okta, as the functionality between both should be identical. This also refactors the controller a little to be able to handle `/set-password` urls, along with `/reset-password` urls.


pvighi (approved):
the code looks good to me. maybe we could put it in code and retest the password create and password reset cases ?

coldlink:
@kelvin-chappell this may affect and require changes in #1890 after this PR is merged in

PR #1939, opened 2022-09-05
Dependencies | September 5th 2022

coldlink:
## What does this change?

Changes from previous #1918 :

Holds back the following:
* Holds back upgrading `@guardian/` components as they're major version updates which broke the build for some reason

## Tested

- [x] CODE
- [x] IE11


src/server/lib/__tests__/headers.test.ts
```diff
@@ -108,7 +108,6 @@ describe('Content Security Policy headers', () => {
           'o14302.ingest.sentry.io',
         ].join(' '),
       );
-      expect(splitCSPHeader).toContain('block-all-mixed-content');
```
kelvin-chappell, line 111: Is this no longer useful or is it no longer being generated?

kelvin-chappell (approved):
👍

src/server/lib/__tests__/headers.test.ts
coldlink: its no longer being generated by helmet as it's been deprecated https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/block-all-mixed-content

kelvin-chappell: Thanks!

PR #1942, opened 2022-09-07
Clear Okta cookie in IDAPI only controllers

coldlink:
## What does this change?

When going live with MVP3 (releasing Okta sessions to production), there will still be some existing endpoints in Gateway still using IDAPI as the backend when using the `useIdapi=true` flag or otherwise, for example the welcome page for new guest users from contributions/subscriptions/support, reset password emails sent from the admin panel etc.

These endpoints all set new identity cookies, but have no awareness of a new Okta session, with no way to set one. So to be on the safe side in these cases we want to clear the Okta session should one exist, so that user sessions don't diverge.


kelvin-chappell (approved):
👍

pvighi (approved):
looks good to me

PR #1944, opened 2022-09-14
Remove useOkta flag from Okta signin script

raphaelkabo:
## What does this change?

We no longer need the `useOkta` flag to be set in the JS script we attach to the default Okta signin gateway to redirect the user to our custom gateway. This simply removes it.

Mahesh:
- Fixes failing Okta tests due to changes in the query parameters


kelvin-chappell (approved):
I remember seeing this and then forgetting about it 👍

PR #1960, opened 2022-09-20
Restore resend email functionality

raphaelkabo:
## What does this change?

Some unintentional recent changes meant that the `OktaResendEmail` function stopped attempting to register users whenever they requested an email resend. This meant that an edge case could be encountered where Gateway attempted to resend an email for an account that did not exist, which would result in an error message being shown to the user. The preferable, earlier functionality meant that there could never be a situtation where a user account did not exist during the resend email flow.

This simply restores the earlier functionality, and adds a comment explaining why this logic is beneficial.


kelvin-chappell (approved):
👍

PR #1961, opened 2022-09-20
Send reset password emails for unvalidated email accounts

raphaelkabo:
This establishes a new flow for users attempting to **sign in** or **register with an existing account** whose email address is not validated (i.e. they have the `emailValidated` flag set to false in their Okta profile and are thus not in the `GuardianUsers-EmailValidated` Okta group). These are all legacy users whose email was never validated in IDAPI for whatever reason.

On **sign in**, a user with an unvalidated email will be navigated to the new page `/signin/email-sent`, without being logged in. We will generate a reset password token for them and send them a specific email which asks them to reset their password in line with our new password policy. As a side-effect of the reset password process, which ensures that they set a strong password, this also validates their email address.

On **register with an existing account**, the backend flow is the same, but the user is simply shown the `/register/email-sent` page, as in any other registration scenario. This is to prevent account enumeration attacks.

As well as keeping track of unvalidated emails, we also keep track of accounts with weak passwords. This new flow enforces password resets for users with weak and strong passwords alike, as long as their emails are unvalidated.

## What does this change?

- Create a new React page and refactor the `EmailSent` component so the new page can extend the original component.
- Add  a `/signin/email-sent` route and a `POST /signin/email-sent/resend` route.
- Add new behaviour to the signin and register routes which checks for an unvalidated email and takes action accordingly. 
- Add a new method, `sendEmailToUnvalidatedUser`, which is used by both new flows.
- Add an email template for the new email.
- Add Cypress and Jest tests for the new flows
- Add metrics for the new flows
- Add documentation for password reset techniques in Gateway (there are now three, each for a different scenario)


src/client/pages/UnvalidatedEmailEmailSentPage.tsx
```diff
@@ -0,0 +1,50 @@
...
+
+  return (
+    <EmailSent
+      email={email}
+      changeEmailPage={buildUrl('/signin')}
+      resendEmailAction={buildUrl('/signin/email-sent/resend')}
+      queryString={queryString}
+      showSuccess={emailSentSuccess}
+      errorMessage={error}
+      noAccountInfo={noAccountInfo}
```
coldlink, line 41: It doesn't seem you're using `noAccountInfo`

So could this just be set to false or not included and be removed from the props?

```suggestion
```

src/email/templates/UnvalidatedEmailResetPassword/UnvalidatedEmailResetPasswordText.ts
```diff
@@ -0,0 +1,14 @@
+export const UnvalidatedEmailResetPasswordText = () => `
+Hello,
+Because your security is extremely important to us, we have changed our password policy. For this reason, you need to reset your password.
```
coldlink, line 3:
```suggestion
Hello,

Because your security is extremely important to us, we have changed our password policy. For this reason, you need to reset your password.
```

coldlink (approved):
Looking really good!

PR #1962, opened 2022-09-21
Upgrade @guardian/libs to the latest version

ob6160:
## What does this change?

We upgrade the `@guardian/libs` package to [^8.0.5](https://github.com/guardian/csnx/releases/tag/%40guardian%2Flibs%408.0.5)

As part of this upgrade one of our `react-testing-library` tests for MainForm started to fail. After looking into this issue, the problem was narrowed down to this change: https://github.com/guardian/libs/pull/390/files#diff-e66f30416e1d836f8f29281cd56fa1d9e3f1c9f7fd97b25a3019c8183006be9fR34 which I believe meant that that the mocked reCAPTCHA script was no longer inserted into the page properly due to `document.scripts[0]` being undefined until the `useRecaptcha` hook is loaded and subsequently evaluates the `loadScript` method. 

The fix was to ensure that we mock the script insertion for each test in a `beforeEach` method for every test. This should ensure that the script is in the page *before* the test begins, avoiding a circumstance where `document.scripts[0]` is undefined by the time `loadScript` is called.

There could be more at play here around how the test environment is bootstrapped, causing a possible race condition — but I didn't want to spend too much time digging into the why as the fix works as intended and isn't hacky.

This PR also includes a timeout bump for the supertest suite around the rate limiter because in some cases it runs slowly inside the GitHub actions environment — if this keeps getting slower and slower it'll need some deeper investigation most likely

### DIFF
--- a/package.json
+++ b/package.json
@@ -132,7 +132,7 @@
     "@guardian/ab-core": "^2.0.0",
     "@guardian/ab-react": "^2.0.1",
     "@guardian/consent-management-platform": "^10.11.1",
-    "@guardian/libs": "^6.2.0",
+    "@guardian/libs": "^8.0.5",
     "@guardian/source-foundations": "^5.3.0",
     "@guardian/source-react-components": "^7.1.1",
     "@guardian/source-react-components-development-kitchen": "^3.1.10",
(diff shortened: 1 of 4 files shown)


coldlink (approved):
All the 😍

PR #1963, opened 2022-09-22
Okta | Jobs changes

coldlink:
## What does this change?

To make the Jobs migration easier, I investigated ways that I could make changes to the jobs flows to Gateway so that it was easier for Jobs to integrate with our changes.

Namely we wanted to facilitate the showing of the `/agree/GRS` page automatically if the user was coming from jobs, as well as setting the `clientId` and `returnUrl` parameters automatically.

The new jobs flow works in a similar way to the native apps flow, Jobs will try to perform the Authorization Code flow, if a session doesn't exist they'll be redirected to the Okta hosted login page, which we intercept and show our own sign in/registration page instead on Gateway.

The first change adds the old `clientId=jobs` query parameter at the point of intercept if the calling OAuth application is `jobs_site`, as well as the `returnUrl=` pointing to the jobs subdomain on the correct environment. These get passed through to Gateway.

On Gateway when a user gets an Okta session set, we look for the `clientId=jobs` query parameter, if this is the case then we redirect the user to `/agree/GRS`. This page already handles any checks that it needs to make, if a user is already a valid jobs user it will redirect them straight back to jobs, otherwise it will show the T&Cs page.

Since we need to complete the Authorization Code flow if initiated by the jobs site, we pass around the `fromURI` parameter, to `/agree/GRS` so that if this parameter exists, once the user accepts the T&Cs or is already a jobs user, they automatically complete the Authorization Code flow when we redirect them to the `fromURI`.

This allows us to keep the existing functionality as it is for production, but also adds a way that we can complete the Jobs flows directly in a purely OAuth2 world going forward in the future.

## Tested
- [x] CODE

## Related
https://github.com/guardian/identity-platform/pull/571

### DIFF
--- a/cypress/integration/ete-okta/jobs_terms.4.cy.ts
+++ b/cypress/integration/ete-okta/jobs_terms.4.cy.ts
@@ -1,4 +1,26 @@
 describe('Jobs terms and conditions flow in Okta', () => {
+  context('Shows the terms and conditions page on Sign In', () => {
+    it('visits /agree/GRS after sign in if clientId=jobs parameter is set', () => {
+      cy.intercept('GET', 'https://jobs.theguardian.com/', (req) => {
+        req.reply(200);
+      });
+
+      cy.createTestUser({
+        isUserEmailValidated: true,
+      })?.then(({ emailAddress, finalPassword }) => {
+        const visitUrl =
+          '/signin?clientId=jobs&returnUrl=https%3A%2F%2Fjobs.theguardian.com%2F';
+        cy.visit(visitUrl);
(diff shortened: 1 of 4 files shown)


ob6160 (commented):
great work — really neat solution to the problem :)

as discussed separately it would be great to see some unit tests for `okta-login.js` as a future addition so we have confidence around making larger changes like this

This could perhaps be achieved through moving the source for this file to easily tested typescript + es6 and then running it through the adequate transpilation steps to create a final output file for use in Okta?

kelvin-chappell (approved):
Looks good to me 👍

PR #1964, opened 2022-09-22
Enforce RequestState type on deepmerge

raphaelkabo:
- `mergeRequestState` is a wrapper around `deepmerge` but with enforced types, which became necessary when we noticed that for an unknown reason, the TypeScript checker wasn't picking up type errors on usages of vanilla `deepmerge`.
- We also ended up fixing a couple of type bugs resulting from `deepmerge` not having been typechecked for a couple of months.
- Split the 'previousPage' prop in `pageData`, used by both the consent rotues and the resend email routes, into two more descriptive props, `resendEmailAction` and `changeEmailPage`, for the resend routes only. The consent routes still use `previousPage` as it makes sense there. (This is part of this ticket because the `pageData` prop is being set in the places where we were originally using `deepmerge`).


kelvin-chappell (approved):
👍

PR #1966, opened 2022-09-26
Fix cookie mismatch issue causing user to appear logged out on Dotcom

coldlink:
## What does this change?

A bug was discovered where if the Identity cookies (specifically `SC_GU_U` and `GU_U`) cookies didn't exist in the user's browser, but the Okta session cookie did (`sid`). It would look like that the user was logged out on dotcom.

When the user would go click on the "Sign In" button, it would seem like the page would refresh, and nothing would happen, the user would still seem to be logged out.

We determined this was caused by the `redirectIfLoggedIn` middleware where it would only check for the presence of the Okta cookie, and not the identity cookies, and redirect the user back to the homepage as long as the okta cookie was valid.

So instead of redirecting the user straight away, we put the user through the Authorization Code flow in Gateway, which will refresh their Okta cookie, and set new identity cookies to the same expiry. We make sure to set the `doNotSetLastAccessCookie` parameter so that the `SC_GU_LA` cookie isn't set, as this is used only when the user manually authenticated again.

The long term solution for this is to show a page when a user navigates to a path which uses the `redirectIfLoggedIn` middleware which shows the user that is currently logged in with an option to "Continue to the Guardian" or log out and sign in as a different user. But this won't be worked on for a little while yet, and we want to fix this issue while users are being affected.

We also refactor the `GET /signin` controller to remove a redundant session check, as this was already being handled by the `redirectIfLoggedIn` controller.

Also fixed related cypress tests relating to this change.

https://user-images.githubusercontent.com/13315440/192265392-8c93158d-1f2b-4f83-b610-5172ce395346.mov

### DIFF
--- a/cypress/integration/mocked/okta_register.3.cy.ts
+++ b/cypress/integration/mocked/okta_register.3.cy.ts
@@ -42,6 +42,7 @@ describe('Okta Register flow', () => {
       cy.visit('/register');
 
       setSidCookie();
+
       cy.get('input[name="email"]').type('example@example.com');
       cy.mockNext(200, {
         userType: 'new',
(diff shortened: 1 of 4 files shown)


raphaelkabo (approved):
Solid fix! ✅

PR #1967, opened 2022-09-28
Always use Okta for social sign in

coldlink:
## What does this change?

- Updates the `/signin/:social` to always use Okta if the switch is enabled, disregarding the `useIdapi` flag.
  - Social users behave the same whether in Okta or the legacy system, so there's no need to switch depending on the flag, and the only reason we'd want to turn it off is if we disable the Okta switch altogether
  - This should potentially stop all traffic going to our [legacy Identity Federation API](https://github.com/guardian/identity-federation-api)
  - This will allow us to determine if there is any traffic still going to that platform, and may allow us to turn it off sooner rather than later
  - We still get traffic using the `useIdapi` flag from some of the emails we still send from IDAPI. While none of them necessarily go to `/signin`, because the `useIdapi` flag is persistable (so persists between routes), it's possible to navigate to `/signin` and then click one of the social buttons theoretically

### DIFF
--- a/src/server/routes/signIn.ts
+++ b/src/server/routes/signIn.ts
@@ -486,15 +486,14 @@ const isValidSocialProvider = (provider: string): boolean =>
 router.get(
   '/signin/:social',
   handleAsyncErrors(async (req: Request, res: ResponseWithRequestState) => {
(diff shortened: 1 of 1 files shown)


kelvin-chappell (approved):
👍

PR #1968, opened 2022-09-29
`SignedInAs` page and functionality

coldlink:
## What does this change?

We're changing the functionality on how certain pages on gateway behave if you already have a valid Okta session set in your browser. Specifically the `/signin` and `/register` pages.

Previously when you visited these pages and had a valid Okta session set in your browser you would automatically get redirected back to the `returnUrl` or homepage by default.

Normally it wouldn't be possible to visit these pages when logged in, unless through a stray link, or navigating directly to one of these pages in your browser. However this is changing as part of our migration. Namely when using the Okta SDK (or any other way to initiate an OAuth flow), a call to `/authorize` would have to be made, if the user was not logged in, then the sign in page would be shown. This is opposed to how it currently works where we sign in a user first before attempting to get any tokens for that user.

This means that it's possible to visit the sign in page or register page as part of this flow if you're already logged in, and it would redirect you automatically back to the calling application whether native or otherwise.

This has changed now so that we show a `SignedInAs` page, so that when you visit these pages on gateway behave if you already have a valid Okta session, we show a page which tells you which user (email) you're logged in as, as well as an option to "Continue" which finishes the OAuth flow/redirects the user to returnUrl depending on context, and an option to "Sign in as a different user" which first signs the user out, and then shows the sign in page again, all while preserving any query parameters required.

There are two ways that this flow is initiated
1. The user navigates directly, or through a stray link, to `/signin` or `/register` while signed in
2. The user navigates through an OAuth `/authorize` call which prompts the user to sign in even if the user has a session (`?prompt=login`)

The behaviour in each of these two cases will differ slightly. 

In 1. the continue link will be the `returnUrl` which is most likely the article or front that the user accessed these pages from. While the signout link will sign the user out, show the sign in screen, and preserve any query parameters from before sign out.

In 2. the continue link will be the `fromURI` parameter which we get from the Okta hosted sign in page interception script, where we redirect the user to our own sign in page. This parameter encodes the URL to complete the OAuth flow for that given request. So clicking continue will complete the OAuth flow for the given user session if they're logged in. The sign out link will sign the user out, show the sign in screen, and preserve any query parameters from before sign out, which includes the `fromURI` parameter, allowing the user to complete the OAuth flow if they then proceed to sign in as a different user.

We also display slight variants in the page design. If we detect if the the user is coming from a native app (the `appClientId` parameter is either the `ios` or `android` app) we show a slight variation of this page. This functionality will also be useful down the line when we need to show different pages or a flow based on if the user is coming from the app.

### Continue

https://user-images.githubusercontent.com/13315440/194082645-625bd764-dbd4-4319-8394-64fca76260da.mov

### Sign in with different email

https://user-images.githubusercontent.com/13315440/194330777-8df160ff-e97a-483a-8909-1f3d152b20ba.mov

## Application changes

Any application which is using the OAuth authorization code flow to log users in should now pass the `prompt=login` parameter to the `/authorize` url in order to use this new functionality. This tells Okta not to silently authenticate a user if a user session exists, and allows us to show our Signed In As page or the normal Sign In or Register page depending on if the user is logged in or not. 

## Related

https://github.com/guardian/identity-platform/pull/575

## Note for reviewers

This PR may be easier to review commit by commit as each change is an individual step to get all the parts working.
Also there are the new chromatic changes to review too.


scripts/okta/okta-login.ts
```diff
@@ -24,9 +24,16 @@ import { getRedirectUrl } from './lib/helper';
 
 // When modifying these files, be sure to code defensively, as we want to handle any errors or unexpected flows that may occur.
 
+// The Okta hosted sign in page doesn't have just one url which it executes on.
+// We've noticed the following two urls (paths):
+// - /login/login.htm
+// - /oauth2/<authorization server id>/v1/authorize
+// each will have a different query string and different config objects which are handled as appropriate in helper.ts.
```
kelvin-chappell, line 31: 😖

kelvin-chappell (approved):
Seems pretty watertight to me as things stand.  And the tests make it clear what's expected.  Good solution! 👍

PR #1970, opened 2022-09-29
Autopopulate registrationLocation field for new users registering via Okta

vlbee:
## What does this change?

Captures a new user's `registrationLocation` in Okta at the time they register for an account, provided they have accepted all CMP consents on the the device they are registering with.

- Adds a [hidden input field](https://github.com/guardian/gateway/blob/ee3b4d017f531c24d773f2b3a8657fd6ddec4c3a/src/client/components/CmpConsentStateHiddenInput.tsx) to the registration form - the value of which is the CMP consented state (Defaults to "false")
- This is submitted along with a user's email for registration via Okta only (not IDAPI)
- The request also includes the Fastly-set `GU_geo_country` cookie. 
- The request hits POST /register which uses a new utilitiy called [`getRegistrationLocation` ](https://github.com/guardian/gateway/blob/ee3b4d017f531c24d773f2b3a8657fd6ddec4c3a/src/client/components/CmpConsentStateHiddenInput.tsx)to derive a users location from the cookie and hidden input field value. This is then submitted to Okta and returned on the User Response (although not used for anything in gateway)

The regions are based on the first columen of this [Data Design list](https://github.com/guardian/data-platform-models/blob/main/dbt/data/seed_country.csv)

## Why?

Earlier this year we added a new user profile field to Okta called the registrationLocation. This is the country/region of a user (independatnt of any billing/correspondence address) which they can self-manage via Manage My Account > Settings. With this PR we will start auto-populating the field for new, consented users. 

The 'registrationLocation' field is stored in a user's Okta profile.

## Notes / Questions: 

- The field is returned by Okta on the UserResponse for a unit test, although not sure this is neccessary as not used otherwise.
- Would it make sense to also capture this location field on the `Set Password` page after a new user has validated their email and is completing the registration process? This would be an extra net, in case they did not have CMP consent on the first page when submitting email. 

#### IN CODE:

- [x] A new user registers via /register page with cmp consent in GB -> `United Kingdom` appears in Okta profile
- [x] A new user registers via /register page without cmp consent in GB -> field empty in Okta profile
- [x] A new user registers via /register page with cmp consent in US -> `United States` appears in Okta profile
- [x] A new user registers via /register page without cmp consent in US -> field empty in Okta profile


src/server/routes/register.ts
```diff
@@ -111,10 +112,13 @@ const OktaRegistration = async (
   req: Request,
   res: ResponseWithRequestState,
 ) => {
-  const { email = '' } = req.body;
-  try {
-    const user = await registerWithOkta(email);
+  const { email = '', _cmpConsentedState = false } = req.body;
 
+  const registrationLocation: RegistrationLocation | undefined =
+    getRegistrationLocation(req, JSON.parse(_cmpConsentedState));
```
coldlink, line 118: Just wondering why the `JSON.parse` is needed here?

Is `_cmpConsentedState` not a boolean normally?

vlbee: It was coming up as a "boolean" 
I believe formdata can only be string or blob  - https://developer.mozilla.org/en-US/docs/Web/API/FormData

src/client/components/Terms.tsx
```diff
@@ -4,7 +4,7 @@ import { space, textSans } from '@guardian/source-foundations';
 import { ExternalLink } from '@/client/components/ExternalLink';
 
 export const termsContainer = css`
-  margin-top: ${space[5]}px;]};
+  margin-top: ${space[5]}px;
```
vlbee, line 7: Prettier error. I thought this was fixed in main? I don't see any issue on the accual sign in page though.

src/server/lib/__tests__/okta/register.test.ts
```diff
@@ -108,14 +109,19 @@ jest.mock('@/server/lib/trackMetric', () => ({
...
+  registrationLocation: RegistrationLocation | undefined = undefined,
+): UserResponse => {
   return {
     id: 'someuserid',
     status: status,
     profile: {
       login: email,
       isGuardianUser: true,
       email: email,
+      registrationLocation: registrationLocation,
```
vlbee, line 124: Is this neccessary? not being used for anything other than the server unit test.

src/server/routes/register.ts
coldlink: I suggest using this method then: https://github.com/guardian/gateway/blob/main/src/server/lib/queryParams.ts#L5

This converts a string to a boolean, and it's safer than using JSON.parse (as the input is coming from the client is untrusted, and JSON.parse can be unsafe)

coldlink: As something similar has caused issues with gateway in the past: https://github.com/guardian/gateway/pull/1317

src/server/lib/__tests__/okta/register.test.ts
raphaelkabo: I suppose if `registrationLocation` can end up in a user response it might be good to keep it in the mocked response in case of unexcpected side-effects?

src/client/components/CmpConsentStateHiddenInput.tsx
```diff
@@ -0,0 +1,19 @@
...
+
+export const CmpConsentedStateHiddenInput = ({
+  cmpConsentedState,
+}: CmpConsentedStateProps) => {
+  return (
+    <>
+      <input
+        type="hidden"
+        name="_cmpConsentedState"
+        value={String(cmpConsentedState)}
```
coldlink, line 15: This can be a template string instead.

```suggestion
        value={`${cmpConsentedState}`}
```

src/server/lib/getRegistrationLocation.ts
```diff
@@ -0,0 +1,279 @@
...
+  if (country === 'CA') return 'Canada';
+  if (Europe.includes(country)) return 'Europe';
+  if (Row.includes(country)) return 'Other';
+
+  return undefined;
+};
+
+const Us: (CountryCode | undefined)[] = ['US', 'AS', 'GU', 'MP', 'PR', 'VI'];
+
+// XK Kosovo removed
```
coldlink, line 35: 👀 very minor in terms of code, very major in terms of political ramifications

I'd like to understand better why `VG` and `XK` have been removed, as they seem like a strange omission, and `removed to validate type checking` doesn't make much sense to me.

coldlink:
my only other suggestion is to clean up the commit history as well if possible, as a few commits looks like they can be merged in with other commits, e.g. `fix` commits

src/server/lib/getRegistrationLocation.ts
raphaelkabo: This interested me so I went digging!

Looks like this is the origin type for `CountryCode`: https://github.com/guardian/csnx/blob/main/libs/%40guardian/libs/src/countries/%40types/CountryCode.ts

It derives from [ISO 3166-1 alpha-2](https://en.wikipedia.org/wiki/ISO_3166-1_alpha-2). That standard _does_ have `VG` for Virgin Islands, so that's a straight up omission on the part of the Guardian libs type. It _doesn't_ have Kosovo, which I believe is because it's not listed in any of the UN sources which make up the list, which is ultimately because it's only partially recognised.

raphaelkabo: > XK is a "user assigned" ISO 3166 code not designated by the standard, but used by the [European Commission](https://en.wikipedia.org/wiki/European_Commission), Switzerland, the [Deutsche Bundesbank](https://en.wikipedia.org/wiki/Deutsche_Bundesbank) and other organisations. However, [ISO 3166-2:RS-KM](https://en.wikipedia.org/wiki/ISO_3166-2:RS) remains in use.

raphaelkabo: I see no reason why we wouldn't want to include `XK` it in our libs - we're not trying to adhere to strict ISO 3166 as far as I'm aware!

coldlink: At least I don't think we should mention that we explicitly left `XK` out, as that may not reflect well on the whole organisation.

coldlink: I guess this thread is the context 🤷🏾

vlbee: I've raised a PR to add VG in the library but will leave XK off for now unless that team chooses to add it to the library as unsure if it needs strict ISO3166 adherence or not -  https://github.com/guardian/csnx/pull/147

src/server/routes/register.ts
vlbee: Moved isStringBoolean to it's own lib file and using instead 👍

src/server/lib/__tests__/okta/register.test.ts
vlbee: Sorry, linked to the wrong file. I added it to the Okta Response in the **User Model** purely to add a server unit test. It is  not being used in gateway no shown back to the user, nor does it need to be returned in the user response. 

However, we may want to keep in the UserResponse model purely so that the response reflects data that has been updated in Okta? I have left it in for this reason.

coldlink (approved):
Looking good to me!

PR #1982, opened 2022-10-03
Switch order of social and email/password components on SignIn and Registration pages

raphaelkabo:
## What does this change?

We're switching the order of the social login buttons the and email/password components on the `SignIn` and `Registration` pages. Previously the social buttons were at the bottom, now they'll be on top.

## Register page

|                          | **Before** | **After** |
|--------------------------|------------|-----------|
| First load | register-before-first | register-after-first |
| With error | register-before-error | register-after-error |

## Sign in page

|  | **Before** | **After** |
|---|---|---|
| First load | siginin-before-first | signin-after-first |
| With error | signin-before-error | signin-after-error |
| Error, no social buttons | signin-before-no-social | signin-after-no-social |

As a side-effect of these changes, a couple of small changes have been made:

- Minor changes to margins across pages using the `MainForm` component. This is because the `MainForm` component had a large margin at the bottom, which presumably was only there to push the bottom of the page down, as there aren't any pages that (1) use `MainForm` and (2) have any elements _below_ the bottom of `MainForm`. I don't _think_ we need this, since it has no real purpose and simply pushes the page down further than its preset 900px minimum height.
- Also on the note of margins, I've standardised some mysterious margins that sat between the bottom of the `MainForm` inputs and the terms/reCAPTCHA texts. I believe they're now consistent!
- I've added tests to ensure T&Cs appear correctly on `SignIn` and `Registration`, as well as the vanilla `MainForm`. This is mostly for my own peace of mind as I had to refactor the logic for when the terms text appears. This is because I needed to split the terms text in two: the T&Cs and privacy policy text needs to appear **above the first interactable button on the page**, so the social buttons when those appear and the submit button otherwise; and the reCAPTCHA text has to always appear above the submit button.
- `MainForm` now handles reCAPTCHA (and, in future, any other submit-generated) errors in two ways. If it receives a setter function, it will pass the error up through the function to the parent component. If it doesn't receive a setter function, it will instead display the error at the top of the `MainForm` component itself. This is so the error associated with a set of email/password inputs, for instance a reCAPTCHA error, will appear directly above _those_ inputs, and not above the social buttons, at the top of `MainForm`'s parent component.
- So that the `MainForm` Storybook stories don't break, I've added in a check which doesn't display a reCAPTCHA error if the reCAPTCHA token is set to `"test"`. Otherwise, the stories would immediately show a reCAPTCHA error.
- I've removed an unused component in `Terms.tsx`, and reorganised the terms texts a little. I also fixed some grammatical mistakes after confirming with the PM.
- I've added a couple of useful Storybook stories and renamed a couple of others to better encapsulate our use cases.

### DIFF
--- a/cypress/integration/ete/registration/register.2.cy.ts
+++ b/cypress/integration/ete/registration/register.2.cy.ts
@@ -65,7 +65,7 @@ describe('Registration flow', () => {
         req.reply(200);
       });
       cy.visit('/signin?clientId=jobs&useIdapi=true');
-      cy.contains("Guardian's Jobs terms & conditions").click();
+      cy.contains('Guardian Jobs terms & conditions').click();
       cy.url().should('eq', guardianJobsTermsOfServiceUrl);
     });
 
@@ -79,7 +79,7 @@ describe('Registration flow', () => {
       });
       cy.visit('/signin?clientId=jobs&useIdapi=true');
       cy.contains('For information about how we use your data')
-        .contains("Guardian Jobs' privacy policy")
+        .contains('Guardian Jobs privacy policy')
         .click();
       cy.url().should('eq', guardianJobsPrivacyPolicyUrl);
     });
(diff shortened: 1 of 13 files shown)


coldlink (approved):
Reviewed the changes, it all looks okay to me, as with the chromatic diffs too.

PR #1983, opened 2022-10-04
Handle missing password for unvalidated user attempting to register

raphaelkabo:
## What does this change?

[This PR](https://github.com/guardian/gateway/pull/1961) prevented users from being able to signin if they had an unvalidated email, and sent them a specific, new sort of email.

In implementing this, we failed to correctly handle a specific case:

```
Given I am an ACTIVE user
And given I have an unvalidated email
And given I don't have a password set
    When I attempt to register
        Then ???
```

This is an edge case scenario, possibly only for users who were imported into Okta from Identity, as `ACTIVE` users cannot have a missing password in regular Okta. This PR handles this state properly:

1. In `sendRegistrationEmailByUserState()`, it  moves the check for an unvalidated email into the `ACTIVE` user case, rather than checking for it before any of the status checks. This is because in all the other user states, we would want to send a regular reset password email instead of the special security one.
2. Inside that case, we check if the user has a password set before we attempt to generate a reset password token for them, as we can't generate a reset password token for a user without a password. If the user **doesn't** have a password, we use the same flow that we use in `sendChangePasswordEmail.ts` to:
    1. Dangerously reset their password, acquiring a reset password token and putting them into `RECOVERY` state
    2. Use that token to set a random placeholder password, returning them to `ACTIVE` state
    3. Finally, send the special security email now that the user has a password set.

```
Given I'm an ACTIVE user
   When I try to register
   Given I don't have a validated email
       Given I have a password set
           Then I should be sent an email with a reset password token
       Given I do NOT have a password set
           Then I should have my password dangerously reset, and then
           sent an email with a reset password token
   Given I have a validated email
       Then I should be sent an email with NO reset password token
   And my status should remain ACTIVE
```

### DIFF
--- a/src/server/lib/okta/register.ts
+++ b/src/server/lib/okta/register.ts
@@ -1,3 +1,4 @@
+import crypto from 'crypto';
 import { Status, UserResponse } from '@/server/models/okta/User';
 import {
   createUser,
@@ -15,6 +16,10 @@ import { sendAccountWithoutPasswordExistsEmail } from '@/email/templates/Account
 import { getConfiguration } from '@/server/lib/getConfiguration';
 import { sendEmailToUnvalidatedUser } from '@/server/lib/unvalidatedEmail';
 import { trackMetric } from '@/server/lib/trackMetric';
+import {
+  validateRecoveryToken,
+  resetPassword as authenticationResetPassword,
+} from './api/authentication';
 
 const { okta } = getConfiguration();
 
@@ -40,21 +45,6 @@ export const sendRegistrationEmailByUserState = async (
   const user = await getUser(email);
   const { id, status } = user;
 
-  // First, check if the user's email is unvalidated. If not, we send
-  // an email asking them to change their password, then return from here
-  // to show the user the regular post-registration 'email sent' page.
-  const groups = await getUserGroups(id);
-  // check if the user has their email validated based on group membership
-  const emailValidated = groups.some(
-    (group) => group.profile.name === 'GuardianUser-EmailValidated',
-  );
-  if (!emailValidated) {
-    await sendEmailToUnvalidatedUser(id, user.profile.email);
-    trackMetric('OktaUnvalidatedUserResendEmail::Success');
-    return user;
-  }
-
-  // The user's email is validated - continue with the user status flows.
   switch (status) {
     case Status.STAGED: {
       /* Given I'm a STAGED user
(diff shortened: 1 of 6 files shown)


src/server/lib/okta/register.ts
```diff
@@ -109,13 +99,62 @@ export const sendRegistrationEmailByUserState = async (
...
-        to: user.profile.email,
-      });
+
+      const doesNotHavePassword = !user.credentials.password;
+
+      const groups = await getUserGroups(id);
+      // check if the user has their email validated based on group membership
+      const emailValidated = groups.some(
+        (group) => group.profile.name === 'GuardianUser-EmailValidated',
+      );
```
pvighi, line 119: why do we get the user groups here ? can't we just check the `emailValidated` boolean flag on the user ? instead of making an additional api call ?

raphaelkabo: We use the groups as our source of truth, because there could be a situation where a user is removed from a group by an admin or new rule, but still has the flag set on their profile. And because the flags don't actually mean anything to Okta, but groups do (in the sense of allowing users to do certain things)

coldlink: Ah I found the actual reason in the description for this commit: https://github.com/guardian/gateway/commit/255ccbebd2dc78824836caff9428dd7401418ea3

coldlink: > We use the groups api to get the group membership, rather than checking the email validated field because
> social users do not have the email validated flag set until after they login for the first time (but are in the email validated group)

coldlink: So there as edge cases where the user might have the group but not the flag, hence the groups should be the source of truth for this!

coldlink: A question you might have is then how do we reconcile these two things?

Well, when the user finishes logging in, and in the OAuth callback route on gateway we check for this, and update the user profile if appropriate:

https://github.com/guardian/gateway/blob/main/src/server/routes/oauth.ts#L162-L182

coldlink: But it's still best to use the groups rather than the field to determine if the user is in the correct state!

pvighi: would this be a problem in other cases ? I mean can we safely rely on the flag everywhere else ?

pvighi: wouldn't this violate the group rule we set in okta ? Or does the rule only enforce that people in the group have the flag set to true but allows adding people to the group with the flag to false ? 🤔

pvighi (approved):
looks good to me

src/server/lib/okta/register.ts
raphaelkabo: I think it's probably best to always rely on the group as a matter of course...

PR #2000, opened 2022-10-12
Improve display of errors on clientside

raphaelkabo:
## What does this change?

🎉 PR #2000!

Building on work previously done in #1982, this PR fixes some bugs in the locations that error mesages were being displayed in the `Main` and `MainForm` components.

- Serverside, we add a new field `RequestState.pageData.formError`. This passes a **form-level** error string from the server to the client. The **page-level** error and success message is, as before, passed in `RequestState.globalMessage` - this doesn't change.
- Most of the error messages from the server are now passed in `formError` instead of `globalMessage`. The only ones which are not being passed this way are in the onboarding flow, which uses its own components.
- Clientside, we extract `formData` from the client state and pass it down into the `MainForm` component by way of the `formErrorMessageFromParent` prop, where it is rendered.
- If an error message is generated in `MainForm` itself - this currently only happens due to a reCAPTCHA error - we replace any error message passed down with the locally generated one.
- Meanwhile, `globalMessage` errors and success messages are passed in all relevant components to `Main`, where they are displayed above any children components.
- We also added a few more tests and Storybook stories to ensure consistency in error display.

| Error type | Generated |    Server prop     | Client displayed |
|------------|-----------|--------------------|------------------|
| Page-level | On server | `globalMessage`      | `<Main>`           |
| Form-level | On server | `pageData.formError` | `<MainForm>`       |
| reCAPTCHA  | On client | -                  | `<MainForm>`       |

Some screenshots of the updated error locations:


raphaelkabo:
[Improve error handling in client forms](https://trello.com/c/Bc3dy1yK/4179-improve-error-handling-in-client-forms)

coldlink (approved):

PR #2008, opened 2022-10-17
Replace express-recaptcha with custom reCAPTCHA middleware

raphaelkabo:
## What does this change?

We have been using the NPM module [express-recaptcha](https://www.npmjs.com/package/express-recaptcha) exclusively to handle validation of  our V2 reCAPTCHA on the serverside. express-recaptcha is quite a large module with a lot of features - it handles reCAPTCHA V2 and V3; and links into rendering engines to display the reCAPTCHA clientside code as well as dealing with reCAPTCHAs on the backend.

We've previously encountered a problem where, on development environments, the express-recaptcha middleware would run the Express route handler for a route multiple times, causing 'headers cannot be sent' errors and in one case (cc. @vlbee) activating an Okta user multiple times in local dev.

This PR removes express-recaptcha entirely, and sets up a new, much simpler middleware, which simply follows Google's instructions for validating a reCAPTCHA token. This allows us to use only basic tools to accomplish this task and doesn't involve a complex external library. It also solves the multiple route handler calls problem in dev environments, which means we might be able to locally test reCAPTCHA in future if we ever feel the need.

Obligatory xkcd:

### DIFF
--- a/src/server/lib/recaptcha.ts
+++ b/src/server/lib/recaptcha.ts
@@ -46,10 +46,6 @@ const handleRecaptcha = async (
   _: Response,
   next: NextFunction,
 ): Promise<void> => {
-  // Skip the reCAPTCHA check if we're running in a test environment
-  if (process.env.RUNNING_IN_CYPRESS === 'true') {
-    return next();
-  }
   const recaptchaToken = req.body['g-recaptcha-response'];
   // If the recaptcha response is missing entirely, throw an error which will be shown to the user.
   if (!recaptchaToken) {
--- a/src/server/lib/recaptcha.ts
+++ b/src/server/lib/recaptcha.ts
@@ -1,18 +1,14 @@
 import { NextFunction, Request, Response } from 'express';
-import { RecaptchaV2 } from 'express-recaptcha';
 import { CaptchaErrors } from '@/shared/model/Errors';
 import { getConfiguration } from '@/server/lib/getConfiguration';
 import { logger } from '@/server/lib/serverSideLogger';
 import { trackMetric } from '@/server/lib/trackMetric';
 import { HttpError } from '@/server/models/Error';
-import { featureSwitches } from '@/shared/lib/featureSwitches';
+import { fetch } from '@/server/lib/fetch';
 
 const {
-  googleRecaptcha: { secretKey, siteKey },
-  stage,
+  googleRecaptcha: { secretKey },
 } = getConfiguration();
-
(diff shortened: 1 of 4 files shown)


raphaelkabo:
[Replace express-recaptcha with simpler custom reCATPCHA middleware](https://trello.com/c/D9YfTdh8/4183-replace-express-recaptcha-with-simpler-custom-recatpcha-middleware)

src/server/lib/recaptcha.ts
```diff
@@ -1,74 +1,87 @@
...
+    {
+      method: 'POST',
+      headers: {
+        'Content-Type': 'application/x-www-form-urlencoded',
+      },
+      body: `secret=${secretKey}&response=${recaptchaToken}`,
+    },
+  );
+  const recaptchaResponseJson =
+    (await recaptchaResponse.json()) as RecaptchaAPIResponse;
```
coldlink, line 65: my only suggestion would be to wrap any `await` calls in a `try`/`catch` block, in case there are any runtime errors, e.g. if the fetch call fails/can't parse the json etc.

src/server/lib/recaptcha.ts
```diff
@@ -1,74 +1,94 @@
 import { NextFunction, Request, Response } from 'express';
-import { RecaptchaV2 } from 'express-recaptcha';
 import { CaptchaErrors } from '@/shared/model/Errors';
 import { getConfiguration } from '@/server/lib/getConfiguration';
 import { logger } from '@/server/lib/serverSideLogger';
 import { trackMetric } from '@/server/lib/trackMetric';
 import { HttpError } from '@/server/models/Error';
-import { featureSwitches } from '@/shared/lib/featureSwitches';
```
coldlink, line 8: It's also worth removing the `recaptchaEnabledDev` key from the features switches object.

coldlink: https://github.com/guardian/gateway/blob/main/src/shared/lib/featureSwitches.ts

raphaelkabo:
This is now passing tests without the recaptcha check needing to be skipped in tests. Looks like we had a particularly flaky test environment yesterday.

coldlink (approved):
Wondering if this would be possible to test first on CODE but with real site key/secret so we can make sure that it's behaving correctly?

raphaelkabo:
Thanks, @coldlink, I've now tested this on CODE and it seems to work just fine! Will merge next week.

PR #2009, opened 2022-10-17
Native app fow: change social CTA button labels and position

raphaelkabo:
## What does this change?

**When accessing Gateway from a native app:**

- The social buttons now all say 'Continue with X' instead of just 'X' for the name of the social auth provider.
- Apple will show at the top and Google at the bottom on iOS; Google will show at the top and Apple at the bottom on Android.
- The buttons no longer collapse to a row on desktop widths; they stay stacked vertically.

On a regular flow, without native app being set, the buttons have not changed.

Screenshot 2022-10-17 at 15 00 36

### DIFF
--- a/src/client/components/SocialButtons.tsx
+++ b/src/client/components/SocialButtons.tsx
@@ -9,18 +9,30 @@ import {
 } from '@guardian/source-react-components';
 import { QueryParams } from '@/shared/model/QueryParams';
 import { buildUrlWithQueryParams } from '@/shared/lib/routeUtils';
+import { IsNativeApp } from '@/shared/model/ClientState';
 
-type SocialButtonProps = {
+type SocialButtonsProps = {
   queryParams: QueryParams;
   marginTop?: boolean;
+  isNativeApp?: IsNativeApp;
 };
 
-const containerStyles = (marginTop = false) => css`
+type SocialButtonProps = {
+  label: string;
+  icon: React.ReactElement;
+  socialProvider: string;
+  queryParams: QueryParams;
+  showGap: boolean;
+  isNativeApp?: IsNativeApp;
+};
+
+const containerStyles = (marginTop = false, isNativeApp: IsNativeApp) => css`
   display: flex;
   flex-direction: column;
-  ${from.tablet} {
-    flex-direction: row;
-  }
+  ${!isNativeApp &&
+  `${from.tablet} {
+      flex-direction: row;
+  }`}
   justify-content: center;
   margin-top: ${marginTop ? space[5] : 0}px;
   ${from.mobileMedium} {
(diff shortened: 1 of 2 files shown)


raphaelkabo:
Converting to draft in anticipation for https://github.com/guardian/gateway/pull/2010 being merged in - we will be putting the social CTA changes behind the `isNativeApp` flag from that PR.

src/client/components/SocialButtons.tsx
```diff
@@ -47,75 +59,94 @@ const iconOverrides = css`
...
+  isNativeApp,
+}: SocialButtonsProps) => {
+  let buttonOrder;
+  if (isNativeApp === 'ios') {
+    buttonOrder = ['apple', 'facebook', 'google'];
+  } else if (isNativeApp === 'android') {
+    buttonOrder = ['google', 'facebook', 'apple'];
+  } else {
+    buttonOrder = ['google', 'facebook', 'apple'];
+  }
```
coldlink, line 128: This should probably be a method which returns the button order based on the `isNativeApp` parameter, rather than reassigning a variable. Makes it testable with a unit test too 😉 

```ts
const getButtonOrder = (isNativeApp?: IsNativeApp): string[] => {
  switch (isNativeApp) {
    case 'ios':
      return ['apple', 'facebook', 'google'];
    case 'android':
      return ['google', 'facebook', 'apple'];
    default:
      return ['google', 'facebook', 'apple'];
  }
}

// then it becomes
export const SocialButtons = ({
  queryParams,
  marginTop,
  isNativeApp,
}: SocialButtonsProps) => {
  const buttonOrder = getButtonOrder(isNativeApp);
...
```

src/client/components/SocialButtons.tsx
```diff
@@ -47,75 +59,94 @@ const iconOverrides = css`
...
+          key={socialProvider}
+          label={socialProvider}
+          icon={
+            socialProvider === 'google' ? (
+              <SvgGoogleBrand />
+            ) : socialProvider === 'apple' ? (
+              <SvgAppleBrand />
+            ) : (
+              <SvgFacebookBrand />
+            )
```
coldlink, line 142: this is a bit hard to read, i suggest something similar to the `getButtonOrder` there should be a method which takes a social provider and returns an icon

the default shouldn't be facebook either, the default should probably be no icon

src/client/components/SocialButtons.tsx
raphaelkabo: Much more functional and stops me from doing weird stuff to get around the immutable array checker, thanks a bunch 😁

src/client/components/SocialButtons.tsx
coldlink: `<></>` is fine!

src/client/components/SocialButtons.tsx
```diff
@@ -47,75 +59,106 @@ const iconOverrides = css`
...
+};
+
+const getButtonOrder = (isNativeApp?: IsNativeApp): string[] => {
+  switch (isNativeApp) {
+    case 'ios':
+      return ['apple', 'facebook', 'google'];
+    case 'android':
+      return ['google', 'facebook', 'apple'];
+    default:
+      return ['google', 'facebook', 'apple'];
```
coldlink, line 139: not the point and probably not for this PR, but i don't think facebook should be in the middle, it should be last 😂

coldlink: actually why not this PR?

coldlink: the screenshot you've got in the description clearly shows facebook last

```suggestion
      return ['apple', 'google', 'facebook'];
    case 'android':
      return ['google', 'apple', 'facebook'];
    default:
      return ['google', 'apple', 'facebook'];
```

coldlink: it's a small enough change to make

PR #2010, opened 2022-10-17
UI changes when accessing Gateway from a native app client

coldlink:
## What does this change?

The signin and registration flows from the native app will in future switch to a browser as part of the OAuth flow with Okta. To ensure a seamless experience and make the flow look more like the native app, we tweak parts of the design when the `appClientId` query parameter is one belonging to an iOS or Android client (i.e. the native apps).

To make this happen on the backend, we now read `appClientId` and check it with Okta. We then pass `isNativeApp` in `pageData` to the clientside pages. `isNativeApp` will either be `'ios'`, `'android'`, or `undefined` (not an app client).

On the frontend, we use this flag to make the following changes:

- Reduce the header margins
- Hide the progress bar on the consents/welcome flow as the native app user goes straight from the password setting page back to the app
- Change 'Sign in to the Guardian' to 'Sign in to the Guardian app' on the already-signed-in page
- Hide the footer entirely
- Never display the CMP banner

We also made the following changes to the general flow, for both desktop and native browsers:

- Change 'this site' to 'this service' in the T&Cs link

Some nice screenshots:

Screen Shot 2022-10-18 at 16 18 31
Screen Shot 2022-10-18 at 16 18 03
Screen Shot 2022-10-18 at 16 17 28
Screen Shot 2022-10-18 at 16 17 02
Screen Shot 2022-10-18 at 16 16 54

### DIFF
--- a/cypress/integration/ete/registration/register.2.cy.ts
+++ b/cypress/integration/ete/registration/register.2.cy.ts
@@ -22,7 +22,7 @@ describe('Registration flow', () => {
         req.reply(200);
       });
       cy.visit('/signin?useIdapi=true');
-      cy.contains('This site is protected by reCAPTCHA and the Google')
+      cy.contains('This service is protected by reCAPTCHA and the Google')
         .contains('privacy policy')
         .click();
       cy.url().should('eq', googlePrivacyPolicyUrl);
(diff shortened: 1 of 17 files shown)


kelvin-chappell (approved):
My only comment is that `isNativeApp` suggests it's a boolean when it seems to be used as a boolean in some places and a thing to match on in others.  But that's not important.  Looks good 👍

PR #2012, opened 2022-10-24
`SignedInAs` Page | Fix "Continue" link to refresh session instead of redirecting

coldlink:
## What does this change?

There is currently a small bug/edge case where a user might have an Okta session, but not an Identity session, which means that they could get stuck into a login loop, as they'll be logged in on `profile.theguardian.com` which uses the Okta session, but not anywhere which hasn't migrated yet, e.g. Dotcom.

https://user-images.githubusercontent.com/13315440/197491280-4e9e6b8f-0322-4395-8d0c-d62e97d0247a.mov

To fix this, instead of redirecting straight to the `returnUrl` on the `SignedInAs` page, we instead take them to `/signin/refresh` which will refresh their Okta session cookie, and create new Identity cookies for the user, so that they remain "logged in".

https://user-images.githubusercontent.com/13315440/197494063-a353293b-cf8c-47f7-8be2-36f4284a9973.mov

### DIFF
--- a/cypress/integration/ete-okta/registration.2.cy.ts
+++ b/cypress/integration/ete-okta/registration.2.cy.ts
@@ -1109,7 +1109,12 @@ describe('Registration flow', () => {
             cy.contains(emailAddress);
             cy.contains('Continue')
               .should('have.attr', 'href')
-              .and('include', `https://${Cypress.env('BASE_URI')}/consents`);
+              .and(
+                'include',
+                `https://${Cypress.env(
+                  'BASE_URI',
+                )}/signin/refresh?returnUrl=https%3A%2F%2Fprofile.thegulocal.com%2Fconsents`,
+              );
             cy.contains('a', 'Sign in')
               .should('have.attr', 'href')
               .and('include', '/signout?returnUrl=');
--- a/cypress/integration/mocked/okta_register.3.cy.ts
+++ b/cypress/integration/mocked/okta_register.3.cy.ts
@@ -58,7 +58,11 @@ describe('Okta Register flow', () => {
       cy.contains('user@example.com');
       cy.contains('Continue')
         .should('have.attr', 'href')
-        .and('include', 'https://m.code.dev-theguardian.com');
+        .and('include', '/signin/refresh')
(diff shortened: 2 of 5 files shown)


cypress/integration/ete-okta/registration.2.cy.ts
```diff
@@ -1109,7 +1109,12 @@ describe('Registration flow', () => {
             cy.contains(emailAddress);
             cy.contains('Continue')
               .should('have.attr', 'href')
-              .and('include', `https://${Cypress.env('BASE_URI')}/consents`);
+              .and(
+                'include',
+                `https://${Cypress.env(
+                  'BASE_URI',
+                )}/signin/refresh?returnUrl=https%3A%2F%2Fprofile.thegulocal.com%2Fconsents`,
```
raphaelkabo, line 1116: Are these URL changes just extra bugs you found or are they directly as a result of the changes?

cypress/integration/mocked/okta_register.3.cy.ts
```diff
@@ -125,7 +129,8 @@ describe('Okta Register flow', () => {
       cy.contains('user@example.com');
       cy.contains('Continue')
         .should('have.attr', 'href')
-        .and('include', 'https://m.code.dev-theguardian.com');
+        .and('include', '/signin/refresh')
+        .and('include', 'returnUrl=https%3A%2F%2Fm.code.dev-theguardian.com');
```
raphaelkabo, line 133: How come the `returnUrl` on line 64 has a slash at the end but the one here doesn't? Weird.

coldlink: I think the test was written with a slash in one test, and without a slash in the other. It doesn't matter too much

cypress/integration/ete-okta/registration.2.cy.ts
raphaelkabo: Oh great 👍

raphaelkabo (approved):
Good fix - this'll do it! 👍

PR #2032, opened 2022-11-01
Use Zod for our Okta interface | 🐙 ➡️ 🆉🅾🅳🤖 

coldlink:
## What does this change?

Updates all our types/models we use with Okta to be defined with [zod](https://github.com/colinhacks/zod) instead! Zod is a TypeScript-first schema declaration and validation library.

This has the benefit of still keeping the same types and type safety at compile time, but also being able to easily validate types at  run time too, e.g. responses from the Okta API.

We were already using Zod in this project, specifically in our rate limiter configuration thanks to @ob6160.

This PR updates all the Okta models and API responses to use Zod, and fixed any tests/issues that were encountered when doing this. Types for the Okta models were taken from the [Okta API documentation](https://developer.okta.com/docs/reference/core-okta-api/), and we've linked to them above the types and schema where appropriate for reference.

## Notes

Unlike TypeScript, Zod creates schemas of data structures which can be checked at both compile and runtime, unlike TS where type checking only happens at compile time.

In Zod you have to first define a schema which defines the structure, types, and validation e.g.

```ts
import { z } from 'zod';

const exampleSchema = z.object({
  keyA: z.string(),
  keyB: z.boolean().nullable().optional(),
  keyC: z.object({
    foo: z.enum(['bar', 'baz']),
  }).optional(),
});
```

You can then infer the typescript type by using `z.infer`:

```ts
type Example = z.infer<typeof exampleSchema>;

// equivalent to
interface Example {
  keyA: string;
  keyB?:  boolean | null;
  keyC?: {
    foo: 'bar' | 'baz';
  };
}
```

You can then use the schema to validate objects at runtime without having to resort to type guards or manual type checking

```ts
// old way
const isExample = (obj: unknown): obj is Example => {
  const maybeExample = obj as Example;
  return !!(maybeExample.keyA && typeOf maybeExample.keyA === 'string' && ... other checks);
};

// or

try { 
 const example = await response.json() as Example;
  // not checking the types at runtime here
  return {
    keyA: example.keyA,
    keyB: example.keyB,
    keyC: example.keyC,
  }
} catch (error) {
 console.error('failed to parse example', error);
}

// becomes simply
try {
  // this checks the types at runtime to make sure it's valid!
  return userResponseSchema.parse(await response.json());
} catch (error) {
 // handle parsing error
}
```

### DIFF
--- a/cypress/integration/mocked-okta/registerController.1.cy.ts
+++ b/cypress/integration/mocked-okta/registerController.1.cy.ts
@@ -27,7 +27,7 @@ userStatuses.forEach((status) => {
             "Then I should be shown the 'Check your email inbox' page",
             () => {
               // Set the correct user status on the response
-              const response = { ...userResponse.response, status };
+              const response = { ...userResponse.response, status: 'STAGED' };
               cy.mockNext(userResponse.code, response);
               cy.get('button[type=submit]').click();
               verifyInRegularEmailSentPage();
--- a/cypress/integration/mocked-okta/resendEmailController.3.cy.ts
+++ b/cypress/integration/mocked-okta/resendEmailController.3.cy.ts
@@ -35,7 +35,7 @@ userStatuses.forEach((status) => {
             "Then I should be shown the 'Check your email inbox' page",
             () => {
               // Set the correct user status on the response
-              const response = { ...userResponse.response, status };
+              const response = { ...userResponse.response, status: 'STAGED' };
               cy.mockNext(userResponse.code, response);
               cy.get('[data-cy="main-form-submit-button"]').click();
               verifyInRegularEmailSentPage();
@@ -175,7 +175,7 @@ userStatuses.forEach((status) => {
             "Then I should be shown the 'Check your email inbox' page",
             () => {
               // Set the correct user status on the response
-              const response = { ...userResponse.response, status };
+              const response = { ...userResponse.response, status: 'STAGED' };
               cy.mockNext(userResponse.code, response);
               cy.get('[data-cy="main-form-submit-button"]').click();
               verifyInRegularEmailSentPage();
@@ -261,7 +261,7 @@ userStatuses.forEach((status) => {
             "Then I should be shown the 'Check your email inbox' page",
             () => {
               // Set the correct user status on the response
-              const response = { ...userResponse.response, status };
+              const response = { ...userResponse.response, status: 'STAGED' };
               cy.mockNext(userResponse.code, response);
               cy.get('[data-cy="main-form-submit-button"]').click();
               verifyInRegularEmailSentPage();
@@ -343,7 +343,7 @@ userStatuses.forEach((status) => {
             "Then I should be shown the 'Check your email inbox' page",
             () => {
               // Set the correct user status on the response
-              const response = { ...userResponse.response, status };
+              const response = { ...userResponse.response, status: 'STAGED' };
               cy.mockNext(userResponse.code, response);
               cy.get('[data-cy="main-form-submit-button"]').click();
               verifyInRegularEmailSentPage();
--- a/cypress/integration/mocked/okta_send_reset_password.4.cy.ts
+++ b/cypress/integration/mocked/okta_send_reset_password.4.cy.ts
@@ -188,6 +188,7 @@ describe('Send password reset email in Okta', () => {
       cy.get('input[name="email"]').type(email);
       cy.mockNext(200, mockUserStaged);
       cy.mockNext(200, {
+        activationUrl: `token_token_token_to`,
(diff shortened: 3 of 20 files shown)


cypress/integration/mocked-okta/registerController.1.cy.ts
```diff
@@ -27,7 +27,7 @@ userStatuses.forEach((status) => {
             "Then I should be shown the 'Check your email inbox' page",
             () => {
               // Set the correct user status on the response
-              const response = { ...userResponse.response, status };
```
coldlink, line 30: This was changed as the actual behaviour in a non-mocked environment would be to create a user in the `STAGED` state

cypress/integration/mocked-okta/resendEmailController.3.cy.ts
```diff
@@ -35,7 +35,7 @@ userStatuses.forEach((status) => {
             "Then I should be shown the 'Check your email inbox' page",
             () => {
               // Set the correct user status on the response
-              const response = { ...userResponse.response, status };
+              const response = { ...userResponse.response, status: 'STAGED' };
```
coldlink, line 38: +1

cypress/integration/mocked-okta/resendEmailController.3.cy.ts
```diff
@@ -175,7 +175,7 @@ userStatuses.forEach((status) => {
...
```
coldlink, line 178: +2

cypress/integration/mocked-okta/resendEmailController.3.cy.ts
```diff
@@ -343,7 +343,7 @@ userStatuses.forEach((status) => {
...
```
coldlink, line 346: +3

cypress/integration/mocked/okta_send_reset_password.4.cy.ts
```diff
@@ -203,6 +204,7 @@ describe('Send password reset email in Okta', () => {
       cy.get('input[name="email"]').type(email);
       cy.mockNext(200, mockUserProvisioned);
       cy.mockNext(200, {
+        activationUrl: `token_token_token_to`,
```
coldlink, line 207: +1

src/server/models/okta/Authentication.ts
```diff
@@ -31,60 +44,79 @@ enum AuthenticationFactorResult {
...
-  secondName: string;
-  locale?: string;
-  login: string;
-  timeZone?: string;
-}
+const authenticationEmbeddedUserProfileSchema = z.object({
+  login: z.string(),
+  // okta says this is not nullable or optional, but there are some
+  // users in our system that do not have a firstName/lastName or it is null
+  firstName: z.string().nullable().optional(),
```
coldlink, line 60: we also use `.nullable()` with `.optional()` everywhere, as in an Okta API response, the response could either be `null` or not defined, hence we need both to handle these cases

src/server/lib/members-data-api/user-attributes.ts
```diff
@@ -79,12 +79,7 @@ export const getUserAttributes = async (
       );
     }
 
-    const unparsed = await response.json();
-
-    const userAttributesResponse =
-      await userAttributesResponseSchema.parseAsync(unparsed);
-
-    return userAttributesResponse;
+    return userAttributesResponseSchema.parse(await response.json());
```
ob6160, line 82: I wonder what we could do if the json we receive doesn't match the schema exactly and `parse()` throws but it still has usable data 🤔 .. do we consider that to be an error, or can we still extract the critical parts that allow us to proceed?

ob6160 (approved):
great work 👏 this is a great improvement for devx and improving our confidence in how we handle API responses at runtime

src/server/lib/members-data-api/user-attributes.ts
coldlink: Zod only looks out for the fields that you've defined in the schema, so as long as we get back the fields that are defined, zod will only attempt to parse those.

It's unlikely that Okta will change their API responses, as it's versioned, and we've marked any possible missing field as nullable or optional, so we're pretty safe.

cypress/integration/mocked-okta/registerController.1.cy.ts
raphaelkabo: Good point!

src/server/lib/members-data-api/user-attributes.ts
ob6160: aha , great point - if fields are removed on Oktas’ end we should probably error anyway as it could be a very bad thing 👍

PR #2046, opened 2022-11-08
Okta | Add support for reset/set password interception on the Okta hosted login page

coldlink:
## What does this change?

Updates the Okta hosted login interception script so that it's possible to pass in a `reset_password_token` and `set_password_token` like the `activation_token` so we can choose where to redirect users from the app during an OAuth flow.

For example, if the app intercepts a reset password link (`/reset-password/123456`), the app will initiate the OAuth flow, and passing the token as an additional parameter like `?...&reset_password_token=123456&...` similar to how they do for the `activation_token` currently. When the interception script sees this parameter is set, we redirect the browser back to `/reset-password/123456` with the OAuth query parameters, so that when the user sets a new password we're able to redirect them straight back to the application.

## App Implementation Notes

The apps team need to intercept the calls to `/reset-password/:token` and `/set-password/:token` like they do for `/welcome/:token`, and then pass the token along to the Okta SDK `.signIn`/`.signInWithBrowser` method, as an additional parameter, so `reset_password_token=<token>` for reset password, `set_password_token=<token>` for set password.

This will include all the OAuth parameters for the reset password flow, allowing the user to redirect back to the application after resetting/setting their password.

See the documentation for registration/welcome page: https://github.com/guardian/gateway/blob/main/docs/okta/native-apps-integration-guide.md#registration


raphaelkabo (approved):
🎉

PR #2047, opened 2022-11-10
Set password directly from account exists email sent to registered active users

raphaelkabo:
## What does this change?

When a user asks for a password reset, we do our very best not to prevent them from continuing to log in if they remembered their password. To that end, we add the following functionality into the controller which runs when an existing user attempts to register:

- If the user is `ACTIVE` and has a password set, we call the `forgotPassword()` method to generate a reset password token for them.
- If the user is `ACTIVE` but does not have a password set, we first set a placeholder password on the user so that we can reset it. Then we generate a reset password token the same way.
- This token is sent to the user in the 'this account already exists' email, where the 'Reset password' link becomes `/reset-password/:token`. Clicking on this link allows the user to immediately enter a new password, as long as they click the link within the expiration time.
- If we are unable to generate a reset password token for the user due to an error somewhere down the line, we simply send them the regular email: a 'this account already exists' email where the 'Reset password' link points to `/reset-password`.


raphaelkabo:
[Enable users to set new passwords directly from "This account already exists" email ( S)](https://trello.com/c/1gtgbEZG/4205-enable-users-to-set-new-passwords-directly-from-this-account-already-exists-email-s)

coldlink:
Is it also possible to get a cypress e2e test for this?

src/server/lib/okta/register.ts
```diff
@@ -152,12 +157,40 @@ export const sendRegistrationEmailByUserState = async (
...
-        // generic email if they don't have a password.
-        await sendAccountExistsEmail({
-          to: user.profile.email,
-        });
+        if (doesNotHavePassword) {
+          // The user does not have a password set, so we send them a generic email
+          // with no reset password token
+          await sendAccountExistsEmail({
+            to: user.profile.email,
+          });
```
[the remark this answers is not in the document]
raphaelkabo: I've refactored this a bit now and I think this will now give us the functionality we want.

src/server/lib/okta/dangerouslySetPlaceholderPassword.ts
```diff
@@ -0,0 +1,50 @@
...
+    });
+    return true;
+  } catch (error) {
+    logger.error(
+      `dangerouslySetPlaceholderPassword failed: Error setting placeholder password for user ${id}`,
+      error,
+    );
+    return false;
+  }
+};
```
coldlink, line 48: I'm not sure the return type of this method should be a boolean.

This is because in this case if an error is encountered the thrown error would be swallowed in this method and not returned to the calling function, which is not the behaviour that was happening before.

This should be changed.

```suggestion
const dangerouslySetPlaceholderPassword = async (
  id: string,
): Promise<void> => {
  try {
    // Generate an recoveryToken OTT and put user into RECOVERY state
    const recoveryToken = await dangerouslyResetPassword(id);
    // Validate the token
    const { stateToken } = await validateRecoveryToken({
      recoveryToken,
    });
    // Check if state token is defined
    if (!stateToken) {
      throw new OktaError({
        message:
          'dangerouslySetPlaceholderPassword failed: state token is undefined',
      });
    }
    // Set the placeholder password as a cryptographically secure UUID
    await resetPassword({
      stateToken,
      newPassword: crypto.randomUUID(),
    });
  } catch (error) {
    logger.error(
      `dangerouslySetPlaceholderPassword failed: Error setting placeholder password for user ${id}`,
      error,
    );
    throw error;
  }
};
```

coldlink (changes requested):

src/server/lib/okta/dangerouslySetPlaceholderPassword.ts
```diff
@@ -0,0 +1,47 @@
...
+/**
+ * This function is used ONLY for users who do not have a password set at all
+ * (i.e. social users or users imported from IDAPI). It does the following:
+ * 1. Runs dangerouslyResetPassword() to generate a OTT (one-time token) to
+ *    allow us to reset the user's password. This puts that user into RECOVERY
+ *    state which will prevent them from logging in again.
+ * 2. Uses the OTT to set a cryptographically secure placeholder password for the user.
+ * After these operations, we can send the user a password reset email.
+ * @param id The Okta user ID
+ * @returns A boolean indicating whether the placeholder password was set successfully.
```
coldlink, line 16: This line comment is no longer relevant

coldlink (approved):

PR #2048, opened 2022-11-10
Upgrade runtime to Node 18 LTS

coldlink:
## What does this change?

Upgrades the project to use Node 18, now in LTS, as the underlying runtime. This also applies to the github actions too (which use the `.nvmrc` file to determine which node version to use).

Since Node 18 now ships with it's own version of `fetch`, we've replaced instances of `node-fetch` with the native `fetch` instead.

This has allowed us to remove two dependencies:
- `node-fetch` -> Use native `fetch` API, functionally identical
- `timeout-signal` -> Core functionality moved to `src/server/lib/timeoutSignal.ts`

Updated the files/tests/implementation to reflect this change too.

## Notes

The AMI also has to be updated to Node 18 too.

~So I've created a temp ami for testing called `gateway-node-18` which does this. Then updated this PR to use this AMI in https://github.com/guardian/gateway/pull/2048/commits/273c5ae2d477c4413c1c37b4cc38bb0f09f3f2cd~

~This commit should be removed pre-merge, and the base `gateway` AMI should instead be updated to use Node 18 before merging this PR in.~

Updated the base gateway AMI to use Node 18 now.

## Tested
- [x] CODE


kelvin-chappell (approved):
👍

PR #2057, opened 2022-11-28
Okta | Show onboarding flow for new social users

coldlink:
## What does this change?

In Okta it is currently not possible to differentiate between new social users, and existing social users. This means that is was not possible to show social users the onboarding flow.

However there is a case where a new social user differentiates form existing social users. This is to do with the email validated status. A new social user will be in the `GuardianUser-EmailValidated` but their `emailValidated` field in their user profile will be `false` or `undefined`. So we have a remediation step in our callback to handle this discrepancy, and set the `emailValidated` flag should the user be in the `GuardianUser-EmailValidated` group. This was introduced in this commit: https://github.com/guardian/gateway/commit/9af078165aa5add60b9e11bfe6731ff25badca6e

This means that the only time this discrepancy will occur is for new social users, meaning at this point we can differentiate them from existing social users. This allows us to use this functionality and set a flag to show the onboarding flow `authState.confirmationPage = consentPages[0].path;`. At any other point in time the user will have both the `GuardianUser-EmailValidated` group and the `emailValidated` flag set, so they would not see the onboarding flow again.

A number of other options were considered for this functionality, but would involve other changes to the Okta configuration.

1. Using Okta groups The first option we tried was to use Okta groups to do this.

The theory was for new social users we assign them to a specific Okta group that we create, we can then check this group, and if a user is in this group show then the onboarding flow. We can then remove them from this group so they're not shown the onboarding flow again.

However although this does work and new social users do see the Onboarding flow. I noticed that all social users started seeing the onboarding flow, whether they were new users, existing users, or had even seen the onboarding flow before. Turns out Okta will always add all social users to the new group if they are not currently in it. So even when removing the group from the user,  the next time they sign in with social, they get added to the group again, and they get shown the onboarding flow again. Which wasn't ideal.

2. Using the account `created` timestamp on the user object.
 
This option was to use the `created` field in the user profile to determine if we should show the onboarding flow or not. This would involve checking this field every time a user went to the oauth callback endpoint, and if it's within a certain time period, say 1 minute, then to show the onboarding flow in that scenario.

However the account `created` field isn't returned by default in either the access or id token. So we'd either have to make an additional API call for every time a user goes to the oauth callback endpoint, or to add a custom claim to add this field to the id token.

Using the API option would require making additional API calls, which we're trying to avoid, due to a limited number of API calls we're able to make to Okta. While adding this field to the id token would require changes to the Okta terraform configuration to add this in, and would add an additional payload to what's only effectively used once.

Also we'd have to chose the time period to check to avoid users seeing the onboarding flow multiple times if we set it too long, e.g. if they sign in immediately after registering on another browser/device, or not seeing the onboarding flow at all if we set it too short.

Hence the reason I went with the approach that we did.

## Tested

- [x] CODE

### New social user

https://user-images.githubusercontent.com/13315440/204318121-ae5747f7-fd65-44ad-809c-9a71df325d50.mov

### Existing social user

https://user-images.githubusercontent.com/13315440/204319149-1fa300d7-9f6d-46f1-92c0-df95c9273967.mov

### DIFF
--- a/src/server/routes/oauth.ts
+++ b/src/server/routes/oauth.ts
@@ -21,6 +21,7 @@ import { IdTokenClaims } from 'openid-client';
 import { updateUser } from '@/server/lib/okta/api/users';
 import { getApp } from '@/server/lib/okta/api/apps';
 import { setUserFeatureCookies } from '@/server/lib/user-features';
+import { consentPages } from './consents';
 
 interface OAuthError {
   error: string;
@@ -166,6 +167,10 @@ router.get(
       // So this is a workaround to set the emailValidated field to true in the Okta user profile
       // if the user is in the GuardianUser-EmailValidated group, based on custom claims we have
       // set up on the id_token.
+      // This scenario only occurs in the case of new social users, as existing users will have
+      // the emailValidated field set to true in their Okta user profile.
+      // So we can use this functionality to show the onboarding flow for new social users, as
+      // there is no other trivial way to do this.
       if (tokenSet.id_token) {
         // extracting the custom claims from the id_token and the sub (user id)
         const { user_groups, email_validated, sub } =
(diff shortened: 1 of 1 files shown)


kelvin-chappell (approved):
An amazingly simple solution! 👍

PR #2058, opened 2022-11-29
Move emails to gateway

raphaelkabo:
This moves the 'Complete Registration' and 'Reset Password' emails to be sent from Gateway (they were previously sent by Okta as a side-effect of Okta API calls).

As a result, this also removes some of the public routes we exposed which returned email template HTML for consumption by the Okta Terraform scripts - we leave the 'accidental email' route because we still need it to allow Okta to send accidental emails.

The 'Verify' email templates are renamed to 'CompleteRegistration' to better match their use case.

- [X] Tested on CODE


raphaelkabo:
[Move remaining Okta emails to Gateway](https://trello.com/c/VVJiQW0s/4221-move-remaining-okta-emails-to-gateway)

src/server/lib/okta/api/users.ts
```diff
@@ -39,8 +42,13 @@ const { okta } = getConfiguration();
...
 export const createUser = async (
   body: UserCreationRequest,
+  sendEmail = true,
 ): Promise<UserResponse> => {
-  const path = buildUrl('/api/v1/users');
+  const path = buildApiUrlWithQueryParams(
+    '/api/v1/users',
+    {},
+    { activate: sendEmail },
+  );
```
coldlink, line 51: Confused by the logic here?

Surely `sendEmail` should be false in this scenario? As we don't want to send an email from Okta in this case.

And `activate`/`sendEmail` as two different names also confusing me too.

What's happening in this scenario?

raphaelkabo: I've added some comments to explain what's going on - essentially, the `activate` param tells Okta to peform the `/api/v1/users/${userId}/lifecycle/activate` operation (with `sendEmail` set to `true`) on the newly created user immediately after creation, which would send them an email if they didn't have a password set on creation. If they _did_ have a password set, it would simply activate them, but we don't use that flow.

raphaelkabo: I've also changed the default value of `sendEmail` to `false` in all three API functions (`activateUser`, `reactivateUser`, and `createUser`).

coldlink: I think we should just remove the `sendEmail` parameter as an option here, and always have `activate` set to `false`. There's no scenario where we'd not want it to be `false` I don't think.

As if a developer were to set it to true, the user would receive an accidental email.

raphaelkabo: Done!

coldlink (approved):
Great work, looking good to me!

PR #2068, opened 2022-12-05
Dependencies | December 5 2022

raphaelkabo:
## What does this change?

Changes from previous #2053:

- Replaces `mjml-react`, which is no longer under development, with the actively supported fork `@faire/mjml-react`.

Holds back the following:

## Tested

- [ ] CODE
- [X] IE11


raphaelkabo:
Looks like we need to keep this one back for the moment until this bug with Cypress gets resolved: https://github.com/cypress-io/cypress/issues/24111

PR #2069, opened 2022-12-06
 Dependencies | December 6 2022

raphaelkabo:
## What does this change?

Changes from previous #2053:

- Updates `react-router-dom` from v6.4.3 to v6.4.5. Despite this being a minor version bump, it caused a Webpack error potentially connected to Babel failing to recognise the `react-router-dom` module file as a valid ESM file. To get around this, we add the Babel `sourceType: 'unambiguous'` Babel option.
 
Holds back the following:
* **NEW:** Keeps `cypress` at v11 because the upgrade to v12 will be handled in a separate PR

## Tested

- [x] CODE
- [X] IE11


coldlink (approved):
Looking okay to me!

Let's review bundle sizes like we said later, starting to balloooooon. 🎈

PR #2070, opened 2022-12-09
Capture registration location for social registrations via onboarding flow

vlbee:
## What does this change?

When a user clicks "next" on any of the onboarding flow pages, and additional hidden input field indicating CMP consent state of the client will be posted to the server. 

On the first page of the flow only ('Stay in touch' page) this will be used by the server to update in the background a new user's registration location if they are consented. 

A new server util has been added called `updateRegistrationLocation`:
- if a registration location can be inferred (cmp consented && available country sent on req header)
- call IDAPI to check if a user already has a location
- If not, post then new location to IDAPI (which then updates the value in Okta)

Note: Okta is the source of truth for the registration_location field, but until the onboarding flow is migrated to Okta sessions, we do not have the user info needed to update Okta directly. 

## Why?

We rolled out a new registration location field and autopopulate this value when a new user registers on web using email/password #1970.

We also want to capture registration location (for consented users) who register via social, but as the social sign in/registration flow is handled entirely by Okta and does not permit us to pass additional parameters (Eg. location) along the authorization flow, the new user onboarding flow is the only place where we can reliably capture this for social.

## Tested in CODE

- [x] a new user creates an account via FB. Their location is undefined in okta but correctly defined after they click continue on the onboarding stay in touch page. 
- [x] a new user creates an account via FB. They set a location in MMA. That location is unchanged after they go through the onboarding flow. 
- [x] a new user rejects cmp consents then creates an account via FB. Their location is undefined in okta and stays undefined once they complete the onboarding flow
- [x] a non-social user with cmp consent is updated of they don't have a location set.

### DIFF
--- a/src/server/lib/updateRegistrationLocation.ts
+++ b/src/server/lib/updateRegistrationLocation.ts
@@ -1,0 +1,44 @@
+import { Request } from 'express';
+import { logger } from '@/server/lib/serverSideLogger';
+import { RegistrationLocation } from '@/server/models/okta/User';
+import { isStringBoolean } from '@/server/lib/isStringBoolean';
+import { getRegistrationLocation } from '@/server/lib/getRegistrationLocation';
+import {
+  read as readIdapiUser,
+  addRegistrationLocation,
+} from '@/server/lib/idapi/user';
+
+/**
+ * Until Gateway/Onboarding journey is migrated to Okta sessions, we don't have access to Okta User ID, only sg_gu_u cookie,
+ * so we need to add reg location via idapi (which updates Okta immediately). When Okta sessions are available, this should be refactored
+ * to use okta directly (Which is the source of truth for the user's registration location field)
+ */
+export const updateRegistrationLocationViaIDAPI = async (
+  ip: string,
+  sc_gu_u: string,
+  req: Request,
+) => {
+  const { _cmpConsentedState = false } = req.body;
+
+  const registrationLocation: RegistrationLocation | undefined =
+    getRegistrationLocation(req, isStringBoolean(_cmpConsentedState));
+
+  if (!!registrationLocation) {
+    try {
+      const user = await readIdapiUser(ip, sc_gu_u);
+      // don't update users who already have a location set
+      if (!!user.privateFields.registrationLocation) {
+        return;
+      }
+      await addRegistrationLocation(registrationLocation, ip, sc_gu_u);
+    } catch (error) {
+      logger.error(
+        `${req.method} ${req.originalUrl} Error updating registrationLocation via IDAPI`,
+        error,
+        {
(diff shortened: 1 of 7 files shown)


src/server/routes/consents.ts
```diff
@@ -347,6 +349,10 @@ router.post(
       const { update, pageTitle: _pageTitle } = consentPages[pageIndex];
       pageTitle = _pageTitle;
 
+      // If on the first page, attempt to update location for consented users.
+      pageTitle == 'Stay in touch' &&
+        updateRegistrationLocationViaIDAPI(req.ip, sc_gu_u, req);
+
```
coldlink, line 355: I'd suggest looking at the `pageIndex` here, and using an `if` statement for readability, also using the strict equality (`===`)

```suggestion
      // If on the first page, attempt to update location for consented users.
      if (pageIndex === 0) {
        updateRegistrationLocationViaIDAPI(req.ip, sc_gu_u, req);
      }
```

If you really wanted to use `pageTitle` for some reason then i suggest using `CONSENTS_PAGES.CONTACT` instead

```suggestion
      // If on the first page, attempt to update location for consented users.
      if (pageTitle === CONSENTS_PAGES.CONTACT) {
        updateRegistrationLocationViaIDAPI(req.ip, sc_gu_u, req);
      }
```

src/server/lib/updateRegistrationLocation.ts
```diff
@@ -0,0 +1,44 @@
...
+    } catch (error) {
+      logger.error(
+        `${req.method} ${req.originalUrl} Error updating registrationLocation via IDAPI`,
+        error,
+        {
+          request_id: req.get('x-request-id'),
+        },
+      );
+    }
+  }
```
coldlink, line 43: I'd refactor this slightly to return early if we can to avoid too much nesting for readability, though this is just preference e.g.

```suggestion
  // don't update users if we don't have a location from the cmp 
  if (!registrationLocation) {
    return
  }

  try {
    const user = await readIdapiUser(ip, sc_gu_u);
    // don't update users who already have a location set
    if (!!user.privateFields.registrationLocation) {
      return;
    }
    await addRegistrationLocation(registrationLocation, ip, sc_gu_u);
  } catch (error) {
    logger.error(
      `${req.method} ${req.originalUrl} Error updating registrationLocation via IDAPI`,
      error,
      {
        request_id: req.get('x-request-id'),
      },
    );
  }
```

coldlink (commented):
Looks good, just a couple of small comments!

src/server/routes/consents.ts
vlbee: Very obvious suggestion 😁👍

coldlink (approved):

PR #2071, opened 2022-12-09
Reducing bundle size!

coldlink:
## What does this change?

A set of health PRs to significantly reduce the sizes of both our legacy browser and modern JS bundles and reduce dependency on external libraries.

### Changed packages

- Remove `react-router-dom`
- Remove `@types/react-router-dom`
- Remove `@sentry/tracing`
- Remove `js-sha1`
- Add `webpack-bundle-analyzer` in dev dependencies to create detailed reports of our bundle sizes

### Code changes

- Improve `import` directives to facilitate tree-shaking
- Remove the large `js-sha1` package and replace it with a function which makes use of builtin browser `crypto.subtle` API. We only generate SHA1 hashes to interface with the Pwned Passwords API. On IE11, which does not have the crypto APIs, the Pwned Passwords API call simply fails, which means users do not get an _immediate_ response on whether their password is safe to use. The ExpressJS backend still makes this check, so IE11 users will receive an appropriate error message after attempting to submit the form.


coldlink:
Moving into smaller PRs

PR #2111, opened 2023-01-16
Fix: Add query params to links missing params

coldlink:
## What does this change?

Some links on the `/email-sent` page didn't include any query parameters, which meant if a user clicked on any of those links then the query parameters would be removed.

This causes a problem on the Okta apps migration, where the query parameters being available are reliant for deeplinking to work. So if a user navigated away from the `/email-sent` page by using one of the links, and then submitted another form, we wouldn't know where to deeplink the user to.

This PR fixes this by attached the query params to these links, which was already being used on the page for the submit action on the "resend email" form functionality.

### DIFF
--- a/src/client/pages/EmailSent.tsx
+++ b/src/client/pages/EmailSent.tsx
@@ -65,7 +65,10 @@ export const EmailSent = ({
       {changeEmailPage && (
         <MainBodyText>
           Wrong email address?{' '}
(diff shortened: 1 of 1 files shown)


mohammad-haque (approved):
Nice spot @coldlink 
From the code change perspective it looks good to me - we can test it once it is available in CODE

PR #2113, opened 2023-01-19
Allow resend from /welcome/* after error

raphaelkabo:
## What does this change?

If a user attempts to submit the form on `/welcome/expired` or `/welcome/resend`, and Gateway responds with an error, the `/welcome/email-sent` page displays a user-facing error message but no option to resend the email. 

image

This is because we don't pass the email address into the `catch` branch which displays the error page, which this PR does:

Screenshot 2023-01-19 at 10 43 45

### DIFF
--- a/src/server/routes/welcome.ts
+++ b/src/server/routes/welcome.ts
@@ -131,8 +131,8 @@ router.post(
 );
 
 const OktaResendEmail = async (req: Request, res: ResponseWithRequestState) => {
+  const { email } = req.body;
   try {
-    const { email } = req.body;
     const state = res.locals;
 
     if (typeof email !== 'undefined') {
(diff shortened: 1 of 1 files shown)


src/server/routes/welcome.ts
```diff
@@ -163,12 +163,19 @@ const OktaResendEmail = async (req: Request, res: ResponseWithRequestState) => {
 
     trackMetric('OktaWelcomeResendEmail::Failure');
 
+    const { email } = req.body;
```
coldlink, line 166: Why not move the `const { email } = req.body;` line outside the try catch block so you don't have to define it twice (on L135 and here on L166)

raphaelkabo: Good catch!

coldlink (approved):

PR #2116, opened 2023-01-20
Reduce bundle size | Sentry optimisation and other imports

coldlink:
### Sentry

Used https://docs.sentry.io/platforms/javascript/configuration/tree-shaking/ to optimise the tree shaking and reduce the
sentry bundle size.

| | Before | After | Difference |
| - | - | - | - |
| Modern | 112.34 kB (30.69 kB gzipped) | 62.96 kB (17.99 kB gzipped) | 49.38 kB (12.70 kB gzipped) |
| Legacy | 472.94 kB (119.07 kB gzipped) | 390.28 kb (101.09 kB gzipped) | 82.66 kB (17.98 kB gzipped) |

### Other changes

Avoid using `import * as X from 'Y'` syntax, which may help optimise other imports

### DIFF
--- a/docs/development.md
+++ b/docs/development.md
@@ -477,7 +477,7 @@ It's also used as [the Guardian Source Design System](https://theguardian.design
 Example of styling and adding it to a `p` tag using Emotion and Source:
 
 ```tsx
-import * as React from 'react';
+import React from 'react';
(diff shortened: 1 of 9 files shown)


Mark-McCracken (approved):
top stuff!

PR #2119, opened 2023-01-20
Localise onboarding flow newsletters for Australia

vlbee:
## What does this change?

We would like to localise the newsletters on the boarding flow based on country - UK, UK, AU.  Currently the Edition of the MorningBriefing newsleter is localized, but the remaining newsletters/consents (Events&Masterclasses) that appear on the /newsletters page is generic. 

This PR introduces a full localisation of the /newsletters page for Australia, and refactors the code so the same can be rolled out for other countries if desired.

What changes:
 - Full localization of the page beyond newsletter edition requires a consented CMP state: this is perpetuated to the server via the `encryptedStateCookie`
 - Introduced a new union type called `Permissioned Geolocation` to use in the localisation `NewsletterMap` to differentiate between Edition-localisation and full-localisation. 
 - Consents and Newsletters are displayed and updated via different code paths - I've added a `ConsentsOnNewslettersPageMap' so that we can manage localisation the same and handle subscription updates correctly.  (Is there a way to streamline this? Perhaps worth another refactor)
 
 **Caveats:**
 If someone manually changes their cmp state whilst on the newsletter page, the newsletters will not update until they navigate away and back to the page via the navigation buttons. However when they land on the Confirmation page it will show consents/newsletters based on the new state. I think this is a pretty small edge case but open to ideas on how to solve.
 
 **Permissioned AU page:**
 
image

**Unpermissioned AU page:**

image

## Accessibility


src/shared/model/QueryParams.ts
```diff
@@ -39,6 +39,8 @@ export interface PersistableQueryParams
   fromURI?: string;
   // This is the client Id of a calling application in Okta (ie iOS app etc)
   appClientId?: string;
+  // used to render server side page after redirection based on client state
+  cmp?: boolean;
```
lindseydew, line 43: Small comment, but since this variable is representing a boolean this could is isCmp

src/server/routes/consents.ts
```diff
@@ -255,6 +263,8 @@ router.get(
     let state = res.locals;
     const sc_gu_u = req.cookies.SC_GU_U;
 
+    // get CMP state from queryParams on redirected request
+    const _cmpConsentedState = isStringBoolean(req.query.cmp as string); // todo code smell
```
lindseydew, line 267: Could perhaps rename the `isStringBoolean` to `parseBooleanParam`. I would probably just accepts all the potential types from the query string in the method to avoid type casting

src/client/components/ConsentsNavigation.tsx
```diff
@@ -23,13 +24,16 @@ interface ConsentsNavigationProps {
 export const ConsentsNavigation = ({
   showContinueButton = true,
 }: ConsentsNavigationProps) => {
+  const _cmpConsentState = useCmpConsent();
```
lindseydew, line 27: Curious as to why you are choosing to prefix the variable name with an underscore? (No strong opinion here, just interested in coding style)

src/client/components/ConsentsNavigation.tsx
```diff
@@ -43,7 +47,7 @@ export const ConsentsNavigation = ({
               {
                 page: previousPage,
               },
-              queryParams,
+              queryParamsWithCmpState,
```
lindseydew, line 50: Since this variable is only used once, perhaps the spread operator could be done in line here.

src/shared/model/QueryParams.ts
lindseydew, line 43: Small comment, but since this variable is representing a boolean this could is isCmp

src/client/components/ConsentsNavigation.tsx
lindseydew: This could perhaps be `__isC`mpConsentState` since it is representing a boolean here

lindseydew (commented):
Looking really good. I've just left some very minor comments on style. I know there is a discussion going on about whether to use the query param or the secure cookie for persisting the state, so these may not be relevant.

src/client/components/ConsentsNavigation.tsx
vlbee: Just following the convention in the code to underscore hidden form values submitted, eg csrfToken.  I think it will help clarify what value is coming in from the form vs the `isCmpConsented` value on the server. 

This draft is very out of date as no  longer using query params.

src/server/routes/consents.ts
```diff
@@ -266,6 +278,10 @@ router.get(
       });
     }
 
+    // Checks if we can localize content
+    const encryptedState = readEncryptedStateCookie(req);
+    const { isCmpConsented } = encryptedState ?? {};
```
coldlink, line 283:
```suggestion
    const { isCmpConsented } = readEncryptedStateCookie(req) ?? {};
```

src/shared/lib/newsletterConsentsPageLocalisation.ts
```diff
@@ -0,0 +1,84 @@
...
+} from '@/shared/model/Geolocation';
+import { Newsletters } from '@/shared/model/Newsletter';
+import { CONSENTS_NEWSLETTERS_PAGE } from '../model/Consent';
+
+// Permissions a geolocation if cmp consent is true
+export const getPermissionedGeolocation = (
+  cmpConsentState: boolean | undefined,
+  geolocation: GeoLocation | undefined,
+): GeoLocation | PermissionedGeolocation | undefined =>
+  !!cmpConsentState && geolocation === 'AU' ? 'AU_permissioned' : geolocation;
```
coldlink, line 13: Just finding this logic slightly hard to read, I wonder if this could be improved somehow?

It might be slightly longer, but an if statement is easier to read, that's just my opinion though

```suggestion
export const getPermissionedGeolocation = (
  cmpConsentState: boolean | undefined,
  geolocation: GeoLocation | undefined,
): GeoLocation | PermissionedGeolocation | undefined => {
  if (!!cmpConsentState && geolocation === 'AU') {
    return 'AU_permissioned'
  }
  return geolocation;
}
```

coldlink:
Logic is pretty good! Great work.

src/shared/lib/newsletterConsentsPageLocalisation.ts
vlbee: 👍 yes, will amend

coldlink (approved):

PR #2120, opened 2023-01-20
Reduce bundle size | Remove `js-sha1` package

coldlink:
## What does this change?

Removes the `js-sha1` package and replaces it with the browser `SubtleCrypto` API.

We add a `cryptoSubtleFeatureTest` to test if we should use this API. If we're able to use the API then we perform the breached checked as normal.

If the feature test fails either because the user is using Internet Explorer or the Web Crypto API is not supported, then we skip the breached check on the client side, and assume a valid password has been entered if between 8 - 72 characters.

The breached password check happens on the server side anyway, so if the user has entered a breached password on the client, we'll pick it up anyway on the server side and ask the user to change it should they be using an unsupported browser.

Tested support using Browserstack in IE11 and older chrome versions.

This saves around 6 kB (2 kB gzipped), and is more secure as it uses browser APIs rather than an npm package.

### DIFF
--- a/src/client/components/PasswordForm.tsx
+++ b/src/client/components/PasswordForm.tsx
@@ -19,8 +19,6 @@ import {
 } from '@guardian/source-foundations';
 import { CsrfFormField } from '@/client/components/CsrfFormField';
 import { css } from '@emotion/react';
-
-import sha1 from 'js-sha1';
 import {
   PasswordAutoComplete,
   PasswordInput,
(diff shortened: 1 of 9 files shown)


src/client/components/PasswordForm.tsx
```diff
@@ -217,12 +216,35 @@ const ValidationMessage = ({
   }
 };
 
+const cryptoSubtleFeatureTest = (browserName?: string): boolean => {
```
raphaelkabo, line 219: This is a good addition!

raphaelkabo (approved):
Good work!

PR #2121, opened 2023-01-20
Reduce bundle size | Remove `query-string`

coldlink:
## What does this change?

Removes the `query-string` package and replaces it with the browser `URLSearchParams` API.

Polyfill `URLSearchParams` for browsers that don't support the API.

Add additional methods and update tests to make sure this functionality remains identitcal to how we use it.

This saves a small amount in our bundle size, as well as removes the need for the dependency directly (which is causing problems to upgrade as it's moved to pure ESM).

Approx 6 kB (2 KB gzipped) in both the modern and legacy bundle.

Also updates the polyfill.io package so that it matches between the `renderer.tsx` (used by gateway pages) and the `okta-login.html` used for the intercept of the Okta hosted sign in page, so that the features match between the two.

### DIFF
--- a/src/server/lib/__tests__/getABForcedVariants.test.ts
+++ b/src/server/lib/__tests__/getABForcedVariants.test.ts
@@ -1,19 +1,15 @@
 import { Request } from 'express';
 import { getABForcedVariants } from '@/server/lib/getABForcedVariants';
 
-const fakeRequest = (queryParameters?: string[]) => ({
-  url: `https://guardian.com${
-    queryParameters && '?' + queryParameters.join('&')
-  }`,
-});
-
 describe('getABForcedVariants', () => {
   describe('when there are variants', () => {
     it('returns the variants stripped of their ab- prefix', () => {
-      const request = fakeRequest([
-        'ab-test1=testparameter1',
-        'ab-test2=testparameter2',
-      ]) as unknown as Request;
+      const request = {
+        query: {
+          'ab-test1': 'testparameter1',
+          'ab-test2': 'testparameter2',
+        },
+      } as unknown as Request;
       const expected = {
         test1: {
           variant: 'testparameter1',
@@ -28,10 +24,11 @@ describe('getABForcedVariants', () => {
   });
   describe('when there are multiple variants of the same key', () => {
     it('only uses the latest query in the string', () => {
-      const request = fakeRequest([
-        'ab-test3=testparameter3',
-        'ab-test3=testparameter4',
-      ]) as unknown as Request;
+      const request = {
+        query: {
+          'ab-test3': ['testparameter3', 'testparameter4'],
+        },
+      } as unknown as Request;
       const expected = {
         test3: {
           variant: 'testparameter4',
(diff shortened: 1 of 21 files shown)


src/server/lib/__tests__/getABForcedVariants.test.ts
```diff
@@ -1,19 +1,15 @@
...
-      const request = fakeRequest([
-        'ab-test1=testparameter1',
-        'ab-test2=testparameter2',
-      ]) as unknown as Request;
+      const request = {
+        query: {
+          'ab-test1': 'testparameter1',
+          'ab-test2': 'testparameter2',
+        },
+      } as unknown as Request;
```
raphaelkabo, line 12: How does `as unknown as Request` work here?

src/shared/model/QueryParams.ts
```diff
@@ -1,6 +1,12 @@
-import { StringifiableRecord } from 'query-string';
 import { ValidClientId } from '../lib/clientId';
 
+export type Stringifiable = string | boolean | number | null | undefined;
```
raphaelkabo, line 3: Does `undefined` need to be here? How does it stringify?

```
> undefined.toString();
VM101:1 Uncaught TypeError: Cannot read properties of undefined (reading 'toString')
    at <anonymous>:1:11
```

src/server/lib/__tests__/getABForcedVariants.test.ts
coldlink: It's basically a bit of a hack to avoid having to generate the whole `Request` object.

It's first saying parse this object as the `unknown` type which is a catch all for all valid types, then from the `unknown` type parse it as anything else, so in this case the `Request` type.

https://basarat.gitbook.io/typescript/type-system/type-assertion#double-assertion

coldlink: As this is for a test, this is fine, but wouldn't really want to use this in actual code!

src/shared/model/QueryParams.ts
coldlink: It's more about how the `URLSearchParams` parses these values

```
> new URLSearchParams({foo: undefined, bar: null, baz: false, x: 1, y: "two"})
URLSearchParams {
  'foo' => 'undefined',
  'bar' => 'null',
  'baz' => 'false',
  'x' => '1',
  'y' => 'two' }
  ```

coldlink: It stringifies everything! Hence it's stringifiable!

raphaelkabo: Wow it really does stringify everything.

src/server/lib/__tests__/getABForcedVariants.test.ts
raphaelkabo: Very useful trick though. 👍

src/shared/model/QueryParams.ts
coldlink: Also if you see how `Stringifiable` is used, without it we wouldn't be able to include any optional parameters in our type definitions.

raphaelkabo (approved):

PR #2125, opened 2023-01-23
Fix | Input/Email related issues

coldlink:
## What does this change?

Worth going through this PR commit by commit when reviewing.

- Add `The Guardian` to the `FromEmailAddress` in email sending from AWS SES https://github.com/guardian/gateway/pull/2125/commits/f1ca64c9cc66e02567e5de7bfeb741ebde00f238
- Fix input validation to only transition validity state on if the input is not empty, and on change/on input when the user stops typing https://github.com/guardian/gateway/pull/2125/commits/45a3570ca6786070584ad189a89941aa8fac47ff
- Allow for custom `autocomplete` values in the `EmailInput` component, as we don't want to autofill the email field on the registration page https://github.com/guardian/gateway/pull/2125/commits/b30b2d11798a57c228714a8c7caadcea8e65111f

https://user-images.githubusercontent.com/13315440/214252346-0609383b-5d9f-4203-973a-65c0f008d5f9.mov

### DIFF
--- a/src/client/lib/hooks/useInputValidityState.ts
+++ b/src/client/lib/hooks/useInputValidityState.ts
@@ -1,5 +1,16 @@
 import { useCallback, useState } from 'react';
 
+const isInputInvalid = (validityState: ValidityState) =>
+  validityState.customError ||
+  validityState.badInput ||
+  validityState.patternMismatch ||
+  validityState.rangeOverflow ||
+  validityState.rangeUnderflow ||
+  validityState.tooLong ||
+  validityState.tooShort ||
+  validityState.stepMismatch ||
+  validityState.typeMismatch;
+
 export enum InputFieldState {
   VALID = 'valid',
   EMPTY = 'empty',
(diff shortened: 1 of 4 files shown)


src/client/lib/hooks/useInputValidityState.ts
```diff
@@ -44,28 +46,20 @@ export const useInputValidityState = (
...
-        e.nativeEvent.composed !== undefined && !e.nativeEvent.composed;
-
-      // We also validate the field when filled via an autocomplete action in the browser.
-      // We can tell when this happens because the `data` property will be undefined.
-      // Usually from a keystroke, `data` would contain the value.
-      const originatingFromAutoComplete =
-        e.nativeEvent.composed !== undefined &&
-        (e.nativeEvent as Event & { data?: string })?.data === undefined;
-
-      if (notOriginatingFromUserInput || originatingFromAutoComplete) {
```
coldlink, line 68: this functionality wasn't actually working as expected, and actually caused issues with the validity not update on input, as both were always set to false, regardless of autocomplete or password manager.

So instead I removed this and replaced this with a check to see if the input was one of the specific "invalid" states and only transition the validity state if not.

raphaelkabo: This makes sense! Simpler is better when we're trying to cover a bunch of browsers at once...

raphaelkabo (approved):

PR #2128, opened 2023-01-25
Persist cmpState in encryptedStateCookie

vlbee:
## What does this change?

Adds the cmpState to the EncryptedState cookie and persists it the though the onboarding flow (Consent pages)

The cmpState is sent to the server on each POST /consents/:page request and added/updated in the EncryptedStateCookie at that point. 

We need to know the cmp state on the server in order to localise newsletters in #2119 (draft)

## Accessibility

### DIFF
--- a/src/server/routes/consents.ts
+++ b/src/server/routes/consents.ts
@@ -43,6 +43,8 @@ import { ConsentPath, RoutePaths } from '@/shared/model/Routes';
 import { PageTitle } from '@/shared/model/PageTitle';
 import { mergeRequestState } from '@/server/lib/requestState';
 import { updateRegistrationLocationViaIDAPI } from '../lib/updateRegistrationLocation';
+import { updateEncryptedStateCookie } from '../lib/encryptedStateCookie';
+import { isStringBoolean } from '../lib/isStringBoolean';
 
 interface ConsentPage {
   page: ConsentPath;
@@ -333,6 +335,7 @@ router.post(
     let state = res.locals;
 
     const sc_gu_u = req.cookies.SC_GU_U;
+    const _cmpConsentedState = isStringBoolean(req.body._cmpConsentedState);
 
     const { page } = req.params;
     let status = 200;
@@ -364,6 +367,10 @@ router.post(
         `${consentPages[pageIndex + 1].path}`,
         state.queryParams,
       );
+
+      updateEncryptedStateCookie(req, res, {
+        cmpConsentedState: _cmpConsentedState,
+      });
 
       return res.redirect(303, url);
     } catch (error) {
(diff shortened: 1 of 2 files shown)


src/server/lib/encryptedStateCookie.ts
```diff
@@ -61,7 +62,20 @@ export const readEncryptedStateCookie = (
...
+        email: decryptedState.email,
+        emailType: decryptedState.emailType,
+        passwordSetOnWelcomePage: isStringBoolean(
+          decryptedState.passwordSetOnWelcomePage,
+        ),
+        status: decryptedState.status,
+        signInRedirect: isStringBoolean(decryptedState.signInRedirect),
+        cmpConsentedState: isStringBoolean(decryptedState.cmpConsentedState),
+        queryParams: decryptedState.queryParams,
+      };
```
coldlink, line 78: you don't actually need to validate the string boolean values here, the data was encoded and encrypted as JSON which supports boolean values!

src/server/routes/consents.ts
```diff
@@ -365,6 +367,10 @@ router.post(
         state.queryParams,
       );
 
+      setEncryptedStateCookie(res, {
+        cmpConsentedState: _cmpConsentedState,
```
coldlink, line 371: Instead what you could do is to use `isStringBoolean` here to covert this string into a boolean

```suggestion
        cmpConsentedState: isStringBoolean(_cmpConsentedState),
```

src/server/lib/encryptedStateCookie.ts
coldlink: so you can leave it as 
```suggestion
return JSON.parse(decrypted);
```

src/server/routes/consents.ts
```diff
@@ -365,6 +367,10 @@ router.post(
         state.queryParams,
       );
 
+      setEncryptedStateCookie(res, {
```
coldlink, line 370: I'd suggest using the `updateEncryptedStateCookie` method instead, as this merges (updates/upserts) the state rather than overwriting it which is what `set` does.

src/server/routes/consents.ts
```diff
@@ -333,6 +334,7 @@ router.post(
     let state = res.locals;
 
     const sc_gu_u = req.cookies.SC_GU_U;
+    const _cmpConsentedState = req.body._cmpConsentedState;
```
lindseydew, line 338: Curious for my own understanding, how is the `__cmpConsentedState` first initialised and added to the request?

src/server/routes/consents.ts
vlbee: Thanks - I just realised this! will push a fix momentarily.

src/server/routes/consents.ts
vlbee: It's value is submitted as an invisible form input on every post req whe the user hits "continue"

src/server/routes/consents.ts
vlbee: I will see if I can amend the form to submit a proper boolean instead of a "boolean"

PR #2130, opened 2023-01-26
Dependency |  Cypress 12

coldlink:
## What does this change?

Updates Cypress to v12, which was a major release.

See changelog for more details: https://docs.cypress.io/guides/references/changelog#12-0-0

Most of this PR is fixing tests that broke as a result of this upgrade.

The main issue was how Cypress changed the session and origin features, which involved significant work to how cookies worked, but does now allow for cross origin testing, which we should really consider looking into, especially if we want to write in production (or more specifically in CODE) E2E tests!

One issue we encountered seemed to be related to `hostOnly` cookies, i.e cookies being set with the `.profile.thegulocal.com` subdomain (which is different to `profile.thegulocal.com` domain).

Cypress now creates cookie with the additional `.` making it a `hostOnly` cookie. But this causes havoc with some of our tests where the domain we're actually looking for is without the `.`, and in some cases causing issues with test behaviour, e.g. clearing cookies.

There is a cypress issue looking at this: https://github.com/cypress-io/cypress/issues/25174

But I've worked around the issue in this PR, we can come back and fix this should any cypress changes occur in the future, either by using a different domain, or a different test case.

Because of the origin changes, there were some issues now when we redirect to `https://m.code.dev-theguardian.com/`, in this case the cookies will no longer be readable, as it's on a different domain to `profile.thegulocal.com`. This issue was fixed by navigating to the signed in `/consents` route after sign in


cypress/integration/ete-okta/sign_out.2.cy.ts
```diff
@@ -35,11 +35,15 @@ describe('Sign out flow', () => {
         // attempt to sign out and redirect to reset password as it's on same domain
```
kelvin-chappell, line 35: I take it there was no relevance to being redirected to `reset-password`, apart from it being on the same domain?

cypress/integration/ete/sign_out/sign_out.3.cy.ts
```diff
@@ -8,6 +8,7 @@ describe('Sign out flow', () => {
 
   context('Signs a user out', () => {
     it('Removes IDAPI log in cookies and dotcom cookies when signing out', () => {
+      Cypress.Cookies.debug(true);
```
kelvin-chappell, line 11: Should this be here?

cypress/integration/ete/terms/jobs_terms.4.cy.ts
```diff
@@ -167,7 +167,7 @@ describe('Jobs terms and conditions flow', () => {
         // load the consents page as its on the same domain
         const termsAcceptPageUrl = `https://${Cypress.env(
           'BASE_URI',
-        )}/agree/GRS?returnUrl=https://jobs.theguardian.com/&useIdapi=true`;
+        )}/agree/GRS?returnUrl=https://jobs.thegulocal.com/&useIdapi=true`;
```
kelvin-chappell, line 170: Why have these references changed?

coldlink: This is to keep it on the same domain as the cypress tests. We were redirecting to jobs.theguardian.com in the test, but with the changes in cypress 12, we can't read the cookies anymore on that domain, so by using jobs.thegulocal.com we're able to still read cookies after the redirect.

Since jobs.thegulocal.com isn't a real url, we intercept it on line 214 https://github.com/guardian/gateway/pull/2130/files#diff-2f4d9a8a163424f357a6813aa1eb117563ca003463cf147167b270ac5824dbb6R214

cypress/integration/ete-okta/sign_out.2.cy.ts
coldlink: updated the comment

cypress/integration/ete/sign_out/sign_out.3.cy.ts
coldlink: good spot! i was using this for debugging the test

cypress/integration/ete/terms/jobs_terms.4.cy.ts
kelvin-chappell: Got it.  Thanks for the explanation.

kelvin-chappell (approved):
All makes sense.  Perhaps there'll be a performance improvement too

PR #2132, opened 2023-01-27
Replace Babel with SWC |🙅🏾 ➡️ 🦀|

coldlink:
## What does this change?

This PR introduces [SWC compiler](https://swc.rs/) - a Rust based compiler to replace Babel.

This should improve

- production build times
- faster development changes
- much smaller legacy webpack bundle

Fixed the `mode` variable, where it was set incorrectly when the `ENVIRONMENT` environment variable wasn't defined (which affected production builds in CI(!)).

Fixed source maps, which now work correctly in dev and prod builds.

The following times were tested on my laptop. CI in GitHub Actions will differ.

## Build times production

| | Babel | SWC |
| - | - | - |
| Server | 16058 ms | 4509 ms |
| Legacy | 18813 ms | 4523 ms |
| Modern | 17114 ms | 3993 ms |

## Build times development uncached

| | Babel | SWC |
| - | - | - |
| Server | 17490 ms | 4040 ms |
| Legacy | 17428 ms | 3974 ms |
| Modern | 17382 ms | 4481 ms |

## Bundle sizes

| | Babel | SWC |
| - | - | - |
| Legacy | 1.08 MB (283.48 KB gzipped) | 531.61 KB (142.77 KB gzipped) |
| Modern | 419.13 KB (127.57 KB gzipped) | 416.1 KB (130.83 KB gzipped) |


package.json
```diff
@@ -5,7 +5,7 @@
   "license": "Apache-2.0",
   "private": true,
   "scripts": {
-    "build": "webpack --mode production && yarn install --modules-folder ./build/node_modules --production --frozen-lockfile",
+    "build": "ENVIRONMENT=production webpack --mode production && yarn install --modules-folder ./build/node_modules --production --frozen-lockfile",
```
raphaelkabo, line 8: We don't really need the `--production` flag to `webpack` here if we're passing the env variable.

raphaelkabo (approved):
🚀 🦀 ✨ ❗

webpack.development.js
```diff
@@ -60,7 +60,7 @@ const browserConfig = (isLegacy = false) => ({
           new ForkTsCheckerWebpackPlugin({
             async: true,
             typescript: {
-              mode: 'write-references', // for better babel-loader perf.
+              mode: 'write-references',
```
mchv, line 63: @coldlink Could you clarify if `'write-references'` remains because it is improving perf of `swc-loader` as well or if there is another reason to keep it?

coldlink: Good spot, after digging into the plugin documentation, it seems that setting this property wasn't actually doing anything now or previously, so I'll remove it.

> Use readonly if you don't want to write anything on the disk, write-dts to write only .d.ts files, write-tsbuildinfo to write only .tsbuildinfo files, write-references to write both .js and .d.ts files of project references (last 2 modes requires build: true).

That documentation suggests that because we're not using the `build: true` property in this plugin, the `mode` property was ignored anyway.

https://github.com/TypeStrong/fork-ts-checker-webpack-plugin#typescript-options

PR #2143, opened 2023-01-31
Upgrade Express types

raphaelkabo:
## Accessibility

### DIFF
--- a/package.json
+++ b/package.json
@@ -63,7 +63,7 @@
     "@types/compression": "^1.7.2",
     "@types/cookie-parser": "^1.4.3",
     "@types/csurf": "^1.11.2",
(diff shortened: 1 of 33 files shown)


coldlink:
There was a reason we used the `res.locals`, as it's against express patterns to directly modify the response (or request) objects. 

https://expressjs.com/en/api.html#res.locals

```
The variables set on res.locals are available within a single request-response cycle, and will not be shared between requests.

This property is useful for exposing request-level information such as the request path name, authenticated user, user settings, and so on to templates rendered within the application.
```

raphaelkabo:
I'm going to redo this - the type errors I was getting before aren't happening now.

PR #2144, opened 2023-01-31
Dependencies | January 30 2023

raphaelkabo:
## What does this change?

Changes from previous https://github.com/guardian/gateway/pull/2124

- Tweaks timeout for in-band Jest tests as some were failing when running `make test`

Holds back the following:

* Keeps `react`, `react-dom` and `react-test-renderer` at v17 because v18 will break IE11
support going forward
* Keeps `@types/react-test-renderer` and `react-test-renderer` at v17 because v18 depends on
React 18
* Keeps `cypress` at v11 because the upgrade to v12 will be handled in a separate PR
* Keeps `@testing-library/cypress` at `8.x.x` as `9.x.x` relies on Cypress 12
* Keeps `query-string` at `7.x.x` as `8.x.x` is now pure ESM package

## Tested

- [ ] CODE
- [ ] IE11

### DIFF
--- a/src/server/lib/__tests__/headers.test.ts
+++ b/src/server/lib/__tests__/headers.test.ts
@@ -7,7 +7,7 @@ import {
 } from './sharedConfig';
 
 // Override the default 5s max timeout for these tests because Supertest takes some time to run.
-jest.setTimeout(10000);
+jest.setTimeout(20000);
 
 /**
  * We mock a section of our app infrastruture so that the routes that Supertest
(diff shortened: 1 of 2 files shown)


kelvin-chappell (commented):
Shouldn't there be a package.json file here?

coldlink:
Yeah somethings definitely not right in this PR, the `yarn.lock` file shouldn't be this drastically changed, and there should be a `package.json` change if you used the `make upgrade` or `yarn upgrade-interactive --latest` command

raphaelkabo:
Ah, the issue was that I didn't run `--latest` 🤦

PR #2147, opened 2023-02-06
Dependencies | February 6th 2023

coldlink:
## What does this change?

Changes from previous https://github.com/guardian/gateway/pull/2124

- Add `tslib` as this is now a peer dependency on most `@guardian/` packages, in turn adding the `importHelpers` flag to `tsconfig` files 

Holds back the following:

* Keeps `react`, `react-dom` and `react-test-renderer` at v17 because v18 will break IE11
support going forward
* Keeps `@types/react-test-renderer` and `react-test-renderer` at v17 because v18 depends on
React 18
* **NEW** Keeps `eslint-plugin-functional` at `4.x.x` as `5.x.x` has many breaking changes which will have to be handled separately
* **NEW** Keeps `@types/express` at `4.17.15` as `4.17.16` causes breaking types for some reason, which will be handled separately 
* **NEW** Keeps `@guardian/ab-*` packages at `2.0.1` as `3.x.x` causes a broken vendors bundle, which will have to be handled separately 

## Tested

- [x] CODE
- [x] IE11


raphaelkabo:
Why the need for `importHelpers`?

raphaelkabo (approved):
✅

PR #2153, opened 2023-02-06
fix(@types/express): update and fix express types

coldlink:
## What does this change?

There were some typing changes in express which caused things to break when we updated it, due to changes they'd made to their own types, which didn't affect functionality.

This updates the `ResponseWithRequestState` type to use the type generic instead of extending the `Response` object


raphaelkabo (approved):
This is working great now!

PR #2155, opened 2023-02-09
Filter consent/confirmation newsletters if AbDefaultWeeklyNewsletter switch is off

vlbee:
Until the default weekly newsletter is cancelled, it will continue to be returned as a user subscription in newsletter api responses. 

Gateway actively chooses which newsletters from that response to show via the[ Newsletter Enum in the model](https://github.com/guardian/gateway/blob/cc92cfed16ba59d406cee38056217828ba0e21c7/src/shared/model/Newsletter.ts#L22).  While the AbDfaultWeeklyNewsletterTtest switch is OFF but the AbTest code is still in the codebase/Enum, we need to filter out the newsletter from the Confirmation page. 

Otherwise there is a small edge case where one of the ~1000 people in the trial who signed up for SaturdayRoundUp when registering, manually navigates back to the onboarding pages and see 'Saturday Roundup' in the confirmation page but not the Newsletters page. 

This will not be an anymore issue when the Saturday Roundup test code is deleted from Gateway in the next couple weeks.  

image

## Accessibility

### DIFF
--- a/src/client/pages/ConsentsConfirmationPage.tsx
+++ b/src/client/pages/ConsentsConfirmationPage.tsx
@@ -2,6 +2,8 @@ import React from 'react';
 import useClientState from '@/client/lib/hooks/useClientState';
 import { ConsentsConfirmation } from '@/client/pages/ConsentsConfirmation';
 import { Consents } from '@/shared/model/Consent';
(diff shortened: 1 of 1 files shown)


lindseydew (approved):
LGTM

PR #2160, opened 2023-02-15
Refactor Sign Out Cookie (`GU_SO` cookie)

coldlink:
## What does this change?

In preparation for the Okta migration for web applications, we want to be able to tell when a user has signed out within the last 24 hours on a given device.

We do this by reusing the existing `GU_SO` cookie (signout cookie) which we previously got from Identity API which was set when a user signed out.

We now set this cookie from Gateway, avoiding the need for Identity API.

This cookie has a value which is the unix epoch time in seconds when the user signed out, and an expiry of 24 hours. It's also set as a non-HTTPOnly cookie so that it can be read in javascript applications.

This allows any applications when reading this cookie to assume that the user is currently signed out, and clear any tokens/user data required if it comes across this cookie.

This is initially part of the "spike" work to test the feasibility of the approach we decided, but if that goes well then this will rolled out.

### DIFF
--- a/src/server/lib/idapi/IDAPICookies.ts
+++ b/src/server/lib/idapi/IDAPICookies.ts
@@ -1,4 +1,4 @@
-import { Response } from 'express';
+import { Response, CookieOptions } from 'express';
 import { getConfiguration } from '@/server/lib/getConfiguration';
 
 interface IdapiCookie {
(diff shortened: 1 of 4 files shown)


kelvin-chappell (approved):
Looks like it will work to me 👍

PR #2161, opened 2023-02-20
Reduce Bundle Size | Use Preact

coldlink:
## What does this change?

1. Use [Preact](https://preactjs.com/).
2. Reduce bundle size
3. ??????
4. Profit!

We've been wanting to switch to Preact for a while for our client side framework since it provides the same API  as React, but much smaller. This reduces the bundle size, leading to quicket, faster, more efficient loading times for our end users!

We followed the [Switching to Preact (from React) guide](https://preactjs.com/guide/v10/switching-to-preact) in order to do this, so we could alias React completely without having to rewrite everything, mostly by using `preact/compat`.

Most of the PR is just aliasing in places like package.json, webpack, storybook, jest, and typescript in order to get it all recognised. We also changed `@testing-library/react` to `@testing-library/preact` (and the hooks equivalent) in order to get tests working too.

We then fix the recaptcha unit tests, which needing minor changes in order for the tests to pass, specifically correctly mocking the callback function, and adding a `waitFor` statement (with a 50ms interval) in order to wait for the callback, specifically the recaptcha render function to fire.

The only other issue then encountered was with the `EmailInput` unit tests where a couple of the tests failed to pass and we were unable to fix, despite these working as expected in browser, storybook, and Cypress. We decided to skip these tests, and add on some Cypress tests which checked the same functionality. If anyone would like to take a look at why these two tests aren't working in jest only then please have a go! We're confident enough that the `EmailInput` component is working as excepted.

Our Cypress tests provide good confidence in the PR working, as we cover all possible user scenarios in our testing.

We've also tested this in CODE to make sure it works in modern and legacy browsers too.

### `main` branch comparison

This compares the bundle to the current main branch at the time of writing which is commit https://github.com/guardian/gateway/commit/e8bbae20f959b824707e6db07e335ca23bdbbb96

| | Before | After | Difference | % Change |
| - | - | - | - | - |
| Modern | 320.91 kB (103.91 KB kB gzipped) | 216.03 kB (71.31 KB kB gzipped) | 104.88 kB (32.60 kB gzipped) | 32.68% (31.37% gzipped) |
| Legacy | 379.17 kB (107.48 kB gzipped) | 276.75 kB (75.24 kB gzipped) | 102.42 kB (32.24 kB gzipped) | 27.01% (30.00% gzipped) |

### #2115 comparison

This compares the bundle to before we started the reduce bundle size work where we added the bundle analyzer, which is commit https://github.com/guardian/gateway/commit/da173ab166ed0be7b3e46dc3194ce59df237d925

| | Before | After | Difference | % Change |
| - | - | - | - | - |
| Modern | 457.07 kB (138.42 kB gzipped) | 216.03 kB kB (71.31 KB kB gzipped) | 241.01 kB (67.11 kB gzipped) | 52.74% (48.48% gzipped) |
| Legacy* | 1170 kB (302.8 kB gzipped) | 276.75 kB (75.24 kB gzipped) | 893.25 kB (227.56 kB gzipped) | 76.35% (75.15% gzipped) |

\* Bundle size was drastically reduced in legacy bundle as part of swc work in #2132


jest.config.js
```diff
@@ -7,6 +7,15 @@ module.exports = {
     '@/([^\\.]*)$': '<rootDir>/src/$1',
     '\\.(jpg|ico|jpeg|png|gif|eot|otf|webp|svg|ttf|woff|woff2|mp4|webm|wav|mp3|m4a|aac|oga)$':
       '<rootDir>/__mocks__/fileMock.js',
+    '^react$': 'preact/compat',
```
sndrs, line 10: is this still required when you have `"react": "npm:@preact/compat"` in your package.json?

coldlink: Ah good spot! I was just aliasing everything everywhere.

Just tried without it and it's working!

Was also able to remove the `react-dom` alias too.

Also removed `@testing-library/react` and `@testing-library/react` by using the preact equivalents.

.storybook/main.js
```diff
@@ -27,6 +27,10 @@ module.exports = {
     config.resolve.alias = {
       ...config.resolve.alias,
       '@': path.join(__dirname, '../src'),
+      react: 'preact/compat',
+      'react-dom/test-utils': 'preact/test-utils',
+      'react-dom': 'preact/compat', // Must be below test-utils
+      'react/jsx-runtime': 'preact/jsx-runtime',
```
sndrs, line 33: i _think_ you might get the same result as with the jest aliasing here. by using the npm aliasing, anytime time node tries to find `react` it will instead find `@preact/compat`, so you may not need these either 🤞

package.json
```diff
@@ -53,8 +53,8 @@
     "@swc/plugin-emotion": "^2.5.43",
     "@testing-library/cypress": "^9.0.0",
     "@testing-library/jest-dom": "^5.16.5",
-    "@testing-library/react": "^12.1.4",
-    "@testing-library/react-hooks": "^7.0.2",
+    "@testing-library/preact": "^3.2.3",
+    "@testing-library/preact-hooks": "^1.1.0",
```
sndrs, line 57: would it help to alias these like you have react?

```suggestion
    "@testing-library/preact": "^3.2.3",
    "@testing-library/preact-hooks": "^1.1.0",
    "@testing-library/react": "npm:@testing-library/preact",
    "@testing-library/react-hooks": "npm:@testing-library/preact-hooks",
```

then the tests can just carry on as if they are using react, as before.

it might help to keep everything feeling like you're using react, even if ultimately you aren't.

.storybook/main.js
coldlink: I did try this after your previous suggestion in jest, but unfortunately it didn't work in either the storybook or webpack config.

package.json
coldlink: I suppose it is confusing in tests if it says "@testing-library/preact" rather than "@testing-library-react"! I'll make a change.

Mark-McCracken (approved):

PR #2163, opened 2023-02-21
Migrate `robots.txt` from Identity Frontend

coldlink:
## What does this change?

Migrates the `robots.txt` file from Identity Frontend to Gateway

https://github.com/guardian/identity-frontend/blob/main/public/robots.txt

VCL Changes to follow

### DIFF
--- a/src/server/routes/core.ts
+++ b/src/server/routes/core.ts
@@ -30,4 +30,12 @@ router.get('/healthcheck', (_, res: Response) => {
   res.status(200).send('200 OK');
 });
 
+router.use('/robots.txt', (_, res: Response) => {
+  res.type('text/plain');
+  res.send(`User-agent: *
+Disallow: /user/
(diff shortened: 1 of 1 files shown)


src/server/routes/core.ts
```diff
@@ -30,4 +30,12 @@ router.get('/healthcheck', (_, res: Response) => {
   res.status(200).send('200 OK');
 });
 
+router.use('/robots.txt', (_, res: Response) => {
+  res.type('text/plain');
+  res.send(`User-agent: *
+Disallow: /user/
+Disallow: /signin?*returnUrl=https://discussion.theguardian.com/comment-permalink/
```
kelvin-chappell, line 37: Could this just be `/signin?*` or are there some queries that we want to be crawled?

coldlink: I have a feeling that it's fine to crawl sign-in etc, but this was specifically set up to make sure specific comments weren't crawled for some reason

kelvin-chappell (approved):
👍

PR #2165, opened 2023-02-21
Migrate consent token endpoint from Identity Frontend to Gateway

raphaelkabo:
## What does this change?

Migrates the 'confirm your subscription'/'confirm your communication preferences' journey from Identity Frontend to Gateway.

On the frontend, created a new page:

**Expired/invalid consent token** - `ResendConsentTokenPage`

Screenshot 2023-02-23 at 11 35 10

And refactored the `EmailSent` page to show a user help text if a `showHelp` flag is present (previously, it would only show the help if specific conditions were met):

Screenshot 2023-02-23 at 11 34 48

On the backend, added the following routes:

- `GET /consent-token/:token/accept`
  - Receives a token and validates it with IDAPI.
  - If the token is valid, redirects the user to an external page managed by Frontend: `/consents/thank-you`.
  - If the token is invalid, displays the new `ResendConsentTokenPage` with the token prefilled into the form, which POSTs to `/consent-token/resend`.

- `POST /consent-token/resend`
  - Receives a token in the post body.
  - Attempts to request IDAPI to resend the email for that token.
  - Whether the request succeeds or fails, redirects the user to `/consent-token/email-sent`.

- `GET /consent-token/email-sent`
  - Displays the `EmailSentPage` with the user help text shown, as above.
 
Also set up Cypress ETE tests to cover this journey.

## Testing

- [x] CODE
- [ ] PROD

1. In Postman or User Admin tool, `POST` to the IDAPI `/consent-email` endpoint with a JSON body of the form:
  ```
  {
      "email": "email@domain.com",
      "set-consents": ["jobs"],
     // OR
      "set-lists": ["morning-briefing"]
  }
  ```
2. The specified email address will receive an email with a link to click to confirm the new consents.
3. The link should point to the Gateway route `/consent-token/:token/accept`, which will then redirect the user to the Frontend route `/consents/thank-you`.
4. A direct query to `/consent-token/nonsense-token/accept` will display a 'link expired' page (again, all in Gateway) and clicking on the 'resend' button will redirect to the page `/consent-token/email-sent` (no email will be sent as the consent token is fake).

## Related PRs

After this PR is merged, we will want to merge in the Fastly VCL changes in https://github.com/guardian/identity-platform/pull/615.

### DIFF
--- a/cypress/integration/ete/consent_token/consent_token.2.cy.ts
+++ b/cypress/integration/ete/consent_token/consent_token.2.cy.ts
@@ -1,0 +1,46 @@
+describe('Consent token flow', () => {
+  it('redirects to the success route when supplied a valid token by a logged in user', () => {
+    cy.createTestUser({
+      isUserEmailValidated: true,
+    }).then(({ emailAddress, cookies }) => {
+      // SC_GU_U is required for the cy.getTestUserDetails
+      const scGuU = cookies.find((cookie) => cookie.key === 'SC_GU_U');
+      if (!scGuU) throw new Error('SC_GU_U cookie not found');
+      cy.setCookie('SC_GU_U', scGuU?.value);
+      // SC_GU_LA is required for the cy.getTestUserDetails
+      const scGuLa = cookies.find((cookie) => cookie.key === 'SC_GU_LA');
+      if (!scGuLa) throw new Error('SC_GU_LA cookie not found');
+      cy.setCookie('SC_GU_LA', scGuLa?.value);
+      cy.sendConsentEmail({
+        emailAddress,
+        consents: ['jobs'],
+      }).then(() => {
+        const timeRequestWasMade = new Date();
+        cy.checkForEmailAndGetDetails(
+          emailAddress,
+          timeRequestWasMade,
+          /consent-token\/([^"]*)/,
+        ).then(({ token }) => {
+          cy.visit(`/consent-token/${token}`, {
+            failOnStatusCode: false,
+          });
+          // /consents/thank-you isn't hosted by Gateway so all we need to check
+          // is that the cy.visit() redirected successfully - this will show a 404 page.
+          cy.url().should(
+            'be.equal',
+            `https://${Cypress.env('BASE_URI')}/consents/thank-you`,
+          );
+        });
+      });
+    });
+  });
+
+  it('shows the email sent page when supplied an invalid token', () => {
+    const invalidToken = 'invalid-consent-token';
+    cy.visit(`/consent-token/${invalidToken}/accept`);
+    cy.contains('This link has expired.');
+    cy.get('button[type=submit]').click();
+    cy.url().should('include', '/consent-token/email-sent');
+    cy.contains('Check your email inbox');
+  });
+});
--- a/src/client/pages/EmailSent.tsx
+++ b/src/client/pages/EmailSent.tsx
@@ -23,6 +23,7 @@ type Props = {
   recaptchaSiteKey?: string;
   formTrackingName?: string;
   formError?: string;
+  showHelp?: boolean;
 };
 
 export const EmailSent = ({
@@ -37,10 +38,12 @@ export const EmailSent = ({
   formTrackingName,
   children,
   formError,
+  showHelp,
 }: PropsWithChildren<Props>) => {
   const [recaptchaErrorMessage, setRecaptchaErrorMessage] = useState('');
   const [recaptchaErrorContext, setRecaptchaErrorContext] =
     useState<ReactNode>(null);
+  const showHelpBox = showHelp || (email && resendEmailAction);
   return (
     <MainLayout
       pageHeader="Check your email inbox"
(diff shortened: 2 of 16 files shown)


raphaelkabo:
[Migrate Identity Frontend to Gateway](https://trello.com/c/O2mv8e90/4297-migrate-identity-frontend-to-gateway)

src/server/routes/consentToken.ts
```diff
@@ -0,0 +1,73 @@
...
+  '/consent-token/:token/accept',
+  handleAsyncErrors(async (req: Request, res: ResponseWithRequestState) => {
+    const ip = req.ip;
+    const sc_gu_u = req.cookies.SC_GU_U;
+    const token = req.params.token;
+    try {
+      await validateConsentToken(ip, sc_gu_u, token, res.locals.requestId);
+      // Redirect to /consents/thank-you (a page managed by Frontend). This is
+      // to retain the legacy behaviour of the route from identity-frontend.
+      return res.redirect('/consents/thank-you');
```
coldlink, line 25: Slightly pedantic, but we should use 303 redirects as that makes the most sense: https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/303

Rather than the Express default which is 301: https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/301

Same suggestion for the other redirects in this file.

```suggestion
      return res.redirect(303, '/consents/thank-you');
```

src/server/routes/consentToken.ts
```diff
@@ -0,0 +1,73 @@
...
+    // This is a pass-through route to IDAPI, which will send a new consent
+    // email to the user if the token matches. We don't need to do anything
+    // with the response here.
+    const { token } = req.body;
+    const ip = req.ip;
+    const sc_gu_u = req.cookies.SC_GU_U;
+    try {
+      await resendConsentEmail(ip, sc_gu_u, token, res.locals.requestId);
+      return res.redirect('/consent-token/email-sent');
+    } catch (error) {
```
coldlink, line 65: can we add a `logger` here, so we can see the error
```suggestion
import { logger } from '@/server/lib/serverSideLogger';
...
    } catch (error) {
    logger.error(`${req.method} ${req.originalUrl}  Error`, error, {
        request_id: res.locals.requestId,
      });
```

src/server/routes/consentToken.ts
```diff
@@ -0,0 +1,73 @@
...
+      // On an error, we redirect to the same page as if the request was
+      // successful.
+      return res.redirect('/consent-token/email-sent');
+    }
```
coldlink, line 69: we don't need `return res.redirect('/consent-token/email-sent');` twice here in this route if we use a finally! 
We can remove the one in both the try and catch
```suggestion
    } finally {
    // Redirect to same page on either success or failure
      return res.redirect('/consent-token/email-sent');
```

src/client/pages/EmailSent.tsx
```diff
@@ -110,21 +112,22 @@ export const EmailSent = ({
...
-              `}
-              href="mailto:userhelp@theguardian.com"
-            >
-              userhelp@guardian.com
-            </ExternalLink>
-            .
-          </MainBodyText>
         </>
       )}
+      {(showHelp || (email && resendEmailAction)) && (
```
coldlink, line 117: Move this into a variable before the `return` statement

```suggestion
const showHelpBox = showHelp || (email && resendEmailAction);
...
      {showHelpBox && (
```

src/server/lib/queryParams.ts
```diff
@@ -33,6 +33,7 @@ export const parseExpressQueryParams = (
     componentEventParams,
     fromURI,
     appClientId,
+    consentToken,
```
coldlink, line 36: does the `consentToken` need to be here, and below?

I don't see it being used as a query parameter anywhere

src/server/lib/idapi/consentToken.ts
```diff
@@ -0,0 +1,74 @@
...
+    APIAddClientAccessToken(APIPostOptions(), ip),
+    sc_gu_u,
+  );
+  try {
+    await idapiFetch({
+      path: '/consent-email/resend/:token',
+      options,
+      tokenisationParam: { token },
+    });
+    return;
```
coldlink, line 63: don't need `return` here though it's fine to keep, same in `validateConsentToken`
```suggestion
```

coldlink (changes requested):

coldlink:
Great work! Just a few suggestions that I have.

cypress/integration/ete/consent_token/consent_token.2.cy.ts
```diff
@@ -0,0 +1,48 @@
...
+      // Sign our new user in
+      cy.visit(
+        `/signin?returnUrl=${encodeURIComponent(
+          `https://${Cypress.env('BASE_URI')}/consents`,
+        )}`,
+      );
+      cy.get('input[name=email]').type(emailAddress);
+      cy.get('input[name=password]').type(finalPassword);
+      cy.get('[data-cy="main-form-submit-button"]').click();
+      cy.url().should('include', '/consents');
```
coldlink, line 15: this can become 
```suggestion
    cy.createTestUser({
      isUserEmailValidated: true,
    }).then(({ emailAddress, cookies }) => {
      // SC_GU_U is required for the cy.getTestUserDetails
      const scGuU = cookies.find((cookie) => cookie.key === 'SC_GU_U');
      if (!scGuU) throw new Error('SC_GU_U cookie not found');
      cy.setCookie('SC_GU_U', scGuU?.value);
```

raphaelkabo:
Thank you, all incorporated!

src/server/routes/consentToken.ts
```diff
@@ -0,0 +1,76 @@
...
+  handleAsyncErrors(async (req: Request, res: ResponseWithRequestState) => {
+    // This is a pass-through route to IDAPI, which will send a new consent
+    // email to the user if the token matches. We don't need to do anything
+    // with the response here.
+    const { token } = req.body;
+    const ip = req.ip;
+    const sc_gu_u = req.cookies.SC_GU_U;
+    try {
+      await resendConsentEmail(ip, sc_gu_u, token, res.locals.requestId);
+      return res.redirect(303, '/consent-token/email-sent');
```
coldlink, line 65:
```suggestion
```

this line is redundant thanks to the `finally`

coldlink (approved):

coldlink:
:shipit:

PR #2172, opened 2023-03-03
Dependencies | Migrate to AWS SDK v3

coldlink:
## What does this change?

`aws-sdk` was showing warning while running gateway that v2 will be deprecated and we should migrate to v3. This PR does just that.

Followed the upgrade guide: https://github.com/aws/aws-sdk-js-v3/blob/main/UPGRADING.md

And looked at the new documentation here: https://docs.aws.amazon.com/AWSJavaScriptSDK/v3/latest/

## Tested?
- [ ] CODE


Mark-McCracken (approved):
Nice work, all good

PR #2176, opened 2023-03-06
Dependencies | March 6th 2023

coldlink:
## What does this change?

Changes from previous https://github.com/guardian/gateway/pull/2168

Holds back the following:

* Keeps `@types/react` and `@types/react-dom` at v17 as we're using `preact/compat`
* Keeps `eslint-plugin-functional` at `4.x.x` as `5.x.x` has many breaking changes which will have to be handled separately
* Keeps `@guardian/ab-*` packages at `2.0.1` as `3.x.x` causes a broken vendors bundle, which will have to be investigated separately

## Tested

- [x] CODE
- [x] IE11


kelvin-chappell (approved):
👍

PR #2185, opened 2023-03-14
Remove Facebook from client side UI

coldlink:
## What does this change?

This PR removes the Facebook social button from the sign in and registration pages, and replaces it with copy detailing how a user who previously used Facebook should now authenticate.

It is only removed from the client side for now, with separate PRsto follow to remove the server route and functionality, so this will be rolled out if phases.

This PR:
1a. UI Changes
1b. Add tracking for new reset password change

To follow:
2. Removing /signin/facebook route and server side changes
3. Remove the UI change after some time has passed

## Tested

- [ ] CODE
- [ ] IE11


kelvin-chappell (approved):
👍

PR #2187, opened 2023-03-16
Upgrade React types to v18

coldlink:
## What does this change?

Updates `@types/react` and `@types/react-dom` to v18 in prep for the department wide upgrade to React 18.

Since we're using `preact/compat` the only things we should need to change are the type definitions, as everything else behaves as it does in React 17, and the build is the same.

The main type changes we needed to make were to explicitly define `children` using `PropsWithChildren` in some of our components.

https://reactjs.org/blog/2022/03/08/react-18-upgrade-guide.html#updates-to-typescript-definitions

The only other issue resided with our use of `mjml-react` which was showing type errors.

On further investigation, the `mjml-react` project has been unmaintained for a while: https://github.com/wix-incubator/mjml-react

It recommended that we switch to a fork called `@faire/mjml-react`: https://github.com/Faire/mjml-react

This fixed all the type issues, and other problems we had with `mjml-react` previously. There were no changes to visual styles too according to chromatic, which is a good sign!

Overall it was a pleasant enough upgrade to React 18 types!

## Tested

This is for the `mjml-react` changes.

- [x] CODE


Mark-McCracken (approved):
All looks good to me, top stuff.

PR #2191, opened 2023-03-20
Remove Facebook from server side functionality

coldlink:
## What does this change?

Remove all facebook functionality from the server side of Gateway

- Route
- Config
- Environment variables


kelvin-chappell (approved):
LGTM 👍

PR #2194, opened 2023-03-20
Update Gateway to use ARM64 instance types in AWS

coldlink:
## What does this change?

Updates gateway to use the `t4g.small` instance types, and an ARM64 based AMI.

## Tested
- [x] CODE

Deployed successfully.
Tested by deploying to CODE, ssh into the new instance and checking logs working, while performing actions on the CODE site, e.g. sign in, reset password, onboarding etc.


pvighi (approved):
looks good

PR #2198, opened 2023-03-27
Dependencies | March 27th 2023

coldlink:
## What does this change?

Changes from previous https://github.com/guardian/gateway/pull/2190

Holds back the following:

* Keeps `typescript` at `~4.9.5` as this is the version we use in the department
* Keeps `preact-render-to-string` at `5.x.x` as `preact/compat` has not yet updated to use `6.x.x`

## Tested

- [x] CODE
- [x] IE11


raphaelkabo (approved):
✅

PR #2216, opened 2023-04-11
Dependencies | April 11th 2023

coldlink:
## What does this change?

Changes from previous https://github.com/guardian/gateway/pull/2203

Holds back the following:

* Keeps `typescript` at `~4.9.5` as this is the version we use in the department
* Keeps `@storybook` packages at v6 as v7 is a major change which will have to be separated into its own PR

## Tested

- [x] CODE
- [x] IE11


Mark-McCracken (approved):
👍

PR #2217, opened 2023-04-11
Revert swc version to fix oob errors

coldlink:
## What does this change?

Fixes `RuntimeError: out of bounds memory access` from `swc` introduced in #2216 for local development on ARM machines. This did not break the production build, but since we can't develop locally otherwise, I'm reverting this.


raphaelkabo (approved):
✅

PR #2245, opened 2023-05-05
WIP: Migrate Gateway routes to Okta

coldlink:
## What does this change?

*tl;dr - migrates things which were previously using IDAPI cookies to authenticate to use OAuth tokens instead*

This is a big PR, I'll explain things in this description as best as possible. You can also follow commit by commit which may make this easier to review.


coldlink:
Making a new PR with fixed tests

PR #2257, opened 2023-05-11
Remove Google Analytics

coldlink:
## What does this change?

Remove GA as we're not upgrading to GA4

Less JS for the client to load, less permissions required for Gateway too.

Also fixes the Cypress Github Action permissions, due to an issue encountered there. It seems the parent action needs the same permissions as the child action.

The issue:
```
The workflow is not valid. .github/workflows/cypress.yml (Line: 45, Col: 3): Error calling workflow 'guardian/gateway/.github/workflows/cypress-ete-okta.yml@62e1b8f9b30c0816bc0e01589f777d26ecaad5dc'. The nested job 'Cypress End to End tests against Okta' is requesting 'id-token: write', but is only allowed 'id-token: none'.
```


src/server/lib/__tests__/headers.test.ts
```diff
@@ -70,48 +70,17 @@ describe('Content Security Policy headers', () => {
...
-          'www.google.com',
-          'www.gstatic.com',
-          'assets.guim.co.uk',
-          "'unsafe-eval'",
-        ].join(' '),
+        `script-src ${baseUri} sourcepoint.theguardian.com gdpr-tcfv2.sp-prod.net ccpa.sp-prod.net ccpa-service.sp-prod.net ccpa-notice.sp-prod.net cdn.privacy-mgmt.com www.google.com www.gstatic.com assets.guim.co.uk 'unsafe-eval'`,
       );
       expect(splitCSPHeader).toContain(
-        `img-src ${baseUri} static.guim.co.uk ophan.theguardian.com www.google-analytics.com www.google.com`,
+        `img-src ${baseUri} static.guim.co.uk ophan.theguardian.com www.google.com`,
```
raphaelkabo, line 79: Is the `www.google.com` here modelling the reCAPTCHA script?

coldlink: yup https://github.com/guardian/gateway/blob/3ce27f49fa3aac254b4e02960ca734c36387a750/src/server/lib/middleware/helmet.ts#L15

raphaelkabo (approved):
GA begone!

PR #2265, opened 2023-05-22
Migrate Gateway routes to Okta OAuth

coldlink:
This is a big PR, I'll explain things in this description as best as possible. You can also follow commit by commit which may/*will* make this easier to review.

## What does this change?

*tl;dr - migrates things which were previously using IDAPI cookies to authenticate to use OAuth tokens instead*

We have enabled this via a feature switch (`gatewayOAuthEnabled`) on `DEV` and `CODE` so we can start developing against it, and so we can merge this change in.
 
1. Set's up the `@okta/jwt-verifier`, and adds methods/functionality in order to validate and use Okta OAuth access and ID tokens, specifically access and id tokens.
    - The tokens are stored as HTTPOnly, HostOnly (`profile` subdomain only), Secure, Signed cookies valid for 1 hour. `GU_ACCESS_TOKEN`, and `GU_ID_TOKEN`
    - Note, this is different from the `openid-client` which we also (already) use, which is used to perform the Authorization Code Flow and get tokens, but doesn't do token verification itself. 
2. Updates the `performAuthorizationCodeFlow` method to be more flexible, as we need it to support the two different types of OAuth tokens we need to support now, and have two Authorization Code callbacks in order to support this:
    - `AUTHENTICATION` - Used for sign in/set password/registration, uses the `/oauth/authorization-code/callback` callback URI, tokens are not stored, and are short lived (5 mins) used to get IDAPI cookies, read user products (for ad-free cookie), and to check user consents to see if we should redirect them to the  post sign in prompt
    - `APPLICATION` - Used for internal user signed in routes, e.g. onboarding flow, jobs pages (`/agree/GRS`), uses the new `/oauth/authorization-code/application-callback` callback URI, tokens are stored, and is used by internal routes as well as making calls to external APIs, e.g. Identity API. Tokens as mentioned, live for 1 hour, refreshed by going through the authorization code flow again
3. Set's up a new "login middleware" to handle OAuth tokens `loginMiddlewareOAuth`, this middleware is used by a route if it needs a user to be logged in and have valid oauth tokens
    - This checks a users login status based on the diagram to check signed in status using the ["web apps integration guide"](https://github.com/guardian/gateway/blob/main/docs/okta/web-apps-integration-guide.md#how-to-know-if-a-reader-is-signed-in)
    - In our case we don't need to check for the `GU_U` or `sid` cookies, as there routes require sign in, so we always perform the authorization code flow to get tokens if they don't exist on these routes.
    - This has a fallback to the old loginMiddleware if using the `useIdapi=true` query parameter flag
4. Updates IDAPI calls to allow for the use of OAuth Access Tokens
    -  Adds `APIAddOAuthAuthorization` helper method, and `APIOptionSelect` which takes in optional sc_gu_u cookie or access token and returns the option header depending on which one was provided
    - Updates select IDAPI calls to use this, `readUserConsents` `update` (consents), `readUserNewsletters`, and `update` (newsletters)
5. Updates the MDAPI `/user-attributes/me` endpoint to allow for the use of okta access tokens or sc_gu_u cookie 
6. Allows for updating the users location field directly in Okta without having to go through IDAPI. `updateRegistrationLocationViaOkta`
7. Refactors all the `/consents` (onboarding) routes to allow for the use of okta access tokens for reading/updating user consents/newsletters
8. Updates the post sign in prompt to allow for the use of okta access tokens for reading/updating the user consent
9. Updates the `AUTHENTICATION` authorization code callback URL to use the access token for more things, namely the MDAPI `/user-attributes/me` endpoint to check if we should set the ad free cookie, and read user's consents to determine if the post sign in prompt should show
10. Updates the jobs (`/agree/GRS`) routes to use the new `loginMiddlewareOauth` instead of manually checking for the Okta session cookie to update a users job information
11. Updates signout to make sure that the OAuth tokens are removed when the user signs out
12. Fix a bunch of tests which were changed, especially in the `mocked` server tests due to changes in API calls and order, the `e2e` tests against IDAPI to make sure that the `useIdapi=true` flag is always set, and the `e2e` tests against Okta for the jobs flow.
13. Adds `e2e` tests against Okta to test the onboarding flow, and post sign in prompt changes.
14. Fixes `nginx` issues due to large response headers
15. Fixes `SignedInAs` page to show the global error should one exist.
16. Fixes other components where query parameters weren't being preserved
17. Set up feature switch (`gatewayOAuthEnabled`) on `DEV` and `CODE` so we can start developing against it, and so we can merge this change in.

## Testing plan
Deploy branch to CODE.

Things to test:
- Sign in and post sign in prompt work successfully, and the `GU_ACCESS_TOKEN` and `GU_ID_TOKEN` is set with 1h expiry when going to the post sign in prompt
- Registration and onboarding flow work successfully, and the `GU_ACCESS_TOKEN` and `GU_ID_TOKEN` is set with 1h expiry when going to the onboarding flow
- Jobs T&C page works successfully, and the `GU_ACCESS_TOKEN` and `GU_ID_TOKEN` is set with 1h expiry during this flow
- Sign out clears the OAuth tokens
- Other functionality should work as expected, e.g. reset password, set password
- Functionality still works in the old system when using the `useIdapi=true` flag


src/server/lib/okta/oauth.ts
```diff
@@ -5,11 +5,36 @@ import { RoutePaths } from '@/shared/model/Routes';
...
 } from '@/server/lib/okta/openid-connect';
 import { closeSession } from './api/sessions';
 
+type Scopes =
+  | 'openid'
+  // the `guardian.identity-api.cookies.create.self.secure scope` is checked on IDAPI to return legacy identity cookies on authentication
+  | 'guardian.identity-api.cookies.create.self.secure';
+
+/**
+ * @param closeExistingSession (optional) - if true, we'll close any existing okta session before calling the authorization code flow
```
raphaelkabo, line 54: Is this default `false`? Might be good to say.

src/server/lib/middleware/login.ts
```diff
@@ -10,9 +10,78 @@ import { logger } from '@/server/lib/serverSideLogger';
...
+  next: NextFunction,
+) => {
+  if (res.locals.queryParams.useIdapi) {
+    return loginMiddleware(req, res, next);
+  }
+
+  // if a user has the GU_SO cookie, they have recently signed out
+  // so we need to clear any existing tokens
+  // and perform the auth code flow to get new tokens to
+  // get new tokens and log the user in if necessary
```
raphaelkabo, line 42: May have been fixed in a later commit: should be "and perform the auth code flow to get new tokens and log the user in if necessary"

src/server/lib/okta/tokens.ts
```diff
@@ -0,0 +1,167 @@
...
+ * Name of the OAuth ID Token cookie to set
+ * @type {string}
+ */
+const OAuthIdTokenCookieName = 'GU_ID_TOKEN';
+
+const issuerDomain = () => {
+  // if we're in cypress mocked, we need to add https to the issuer domain, just to get it
+  // to validate, even if we never actually use the verifier in cypress mocked tests
+  if (process.env.RUNNING_IN_CYPRESS_MOCKED === 'true') {
+    return okta.orgUrl.replace('http://', 'https://');
```
raphaelkabo, line 38: Why does `okta.orgUrl` have 'http' in it in the first place? Is that for local testing?

src/server/lib/okta/tokens.ts
```diff
@@ -0,0 +1,167 @@
...
+/**
+ * Token cookie options
+ * Expiry is set to 1 hour by default, which matches the token expiry
+ */
+const OAuthTokenCookieOptions = (): CookieOptions => ({
+  httpOnly: true,
+  secure: !baseUri.includes('localhost'),
+  signed: !baseUri.includes('localhost'),
+  sameSite: 'lax',
+  maxAge: 3600000,
```
raphaelkabo: There's no reason why we shouldn't move these numbers into constants or types somewhere central. Seems like a more reliable way to ensure we get the right times!

src/server/lib/okta/tokens.ts
coldlink: yeah, its for mocked local testing, where we sometimes need to check if we've redirected to the oauth flow using the mocked server, or return a generic response to test specific error cases

src/server/routes/oauth.ts
```diff
@@ -296,14 +330,101 @@ router.get(
...
+
+      // track the success metric
+      trackMetric('OAuthAuthorization::Success');
+
+      const returnUrl = authState.confirmationPage
+        ? addQueryParamsToPath(
+            authState.confirmationPage,
+            authState.queryParams,
+          )
+        : authState.queryParams.returnUrl;
```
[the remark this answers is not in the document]
coldlink: thats right, the `returnUrl` is generally links to external pages, e.g. theguardian.com/MMA etc, where we don't need to preserve gateway specific query params.

the `confirmationPage` is an internal gateway page where we may need query parameters

src/server/lib/updateRegistrationLocation.ts
```diff
@@ -7,12 +7,9 @@ import {
   read as readIdapiUser,
   addRegistrationLocation,
 } from '@/server/lib/idapi/user';
+import { Jwt } from '@okta/jwt-verifier';
+import { getUser, updateUser } from '@/server/lib/okta/api/users';
 
-/**
- * Until Gateway/Onboarding journey is migrated to Okta sessions, we don't have access to Okta User ID, only sg_gu_u cookie,
```
raphaelkabo, line 12: Should be `sc_gu_u cookie`

cypress/integration/ete/terms/jobs_terms.4.cy.ts
```diff
@@ -16,11 +16,11 @@ describe('Jobs terms and conditions flow', () => {
       // load the consents page as its on the same domain
       const termsAcceptPageUrl = `https://${Cypress.env(
         'BASE_URI',
-      )}/agree/GRS?returnUrl=https://profile.thegulocal.com/healthcheck`;
+      )}/agree/GRS?useIdapi=true`;
       cy.visit(termsAcceptPageUrl);
       cy.url().should(
         'include',
-        'https://profile.thegulocal.com/signin?returnUrl=https%3A%2F%2Fprofile.thegulocal.com%2Fhealthcheck',
+        'https://profile.thegulocal.com/signin?returnUrl=https%3A%2F%2Fprofile.thegulocal.com%2Fagree%2FGRS&useIdapi=true',
```
raphaelkabo, line 23: Hah, nice catch...

raphaelkabo (commented):
First pass review done, all looking really solid overall! ✅

src/server/routes/verifyEmail.ts
```diff
@@ -183,6 +183,9 @@ router.get(
           addQueryParamsToPath(
             `${consentPages[0].path}`,
             res.locals.queryParams,
+            {
+              useIdapi: true,
```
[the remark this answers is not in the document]
coldlink: It's because the `/verify-email` routes are only IDAPI specific, and only set the IDAPI cookies when a user verifies their email using this flow, so no Okta session. So when redirecting to the onboarding flow from the `/verify-email` routes we have to make sure that `useIdapi` param is set so that it uses the makes sure to use IDAPI logic to check if the user is logged in, and not the Okta session.

src/server/lib/postSignInController.ts
```diff
@@ -38,7 +48,7 @@ const postSignInController = async (
       ({ key }) => key === 'SC_GU_U',
     )?.value;
 
-    if (!sc_gu_u) {
+    if (!sc_gu_u && !oauthTokens?.access_token) {
```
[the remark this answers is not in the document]
coldlink: Updated to make it clearer to read!

src/server/routes/oauth.ts
```diff
@@ -296,14 +330,101 @@ router.get(
...
           request_id: res.locals.requestId,
         });
+        trackMetric('OAuthAuthorization::Failure');
+        return redirectForGenericError(req, res);
+      }
+
+      const [tokenSet, authState] = result;
+
+      // this is just to handle potential errors where we don't get back an access token
+      if (!tokenSet.access_token || !tokenSet.id_token) {
```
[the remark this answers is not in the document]
coldlink: Updated to make it clearer to read!

src/server/lib/middleware/login.ts
coldlink: Updated the comment

src/server/routes/oauth.ts
```diff
@@ -54,102 +56,134 @@ const isOAuthError = (
...
+      callbackParams.error === OpenIdErrors.ACCESS_DENIED &&
+      callbackParams.error_description ===
+        OpenIdErrorDescriptions.ACCOUNT_LINKING_DENIED_GROUPS
+    ) {
+      res.redirect(
+        addQueryParamsToPath('/signin', authState.queryParams, {
+          error: FederationErrors.SOCIAL_SIGNIN_BLOCKED,
+        }),
+      );
+      return;
```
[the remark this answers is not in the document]
coldlink: see https://github.com/guardian/gateway/pull/2265#discussion_r1210000567

src/server/routes/oauth.ts
```diff
@@ -54,102 +56,134 @@ const isOAuthError = (
...
-      // check for specific oauth errors and handle them as required
-      if (isOAuthError(callbackParams)) {
-        // check if the callback params contain an login_required error
-        // used to check if a session existed before the user is shown a sign in page
-        if (callbackParams.error === OpenIdErrors.LOGIN_REQUIRED) {
-          return res.redirect(
-            addQueryParamsToPath('/signin', authState.queryParams),
-          );
-        }
+      if (res.headersSent) {
```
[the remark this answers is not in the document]
coldlink: ok this breaks a bunch of tests andi don't know why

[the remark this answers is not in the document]
coldlink: All fixed, this commit is the new one: [`0cce3a7` (#2265)](https://github.com/guardian/gateway/pull/2265/commits/0cce3a7ff327b1d3bdedccf78239b3c540deb829)

raphaelkabo (approved):
The new refactor is great, really nice work. It all looks good to me - as long as tests in CODE pass, I'm happy for this to go live!

raphaelkabo (approved):
✅ Easy, straightforward, love it. Great work!

PR #2284, opened 2023-06-13
Dependencies | June 20th 2023 | TypeScript 5, Storybook 7, and others

coldlink:
## What does this change?

- Upgrades to TypeScript v5 now that `@guardian/*` packages use it
- Upgrades to Storybook 7, mostly following the [automated](https://storybook.js.org/docs/react/migration-guide#automatic-upgrade) method, with additional changes by reading the [migration guide](https://github.com/storybookjs/storybook/blob/next/MIGRATION.md#from-version-65x-to-700).
- Updates dependencies since https://github.com/guardian/gateway/pull/2282
- Fixes issue with storybook and typescript 5 by manually installing and setting `react-docgen-typescript-plugin`
- Fixes type issues as shown in `src/email/testUtils.tsx`, didn't spend too long at it, and just set it to `any` as it's just used for email templates in storybook
- `yarn.lock` had to be rebuilt due to issues with storybook which were fixed by removing yarn.lock, node_modules, and reinstalling dependencies, hence the large number of line changes, and making sure that it still works on DEV/CODE envs.
    - From solution here: https://github.com/storybookjs/storybook/issues/22591#issuecomment-1552782423

## Tested
- [x] DEV - Sign In/Registration/Reset Password flows
- [x] CODE - Sign In/Registration/Reset Password flows


.storybook/preview-head.html
```diff
@@ -1,190 +1,5 @@
...
-        format('woff'),
-      url(https://assets.guim.co.uk/static/frontend/fonts/guardian-textsans/noalts-not-hinted/GuardianTextSans-RegularItalic.ttf)
-        format('ttf');
-    font-display: swap;
-  }
-</style>
+<script>
+  // fix for chromatic/storybook failing to render a component
+  // as the window.guardian object was not defined
+  window.guardian = {};
```
raphaelkabo, line 4: I guess we don't need the font declarations and other nonsense any more?

raphaelkabo (approved):
✅  Server and Storybook both run great, and I swear Storybook is faster, too! Here's something for 10% time: run Storybook via `make storybook`. It's the only thing not in our Makefile!

.storybook/preview-head.html
coldlink: Well that's all still there, just under the `GlobalStyles` component now in preview.js

raphaelkabo: Oh, duh. :P

PR #2296, opened 2023-06-23
Send ophan component_event for subscriptions

tomrf1:
A [recent PR](https://github.com/guardian/gateway/pull/2261) enabled one-click subscribes, in addition to unsubscribes.

We'd like to track successful newsletter subscriptions in ophan. This PR changes the `SubscriptionSuccessPage` to send a component_event to ophan.

While working on this I noticed that the ophan model is slightly wrong in this repo - `componentEvent` data should be nested.

Screenshot 2023-06-26 at 08 35 30

PR #2301, opened 2023-06-27
Only track subscribe events

tomrf1:
A [previous PR](https://github.com/guardian/gateway/pull/2296) added component_event tracking for newsletter subscriptions.
It should only track subscribes, not unsubscribes. This PR fixes that.


kelvin-chappell (approved):
👍

PR #2327, opened 2023-07-13
Enable OAuth on production

coldlink:
## What does this change?

This enables the OAuth switch on production.


kelvin-chappell (approved):
🚀

PR #2332, opened 2023-07-24
Increase threshold on 5xx alarm

coldlink:
## What does this change?

The 5xx alarm has been going off quite a lot recently, and it seems like it's due to a downstream issue with braze. For now I'm increasing the threshold to 15 to see if it hopefully helps with the spam.

Screenshot 2023-07-24 at 12 15 35
Screenshot 2023-07-24 at 12 14 05


pvighi (approved):
this would probably help with the alarm noise, I just worries me that we might miss other issues

PR #2345, opened 2023-08-01
Ab test 3 stage sign in flow

lindseydew:
## What does this change?

This adds an AB test that will direct 50% of the audience to a sign in flow that places the marketing consents and newletters opt ins on one page, so that the sign in flow becomes a 3 stage process rather than a 4 stage process. The hypothesis is that more users will complete registration with a shorter process. 

## How to test

Add a cookie `GU_mvt_id` with value `1` to your browser and visit `https://profile.thegulocal.com/consents`. This should redirect you to `https://profile.thegulocal.com/consents/our_content`

## Code tests:

- [x] MVT even, A new social registration is redirected to the 3 step flow
- [x] MVT odd, A new social registration is redirected to the 4 step flow
- [x] MVT even, A new password registration is redirected to the 3 step flow
- [x] MVT odd, A new social registration is redirected to the 4 step flow

## Images
Currently it looks like this:
our_content_with_image

The image used is this one: 
supporter

There is a link to the original design file here: https://www.figma.com/file/QjyXYCypmkedbk9gp0GJph/Global-New-Year-2022?type=design&node-id=983-20386&mode=design&t=OgVqYZ7kyu4QIOSd-0 

Would be great to confirm best practises to use here

## Accessibility


src/shared/model/experiments/tests/abSimplifyRegistrationFlowTest.ts
```diff
@@ -0,0 +1,31 @@
+import { ABTest } from '@guardian/ab-core';
+
+export const abSimplifyRegistrationFlowTest: ABTest = {
+	id: 'SimplifyRegistrationFlowTest', // This ID must match the Server Side AB Test
+	start: '2023-08-01',
+	expiry: '2023-09-30', // Remember that the server side test expiry can be different
+	author: 'Personalisation',
+	description: '',
+	audience: 0.5,
```
vlbee, line 9: I would triple check this.

vlbee:
I would suggest creating a story for the Our_Content page so it's ready to go if this is successful and rolled out.

vlbee:
Have we checked the ab test bucketing and redirection to the correct page flow works as expected for all platforms (web & apps) as well as for new registrations via social? These have a slightly differently flow I recall.

src/shared/__tests__/routeUtils.test.ts
```diff
@@ -15,8 +15,8 @@ describe('routeUtils', () => {
 			expect(output).toEqual('/reset-password/tokenValue');
 		});
 		test('a non-parameterised path requires no object', () => {
-			const output = buildUrl('/consents/communication');
-			expect(output).toEqual('/consents/communication');
```
vlbee, line 19: Why was this changed and not appended? Just checkign we aren't losing any checks.

src/server/routes/consents.ts
```diff
@@ -128,188 +130,296 @@ const getUserNewsletterSubscriptions = async ({
...
+		const isInABTestVariant = this.ab.isUserInVariant(
+			abSimplifyRegistrationFlowTest.id,
+			abSimplifyRegistrationFlowTest.variants[0].id,
+		);
+		return isInABTestVariant;
+	};
+
+	constructor(readonly ab: ABTestAPI) {
+		if (this.inSimplifyRegistrationFlowTest()) {
+			this.pages = [OUR_CONTENT, YOUR_DATA(OUR_CONTENT.page), REVIEW];
```
vlbee, line 406: I think for legibility, it might be easier to extract these two arrays into named variables or a Map so you can just match on the key, and so it's very clear what's going on at a glance. Especially if we ever roll out permanent flow personalistation.

lindseydew: yeah good idea, updated with a comment too

src/shared/model/experiments/tests/abSimplifyRegistrationFlowTest.ts
```diff
@@ -0,0 +1,31 @@
...
+	audienceOffset: 0,
+	successMeasure:
+		'We believe that shortening the onboarding flow by having 3 steps rather than 4 will increase the completion rate of registration by X%',
+	audienceCriteria: '50% of the audience',
+	idealOutcome: 'Increase of completion of the registration flow',
+	showForSensitive: true, // Should this A/B test run on sensitive articles?
+	canRun: () => true, // Check for things like user or page sections
+	variants: [
+		{
+			id: 'variant', // toggle is on
```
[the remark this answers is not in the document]
lindseydew: Ah I see yeah, made the audience 100%

lindseydew:
> I would suggest creating a story for the Our_Content page so it's ready to go if this is successful and rolled out.

yeah that has been added here: 
https://github.com/guardian/gateway/pull/2345/files#diff-82ea2970d9eea5979f4e5408aee64e2729a9c210b982106171eb82105952cce8

src/shared/model/Routes.ts
```diff
@@ -12,6 +17,7 @@ export const ValidRoutePathsArray = [
 	'/agree/GRS',
 	'/consents',
 	'/consents/:page',
+	'/consents/our_content',
```
coldlink, line 20: our general convention is to use `-` instead of `_` in urls, so if possible would be great to refactor this to 
```suggestion
	'/consents/our-content',
```

src/client/pages/ConsentsOurContent.tsx
```diff
@@ -0,0 +1,120 @@
+import React from 'react';
```
coldlink, line 1: Can we get a storybook for this page please!

src/shared/model/Routes.ts
lindseydew: Yes, I can change that 👍

src/client/pages/ConsentsOurContent.tsx
lindseydew: Yes, this has actually turned out to be a bit more complicated than I first thought. It seems adding the check for an AB test here: `const ABTestAPI = useAB();` breaks the story book components. 

I've moved the check to further up the pages here: https://github.com/guardian/gateway/pull/2349 but this does introduce a code change to a lot of different files, so I'm not sure if this is the best approach

coldlink: That's actually the approach I'd have suggested, add the `useAB` check on the `*Page.tsx` file, and the component itself takes a prop to determine if it's in the test or not.

The components should be able to exist without any external api/state etc being needed to be call, anything they need should be passed in as props.

lindseydew: Ok yeah, fair enough. We can take this approach in that case

lindseydew: If we use the `clientState` to figure out the AB test instead then this is what the changes look like: https://github.com/guardian/gateway/pull/2350

It's more convenient from a code point of view, but I noticed it says it's not a recommended approach from here: https://github.com/guardian/gateway/blob/d5ceda4ecbe40aba7b8816d9ef5775b296dc6256/src/client/components/ABTestDemo.tsx#L56C1-L64C26

Would there be any concerns with taking this approach?

src/server/routes/consents.ts
```diff
@@ -127,198 +129,317 @@ const getUserNewsletterSubscriptions = async ({
...
+	constructor(readonly ab: ABTestAPI) {
+		/***
+		  If user is in the AB test bucket, show them the 3 step flow,
+		  otherwise show the original 4 step flow
+		***/
+		if (this.inSimplifyRegistrationFlowTest()) {
+			this.pages = this.threeStageAbTestPages;
+		} else this.pages = this.originalPages;
+	}
+}
```
coldlink, line 421: a class seems a bit overkill for this functionality, and is hard to reason about, would it be easier to turn this into a function instead which returns the pages?

```suggestion
export const getConsentPages = (abTestApi: ABTestAPI): SomePagesType => {
  const isInABTestVariant = abTestApi.isUserInVariant(
			abSimplifyRegistrationFlowTest.id,
			abSimplifyRegistrationFlowTest.variants[0].id,
	);
	
	if (isInABTestVariant) {
	  return [OUR_CONTENT, YOUR_DATA(OUR_CONTENT.page), REVIEW]
	}
	return [
		COMMUNICATION,
		NEWSLETTERS,
		YOUR_DATA(NEWSLETTERS.page),
		REVIEW,
	]
}
```

or something along those lines?

cypress/integration/mocked/onboarding_flow.1.cy.ts
```diff
@@ -297,6 +298,219 @@ describe('Onboarding flow', () => {
 				.should('have.attr', 'href')
 				.and('include', decodeURIComponent(returnUrl));
 		});
+		//@AB_TEST: 3 Stage Registration Flow Test START
+		describe('AB Test 3 stage registration flow', () => {
```
coldlink, line 302: A mocked test is great, but we generally try to use end to end tests now, so an `ete-okta` test will be better as that will test real APIs, you can use `setMvtId` within the end to end tests too. The mocked test is still helpful though.

src/server/routes/consents.ts
lindseydew: Yeah that's nicer, thanks

src/client/components/ConsentsSubHeader.stories.tsx
```diff
@@ -39,7 +39,6 @@ export const Password = () => (
 	/>
 );
 Password.storyName = 'With Password as current';
-
```
vlbee, line 42: extra modified file for no reaseon

coldlink (approved):
looks great to me!

only suggestion would be to clean up the commit history slightly we've been trying to keep a relatively easy to read commit history for our future selves to give us thanks

lindseydew:
> looks great to me!
> 
> only suggestion would be to clean up the commit history slightly, we've been trying to keep a relatively easy to read commit history for our future selves to give us thanks

Sure thing, I can squash and merge if you like?

coldlink:
> > looks great to me!
> > only suggestion would be to clean up the commit history slightly, we've been trying to keep a relatively easy to read commit history for our future selves to give us thanks
> 
> Sure thing, I can squash and merge if you like?

Yeah that works!

PR #2360, opened 2023-08-14
Dependencies | August 14th 2023

coldlink:
## What does this change?

Changes from previous https://github.com/guardian/gateway/pull/2347

- Major bump of `@testing-library/jest-dom` which required minor test changes

Holds back the following:

## Tested

- [x] CODE
- [x] IE11


raphaelkabo (approved):
👍

PR #2361, opened 2023-08-15
Replace Opinion with Soccer newsletter for US

vlbee:
## What does this change?

Removes Opinion newsletter and replaces with Soccer newsletter for US newsletter onboarding flow. 

Please not the soccur newsletter with ID 6030 does not exist in CODE if testing in code, so will not render on the page. 

Code newsletters are not being updated, so I will raise a separate PR to use PROD idapi endpoint for reading all newsletter information, when otherwise pointing to CODE. This means some newsletter updates in CODE gateway might not actually persist because in the rest of CODE that newsletter ID does not yet exist. 

## Images

Soccer newsletter title/description not uploaded to CODE environment, hence the weird text as I've used a random test id just to get it to render for the image. 
US image title was very wide, so could not get it to fit so had to add some css override code to add background filler top and bottom. 

image

image

## Accessibility

-   [ ] [Tested with screen reader](https://github.com/guardian/accessibility/blob/main/people-and-technology/03-visual.md#screen-reader)
-   [ ] [Navigable with keyboard](https://github.com/guardian/accessibility/blob/main/people-and-technology/02-physical.md#keyboard)
-   [ ] [Colour contrast passed](https://github.com/guardian/accessibility/blob/main/people-and-technology/03-visual.md#contrast)
-   [ ] [The change doesn't use only colour to convey meaning](https://github.com/guardian/accessibility/blob/main/people-and-technology/03-visual.md#use-of-colour)

PR #2362, opened 2023-08-21
WIP: Update webpack image loader to be able to support .avif

vlbee:
## What does this change?
Branches off #2361 which will be merged first (newsletter image update)

Webpack Impage Loader plugin does not support .avif and has been [deprecated](https://github.com/tcoopman/image-webpack-loader#readme) in favour of [Image-Minimizer-Webpack-Plugin](https://github.com/webpack-contrib/image-minimizer-webpack-plugin/)

.avif has been [adopted by all browsers except Edge/IE](https://caniuse.com/?search=avif) and is already being serverd on dotcom pages eg. newsletter sign up pages.  Images can be as small as 16kb will still excellent quality. 

The new plug in can use one of 4 tools to optimize/generate images - I've chosen [Sharp](https://github.com/lovell/sharp)

TODO
- test nothing breaks
- test optimum sharp settings for various media types
- add newsletter/image readme

## Why?

It's often tricky to get consistently sized files (resolution, quality, filetype) from Design. I'd like to add automatic image resizing/converstion to avif where possible so perfomance is not degraded. 

TODO - benchmark FCP load times for newsletter page pre and post update. Tbc how best to do this for pages requiring sign in.


coldlink:
Updated webpack to use the new image loader in https://github.com/guardian/gateway/pull/2434

PR #2378, opened 2023-08-24
Use `customer.help@theguardian.com` as contact email

coldlink:
## What does this change?

The user help team has asked us to replace all instances of their old email with `customer.help@theguardian.com`. This PR does a find and replace to change those, and fixes a mistake where we used `@guardian.com` instead of `@theguardian.com`


raphaelkabo (approved):
yep.

PR #2391, opened 2023-09-04
Dependencies | September 4th 2023

coldlink:
## What does this change?

Changes from previous https://github.com/guardian/gateway/pull/2390

- Updates Github Actions where possible

Holds back the following:

## Tested

- [x] CODE


kelvin-chappell (approved):
👍

PR #2405, opened 2023-09-12
Group dependabot updates

kelvin-chappell:
As we tend to bump all the dependencies in one go once a week we might as well group them together in a single PR.

There might be some logical subgroups that we want to use instead but this seemed like a sensible first iteration.

See https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#groups


kelvin-chappell:
Closing this for now as the dependency grouping needs a bit more thought.

PR #2406, opened 2023-09-18
Dependencies | September 18th 2023

coldlink:
## What does this change?

Changes from previous https://github.com/guardian/gateway/pull/2391

Holds back the following:

* `@aws/*` packages have some slight typescript type issues which haven't been resolved yet, will try upgrading next week
```
Type 'NodeHttpHandler' is not assignable to type 'RequestHandler<any, any, HttpHandlerOptions> & RequestHandler<HttpRequest, HttpResponse, HttpHandlerOptions> & { ...; }'.
Type 'NodeHttpHandler' is missing the following properties from type '{ updateHttpClientConfig(key: never, value: never): void; httpHandlerConfigs(): {}; }': updateHttpClientConfig, httpHandlerConfigs
``` 

## Tested

- [x] CODE


raphaelkabo (approved):
🚀

PR #2422, opened 2023-09-27
Okta OIE Prep | Migrate sessions to use the `/me` endpoint

coldlink:
## What does this change?

tl;dr This PR fixes a blocker related to sessions within OIE. Safe to merge to CODE and PROD

Firstly, the use of the session ID cookie (`sid`) isn't supported in Identity Engine. The new `idx` cookie is used with Identity Engine.

Next, the Okta `/v1/sessions/{sessionid}` API has been deprecated, and doesn't support the `idx` cookie, only the only `sid` cookie. Okta recommends that we move away from using this API and use the [My Session Management endpoints](https://developer.okta.com/docs/api/openapi/okta-management/management/tag/Session/#tag/Session/operation/getCurrentSession) instead (`/v1/sessions/me` API endpoints).

All the related changes to the way sessions work in Okta after the upgrade are documented here: https://developer.okta.com/docs/guides/oie-upgrade-sessions-api/main/

From testing the behaviour in Okta dev, the following points were determined:
- `sid` cookie was used to create an `idx` cookie when going through oauth flow/sign in agin
- `sid` cookie is still set by okta on sign in/oauth, with expiry time set to when the session expires
- `idx` cookie also set by okta on sign in/oauth, but is only set as a session cookie, need to investigate further the implications of this
- The `/v1/sessions/me` API authenticates with a cookie, and not an Okta API token. It can be authenticated with the `sid` cookie, or the `idx` cookie, or both cookies provided at the same time
- This points to the outcome that the `sid` cookie should work alongside the `idx` cookie, i.e dual-running, and our code should be prepared so that we can use either one interchangeably or together, even post OIE migration

This PR replaces the `/v1/sessions/:sessionId` API methods (`getSession` and `closeSession`) with the `/v1/sessions/me` API (`getCurrentSession` and `closeCurrentSession`) which authenticates with either/both of the `sid`/`idx` cookies instead of the API token which the old endpoint was doing. Responses between the APIs are identical. Thankfully as we're running Okta and Gateway on the same domain, it makes reading cookies a breeze, so we just read the cookies from the request and forward them onto Okta. We also update Cypress to use these new endpoints too.

This is safe to deploy to CODE and PROD, as the `/v1/sessions/me` endpoint is fully compatible with non-OIE tenants, which we're currently using in CODE and PROD.

The Cypress tests currently don't use the `idx` cookie, as we usually run Cypress against the CODE environment, so when we do the Okta OIE upgrade in CODE, we can update the cypress tests once upgraded on CODE!

## Tested
- [x] DEV - Okta Identity Engine
- [x] CODE - Okta Classic

### DIFF
--- a/cypress/integration/ete-okta/reauthenticate.4.cy.ts
+++ b/cypress/integration/ete-okta/reauthenticate.4.cy.ts
@@ -39,7 +39,7 @@ describe('Reauthenticate flow, Okta enabled', () => {
 					const sid = sidCookie?.value;
 					expect(sid).to.exist;
 					if (sid) {
-						cy.getCurrentOktaSession(sid).then((session) => {
+						cy.getCurrentOktaSession({ sid }).then((session) => {
 							expect(session.login).to.equal(emailAddress);
 						});
 					}
@@ -87,7 +87,7 @@ describe('Reauthenticate flow, Okta enabled', () => {
 									const sid = sidCookie?.value;
 									expect(sid).to.exist;
 									if (sid) {
-										cy.getCurrentOktaSession(sid).then((session) => {
+										cy.getCurrentOktaSession({ sid }).then((session) => {
 											expect(session.login).to.equal(emailAddressB);
 										});
 									}
--- a/src/server/lib/okta/api/sessions.ts
+++ b/src/server/lib/okta/api/sessions.ts
@@ -4,14 +4,17 @@ import { buildUrl } from '@/shared/lib/routeUtils';
 import { joinUrl } from '@guardian/libs';
 import { getConfiguration } from '../../getConfiguration';
 import { handleErrorResponse } from './errors';
-import { defaultHeaders, authorizationHeader } from './headers';
+import { defaultHeaders } from './headers';
 
 const { okta } = getConfiguration();
 
 /**
- * Get a User session by session Id
+ * Get a User session by session cookie
  *
- * https://developer.okta.com/docs/reference/api/sessions/#get-session
+ * For Identity Classic the cookie is called `sid`
+ * For Identity Engine the cookie is called `idx`
+ *
+ * https://developer.okta.com/docs/api/openapi/okta-management/management/tag/Session/#tag/Session/operation/getCurrentSession
  *
  * The Okta API returns the users session details on success or
  * returns a 404 on error.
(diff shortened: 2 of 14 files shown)


src/server/lib/okta/api/sessions.ts
```diff
@@ -22,31 +25,65 @@ const { okta } = getConfiguration();
...
+		headers: { ...defaultHeaders, Cookie },
+		credentials: 'include',
+	});
+
+	if (response.ok) {
+		try {
+			return sessionSchema.parse(await response.json());
+		} catch (error) {
+			throw new OktaError({
+				message: 'Could not parse Okta session response',
```
kelvin-chappell, line 49: Would it be worth adding the response to the error message to help understand what went wrong?

cypress/integration/ete-okta/reauthenticate.4.cy.ts
```diff
@@ -39,7 +39,7 @@ describe('Reauthenticate flow, Okta enabled', () => {
 					const sid = sidCookie?.value;
 					expect(sid).to.exist;
 					if (sid) {
-						cy.getCurrentOktaSession(sid).then((session) => {
+						cy.getCurrentOktaSession({ sid }).then((session) => {
```
kelvin-chappell, line 42: What's the effect of the brackets?

src/server/lib/okta/api/sessions.ts
coldlink: this error is only caught when the [zod](https://github.com/colinhacks/zod) [`.parse`](https://github.com/colinhacks/zod?tab=readme-ov-file#parse) method fails, where it just throws a generic error. Or if the [`response.json()`](https://developer.mozilla.org/en-US/docs/Web/API/Response/json_static#exceptions) throws an error, again if it fails to parse.

So we don't actually need to specifically handle the `error` variable itself.

cypress/integration/ete-okta/reauthenticate.4.cy.ts
coldlink: I changed the first parameter of the `getCurrentOktaSession` method to now be an object instead of a value, so effectively i'm just turning the parameter i'm passing into `getCurrentOktaSession` an object with the following shape: 
```js
{
  sid: sid
}
```

This makes it easy to add or change parameters in the future, e.g. adding the `idx` cookie would become 
`.getCurrentOktaSession({ sid, idx })` or just `.getCurrentOktaSession({ idx }) instead of having to define multiple parameters e.g. `getCurrentOktaSession(undefined, idx)`

kelvin-chappell (approved):
All very clear and it's good to know that we should be able to transition from `sid` to `idx` seamlessly 👍

PR #2424, opened 2023-10-03
Okta OIE Prep | Update Sign In Page Interception Code

coldlink:
## What does this change?

The Okta Identity Upgrade dashboard gives a warning that "Customization detected in Okta Hosted Sign-In Widget. Review code for deprecated CSS/Javascript that may impact sign-in". This is because we use custom code in the Okta Hosted Sign In Page to intercept calls to that page and redirect to our own sign in page in Gateway instead.

This PR is that it modifies the interception code to work with both Identity Classic and Identity Engine. Specifically handling the change from `OktaUtil.getRequestContext().target.clientId` to `OktaUtil.getRequestContext().app.value.id`  as suggested in this [article](https://support.okta.com/help/s/article/Hosted-SIW-Function-Change-OktaUtilgetRequestContexttargetclientId-for-switch-statement-on-application-context?language=en_US).

This was the only change required, and tests have been added to mock this behaviour.

I also went through this document about [Deprecated JavaScript methods in the widget](https://developer.okta.com/docs/guides/oie-upgrade-sign-in-widget-deprecated-methods/main/), but nothing in there affected any of our code.

## Tested
- [x] DEV - Identity Engine - Logged into the Okta sample application
- [x] CODE - Identity Classic - Logged into the Okta sample application


kelvin-chappell (approved):
All looks good to me 👍

PR #2427, opened 2023-10-06
OAuth | Correctly assert `clientId` when verifying tokens

coldlink:
## What does this change?

The Okta JWT verifier by default only verifies if a given access token is valid and issued by a given authorisation server rather than checking the client id of the application it was issued for.

The client id is checked already for ID tokens by passing the `expectedClientId` parameter as the 2nd argument to `.verifyIDToken`.

This could potentially allow someone to use an access token from another application from the same authorisation server within Gateway if they were to change the `GU_ACCESS_TOKEN` cookie value.

To prevent this in Gateway we want to make sure the access tokens that Gateway is using and storing have the client id that belongs to gateway for both the ID token and the Access Token.

To do this, Okta suggest using a ["Custom Claims Assertation"](https://github.com/okta/okta-jwt-verifier-js/blob/master/README.md#custom-claims-assertions) when setting up the `OktaJwtVerifier` object e.g.

```js
const verifier = new OktaJwtVerifier({
  issuer: 'https://{yourOktaDomain}/oauth2/default',
  clientId: '{clientId}'
  assertClaims: {
    cid: '{clientId}'
  }
});
```

This works as expected for the access token which is what we want, however for the ID token it also tries to run the `assertClaims` and the `cid` claim doesn't exist in the ID token (the client id is under the `aud` claim). This means we can't use the same `OktaJwtVerifier` object to verify both Access tokens and ID tokens.

So in this PR I set up two objects individually for each token type so we can successfully verify the client id with each!

The `oauthAccessTokenVerifier` for verifying access tokens, which includes the `assertClaims`, and the `oauthIDTokenVerifier` for verifying ID tokens, which doesn't include the `assertClaims`.

I also make sure that the correct function can only be used with either using the typescript `Pick` type, so that `verifyAccessToken` can only be used with `oauthAccessTokenVerifier`, and `verifyIdToken` can only be used with `oauthIDTokenVerifier`

### DIFF
--- a/src/server/lib/okta/tokens.ts
+++ b/src/server/lib/okta/tokens.ts
@@ -46,17 +46,55 @@ const issuerDomain = () => {
 };
 
 /**
- * @name oauthTokenVerifier
- * @type {OktaJwtVerifier}
- * @description Okta JWT Verifier
+ * @name verifierSharedOptions
+ * @type {OktaJwtVerifier.VerifierOptions}
+ * @description Okta JWT Verifier - Shared Options between the access token and ID token verifiers
  */
-const oauthTokenVerifier = new OktaJwtVerifier({
+const verifierSharedOptions: OktaJwtVerifier.VerifierOptions = {
 	issuer: joinUrl(issuerDomain(), '/oauth2/', okta.authServerId),
-});
+	clientId: okta.clientId,
+};
+
+/**
+ * @name oauthAccessTokenVerifier
+ * @type {OktaJwtVerifier}
+ * @description Okta JWT Verifier - Access Token
+ *
+ * This verifier is used to verify the access token, as Okta JWT Verifier doesn't verify
+ * the clientId claim access, we need to assert it manually for the access token (`cid` claim) which is different
+ * from the ID token (`aud` claim) where Okta verifies it through the `expectedClientId` parameter in `verifyIdToken`.
+ *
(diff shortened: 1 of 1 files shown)


src/server/lib/okta/tokens.ts
```diff
@@ -46,17 +46,47 @@ const issuerDomain = () => {
...
+ * we send it to other APIs.
+ */
+const oauthAccessTokenVerifier: Pick<OktaJwtVerifier, 'verifyAccessToken'> =
+	new OktaJwtVerifier({
+		issuer: joinUrl(issuerDomain(), '/oauth2/', okta.authServerId),
+		clientId: okta.clientId,
+		assertClaims: {
+			cid: okta.clientId,
+		},
+	});
```
kelvin-chappell, line 67: To avoid repetition in both verifiers, could this be implemented by a common method that then does the `assertClaim` in `oauthAccessTokenVerifier` only?

coldlink: I tried having a look, but this doesn't seem possible to have a common method for both.

However to avoid config repetion, I have separate the `issuer` and `clientId` into its own shared options object, and use that as the base for the verifier instantiation

kelvin-chappell (approved):
Good idea! That's the benefit of the dev standup 👍

PR #2434, opened 2023-10-16
Dependencies | October 16th 2023

coldlink:
## What does this change?

Changes from previous https://github.com/guardian/gateway/pull/2428

- Skips cypress test due to OIE upgrade, will re-enable once fixed on CODE
- Replace [`image-webpack-loader`](https://github.com/tcoopman/image-webpack-loader) with [`image-minimizer-webpack-plugin`](https://github.com/webpack-contrib/image-minimizer-webpack-plugin) as the former has been deprecated, doesn't support `avif`, and fixes a [downstream security issue](https://github.com/guardian/gateway/security/dependabot/50)
  - Based on what @vlbee had started in https://github.com/guardian/gateway/pull/2362 

Holds back the following:

## Tested

- [x] CODE


kelvin-chappell (approved):
👍

PR #2444, opened 2023-10-17
Handle maxAge parameter

raphaelkabo:
## What does this change?

When OAuth2 authorization code flow requests come to Gateway, we can send a [`max_age` parameter](https://developer.okta.com/docs/reference/api/oidc/#request-parameters), which is the "allowable elapsed time, in seconds, since the last time the end user was actively authenticated by Okta". This allows us to ensure that an MMA access token only has a lifetime of 30 minutes.

Usually, when Gateway receives an authentication request from Okta, and an Okta session is already active, Gateway will show the 'signed in as' page. For requests coming from MMA, we _don't_ want this page to show. Instead, we always want the user to end up on the `/reauthenticate` page, and perform full auth via password or social.

This PR adds functionality to extract and parse the `max_age` property during the OAuth flow, and pass it on to the rest of Gateway as a `maxAge` query parameter. In the `redirectIfLoggedIn` middleware, we now check for the existence of the `maxAge` parameter. If present, this means that the Okta session is valid but it's older than `maxAge` - in this case, we simply redirect the user to `/reauthenticate`. From here, the user can only receive fresh OAuth tokens by performing full auth.


kelvin-chappell (approved):
All makes sense and code looks fine to me 👍

PR #2450, opened 2023-10-20
Okta Identity Engine |  Switch from `sid` cookie to `idx` cookie in tests and documentation

coldlink:
## What does this change?

As it says in the title, this PR switches from the `sid` cookie to the `idx` cookie in tests and documentation, in some cases we still keep both where the `sid` and `idx` cookie might be set together, e.g. on Sign Out tests, and tests where we need to check if certain cookies exist or not.


raphaelkabo (approved):
`s/sid/idx/g` 👍

PR #2210, opened 2023-04-05
Update documentation for Tokens, Session, and Native and Web App Integration

coldlink:
## What does this change?

Updates our native apps integration guide in order to be better up to date with how to go about implementing sign in, registration, and reset/set password for native iOS and Android apps.

Adds documentation on how to integrate web applications with Okta, based on the preliminary research done for the spike. This document should be finalised once the outcomes from the spike have been decided.

Also adds information on the difference between all the OAuth tokens, and adds documentation on the different meaning of the word "sessions" when it comes to Okta.

PR #2451, opened 2023-10-20
Okta |  Remove "fallback" option for apps

coldlink:
## What does this change?

The fallback option was considered as part of the initial Okta integration with Native Apps in order to provide an option for apps to still authenticate even if the Okta Hosted Sign In page interception code was to fail.

However this was never implemented in apps as we were unable to get this to work, a lack of resources, the fact that the interception code proved remarkably stable, and now that we've removed SSO option on iOS this will not be possible to implement anyway. We'd already marked the option as deprecated a while back: https://github.com/guardian/gateway/pull/2210

So this PR removes all the related fallback code, from the `force_fallback` testing implementation on the interception code, to the apple site association file, and the OAuth callback route where the fallback was implemented, along with any related documentation to avoid confusion.


raphaelkabo (approved):
Glad to be removing this because I never fully understood it! 👀

PR #2453, opened 2023-10-26
Upgrade to Node 20 LTS

coldlink:
## What does this change?

Node 20 is now in [LTS](https://nodejs.org/en/blog/release/v20.9.0), upgrade to it from Node 18. The only thing we had to change in gateway was the `.nvmrc` file. GitHub actions will use this file in order to determine what version of Node to install, as will your computer if you have `fnm` or `nvm` installed.

## Notes

The AMI also has to be updated to Node 20.9.0 too (latest LTS version at time of writing).

## Tested
- [x] CODE


kelvin-chappell (approved):
🔍 👍

PR #2461, opened 2023-11-01
Signed in as Page | Handle `login_required` error with helpful messaging

coldlink:
## What does this change?

Users would sometimes click the "Continue" on the "Signed in as" page, and they would be redirected back to the same page because of an error, specifically when there was a `login_required` error if there was an issue with the users session

However this error would sometimes not propagate to the "Signed in as" page. This PR makes sure it does, and adds in helpful messaging to unblock the user, namely by Signing Out and back in again, clearing cache/cookies, or contacting customer support.

Screenshot 2023-11-01 at 10 06 28


raphaelkabo (approved):
✅

PR #2462, opened 2023-11-02
Sign out from current browser only

coldlink:
## What does this change?

We are currently signing users out from *ALL* browsers/devices/apps when the user clicks "Sign Out" on the web, or navigates to the sign out page directly (https://profile.theguardian.com/signout), and we clear all OAuth sessions too.

This was done to replicate the behaviour as it occurred in the IDAPI only system. However this doesn't provide a great user experience, as users would often get confused why they would get signed out on the app seemingly randomly.

This PR updates this behaviour so we only sign out on the current device/browser.

This is done by no longer calling the IDAPI `/unauth` endpoint which would invalidate all IDAPI user sessions, and only clears the IDAPI cookies now.

And for Okta we no longer use the "Clear user sessions" endpoint, which like IDAPI would close all user sessions for a given user, and revoke all OAuth tokens, we now use the "Close Current User Session" endpoint, and clear the Okta cookies, but not revoke the OAuth tokens. OAuth tokens will automatically be deleted when an application see's our `GU_SO` cookie which tells the application that the user has signed out recently and should clear any tokens.

https://github.com/guardian/gateway/blob/229281c273e82f25c17cd256bc2fa2c7bdb310ca/docs/okta/web-apps-integration-guide.md?plain=1#L182

Ideally in the future we'd also want to implement Single Logout (SLO) which during sign out will call any applications individually and revoke the tokens for that application rather than relying on the GU_SO cookie.


docs/okta/signout.md
```diff
@@ -4,23 +4,21 @@ This document describes how we've implemented the sign out flow with Okta in Gat
...
 
-This endpoint requires the okta `userId`, which we get from the okta sessions api using the current okta session id cookie `idx`
-
-We then use this user id to clear all Okta sessions for that user.
+This endpoint requires the okta session cookie, which we use the current okta session id cookie `idx`
 
 In the scenario where the okta call to remove the session fails (and we have a orphaned session), we delete the `idx` cookie anyway, and we plan in the future to allow the users to see their active sessions in manage my account section of the site where they can remove them manually.
 
-If there is no existing session then we'll render the sign in page.
+If there is no existing session then we'll just redirect to the `returnUrl` and clear any cookies should they exisy.
```
raphaelkabo, line 17:
```suggestion
If there is no existing session then we'll just redirect to the `returnUrl` and clear any cookies should they exist.
```

raphaelkabo (changes requested):
This looks great! Nice and simple fix. ✅

raphaelkabo (approved):

PR #2463, opened 2023-11-02
Okta | Remove `sid` cookie and only rely on the `idx` cookie

coldlink:
## What does this change?

Remove the Okta `sid` cookie, which was the legacy session cookie in Okta Identity Classic, and only rely on the new Okta Identity Engine `idx` cookie to determine if a session is available for the user.

In cypress nginx config, no longer send the `sid` cookies in requests to the application, thus making sure we and Okta APIs only rely on the `idx` cookie.

# Why?

When we upgraded to Okta Identity Engine on PROD, we have an issue with existing Okta Identity Classic sessions not being transferred to the new Okta Identity Engine sessions, for as yet unknown reasons. While we're in contact with Okta support, we're prepping for the situation where we want to sign users out gradually and get them to log in again, thus getting a new session in Identity Engine.

The other reason is that now that we're in Identity Engine, we only really want to rely on a single session identifier, and avoid confusion with having 2 session cookies, so only rely on the `idx` cookie going forward. 

In both scenarios, users who only have a `sid` cookie will silently be logged out over a period of 90 days (our max session length) as we no longer refresh sessions for those users. Therefore overtime users will have to log in again in order to get a session set. Once they've logged in again, session refreshing will work as expected as the `idx` cookie will be used to refresh the session.

# Tested
- [x] CODE

- The `sid` cookie is ignored if it's already set, so behaves as if no `sid` cookie
- Going to `/signin/refresh` if you have a `sid` cookie and identity cookies does nothing (i.e doesn't set new cookies), so the user will eventually get logged out
- If `idx` cookie set, then refresh works as expected, `idx` cookie is set, `sid` cookie not set, identity cookies are refreshed

# Related PR

- Removing `sid` cookie from Fastly and Nginx configuration: https://github.com/guardian/identity-platform/pull/688
  - This has to go in at the same time as this PR

### DIFF
--- a/scripts/nginx.conf
+++ b/scripts/nginx.conf
@@ -58,6 +58,20 @@ http {
         proxy_next_upstream error timeout http_404 non_idempotent;
         proxy_set_header "X-Forwarded-Proto" "https";
         proxy_set_header "X-GU-Okta-Env" "profile.code.dev-theguardian.com";
+        ######
+        # remove `sid` cookie in requests to Gateway
+        # save original "Cookie" header value
+        set $altered_cookie $http_cookie;
+        # check if the "sid" cookie is present
+        if ($http_cookie ~ '(.*)(^|;\s)sid=("[^"]*"|[^\s]*[^;]?)(\2|$|;$)(?:;\s)?(.*)') {
+          # cut "sid" cookie from the string
+          set $altered_cookie $1$4$5;
+        }
+        # hide original "Cookie" header
+        proxy_hide_header Cookie;
+        # set "Cookie" header to the new value
+        proxy_set_header  Cookie $altered_cookie;
+        ######
         # handle nginx backend errors by proxying to gateway
         error_page 502 = @fallback;
       }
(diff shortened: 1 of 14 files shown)


scripts/nginx.conf
```diff
@@ -58,6 +58,20 @@ http {
         proxy_next_upstream error timeout http_404 non_idempotent;
         proxy_set_header "X-Forwarded-Proto" "https";
         proxy_set_header "X-GU-Okta-Env" "profile.code.dev-theguardian.com";
+        ######
+        # remove `sid` cookie in requests to Gateway
+        # save original "Cookie" header value
+        set $altered_cookie $http_cookie;
+        # check if the "sid" cookie is present
+        if ($http_cookie ~ '(.*)(^|;\s)sid=("[^"]*"|[^\s]*[^;]?)(\2|$|;$)(?:;\s)?(.*)') {
```
kelvin-chappell, line 66: This regex is making my head spin.  I had a play with it and it seems to do the right thing.  Is it a well-used pattern?

kelvin-chappell (approved):
Looks good. 👍 

Had a question about the regex.

And I wonder if future people are going to get confused by 'Okta session' cookies and 'session' cookies in general.  But I suppose if the new cookie is a session cookie it isn't going to matter.

kelvin-chappell (approved):

PR #2472, opened 2023-11-06
Add `GET /signout/all` endpoint

coldlink:
In https://github.com/guardian/gateway/pull/2462 we updated sign out so that we only sign out on the current device/browser, instead of AL browsers/devices/apps when the user clicks "Sign Out" on the web, or navigates to the sign out link directly.

This provides better user experience where if they are logged into multiple devices or apps they will no longer get logged out of those devices when they sign out on the web on another device.

However a case by user help was made that they used the old way of signing out in order to reset a user's session.

So this PR adds a new endpoint built specifically for this use case:

`GET /signout/all` which works in the same way that `GET /signout` did pre https://github.com/guardian/gateway/pull/2462 by revert cherrypicking some of the code that was removed.

This way most users who sign out will get the expected behaviour of only signing out on the current device/browser when they click on a link to sign out rather than all devices, and if required user help can send users to the new endpoint in order to sign users out from all devices/browsers if required!


kelvin-chappell (approved):
Looks good to me! 👍

PR #2486, opened 2023-11-23
Update Aus newsletters The Crunch

vlbee:
## What does this change?
For Australia: 
Replaces "Saved For Later" newsletter with "The Crunch" (id 6034')

image

Launches on Thursday. 

Newsletter data not updated fully in code

## Accessibility


dblatcher (commented):
This looks right to me - I think you could get around the "newsletter not available on CODE" issue by taking the list of newsletters directly from the newsletters API rather than IdApi newsletters endpoint (which effectively re-publishes the same data). The read-only endpoint on the newsletters API  is public so the PROD version could be accessed by CODE/local gateway.

Do let me know if you'd like any help trying that out - I feel it should be possible to cut down the amount of code you need to change to switch the newsletters - maybe make it fully configurable.

vlbee:
> This looks right to me - I think you could get around the "newsletter not available on CODE" issue by taking the list of newsletters directly from the newsletters API rather than IdApi newsletters endpoint (which effectively re-publishes the same data). The read-only endpoint on the newsletters API is public so the PROD version could be accessed by CODE/local gateway.
> 
> Do let me know if you'd like any help trying that out - I feel it should be possible to cut down the amount of code you need to change to switch the newsletters - maybe make it fully configurable.

great idea - could you point me the newsletter api docs?

nazeerahmakda:
Hi Vanessa, yes, this should be launched on Thursday!

src/client/assets/newsletters/au-the-crunch.jpg
coldlink, line 1: If the saved for later image is no longer being used, can it be deleted from the assets folder?

dblatcher:
> > This looks right to me - I think you could get around the "newsletter not available on CODE" issue by taking the list of newsletters directly from the newsletters API rather than IdApi newsletters endpoint (which effectively re-publishes the same data). The read-only endpoint on the newsletters API is public so the PROD version could be accessed by CODE/local gateway.
> > Do let me know if you'd like any help trying that out - I feel it should be possible to cut down the amount of code you need to change to switch the newsletters - maybe make it fully configurable.
> 
> great idea - could you point me the newsletter api docs?

Afraid there isn't much documentation for the api currently - the project that publishes it is https://github.com/guardian/newsletters-nx. I'd be happy to help getting it to work on gateway - it woudl be a good test case for building out an implementation guide in the newsletters-nx repo

PR #2488, opened 2023-11-23
Okta | Add `page` parameter to interception code

coldlink:
## What does this change?

The personalisation team are starting work on a sign in teaser prompt for the app. The designs for the prompt have two separate buttons one for sign-in and another for registration. 

Currently in the app, all sign-in/registration routes land the user on the sign-in tab. 

This PR adds a `page` parameter to the Okta sign in page interception code, so the app can ask us which page to show.

We validate this against a list using a `switch` statement.

Currently the following are valid

- `register` - goes to the `/register` page
- `google` - goes to `/signin/google` which instantly starts google login/registration flow
- `apple` - goes to `/signin/apple` which instantly starts apple login/registration flow

If the `page` parameter is invalid, we redirect to the `signin` page as default behaviour.


scripts/okta/lib/helper.ts
```diff
@@ -288,6 +291,23 @@ export const getRedirectUrl = (
...
+			// show a specific page on gateway
+			case 'register':
+				return `/${page}?${params.toString()}`;
+			// instantly start google or apple authentication through gateway
+			case 'google':
+			case 'apple':
+				return `/signin/${page}?${params.toString()}`;
+			default:
+				// if the page parameter is not one of the allowed pages, then break
+				break;
```
kelvin-chappell, line 307: Perhaps default should be the default path and then get rid of L312:
```suggestion
				return `/signin?${params.toString()}`;;
```

kelvin-chappell (approved):
Looks good! 👍

PR #2489, opened 2023-11-27
Dependencies | November 27th 2023

coldlink:
## What does this change?

Changes from previous https://github.com/guardian/gateway/pull/2483

- Brings `@guardian/ab-react` code into project rather than dependency as for a long time (see every last dependency PR) couldn't upgrade as when bundling it caused a broken vendors bundle for unknown reasons.
  - `@guardian/ab-react` code: https://github.com/guardian/csnx/blob/main/libs/%40guardian/ab-react/src/context.tsx

Holds back the following:

## Tested

- [x] CODE


kelvin-chappell (approved):
👍

PR #2491, opened 2023-11-30
Fix Apps Email Registration by combining the Okta activation/recovery token with the encrypted registration consents

coldlink:
### Email Registration Issue

Since https://github.com/guardian/gateway/pull/2455 went live on November 23rd 2023, this had the side effect of breaking email registration via the app flow.

This was because of the way that apps were handling the registration link within the apps.

This works by apps looking out for a specific link/url that begins with `https://profile.theguardian.com/welcome/al_` for Android and `https://profile.theguardian.com/welcome/il_` for iOS. The app would intercept that link, and attempt to extract the token from the end of the URL. It would then start an OAuth flow, and send us the token at the end of the URL. We'd intercept that token and open the registration, reset password, or set password page accordingly.

More information about this can be seen in our documentation, specifically in the [`Registration / Reset Password / Set Password` section of the `Native Apps Integration Guide`](https://github.com/guardian/gateway/blob/main/docs/okta/native-apps-integration-guide.md#registration--reset-password--set-password)

However the way apps were getting the token from the end of the url was rudimentary. It was just looking for the last slash (`/`) and taking anything after that slash to be the token. This worked fine until we introduced the soft opt in flow in the PR mentioned at the start, were we introduced another path after the token which contained the users registration consent state in an encrypted format.

So when the apps intercepted the URL, it would instead send us the incorrect token, namely the encrypted registration consent, as that was now the last part of the url, rather than the actual activation/recovery token.

This PR refactors this behaviour entirely to go back and using a single path. We now encrypt the Okta token before we send it to the user, and as a part of this process we also attach the encrypted registration consents to the token should it exist. This means we only send the user one string which contains the app prefix, Okta token, and registration consent state. This way everything remains as a single path, and so the apps can happily intercept this request, and send us the encrypted string which we can decode back into the Okta token and encrypted registration consent (should it exist).

### Redirect loop issue

Another issue that has been going on longer than the previous issue is the "redirect loop" issue, where if an `activation_token` or `reset_password_token` or `set_password_token` was passed to the OAuth SDK as an additional parameter, the in app browser would continuously get stuck in a redirect loop.

This was identified as the app intercepting the redirect link, and constantly attempting to open a new in app browser with the redirect link, hence the loop.

This behaviour previously used to work correctly. Where `window.location.replace` would replace the current location with a new one, and not cause a system level redirect. But these seems to have changed in probably recent changes to Android (at least).

This was fixed by removing the app token prefix (`al_` for android, and `il_` for ios) from the token in the Okta interception code before redirecting to the welcome/reset password/set password page depending on the given token. 

## Tested
- [x] CODE - WEB
- [x] CODE - Android
  - App spinner still occurs, but the user is signed in if you close the app and reopen it, have mentioned to apps team 
- [x] CODE - iOS


raphaelkabo (approved):
This is great work, with the caveat we discussed that we don't need to be encrypting the payload twice (first encrypting the consents, then the whole thing), so we've made a note to fix that down the line. It's just a QOL thing anyway.

PR #2496, opened 2023-12-05
Soft opt in minor edge cases/refactor

coldlink:
## What does this change?

Update file structure so that the  `@/shared` folder should only ever import from the `@/shared` folder, so move around types and definitions to deal with this. This avoids the client/server bundle pollution while still allowing either to share code.

Move type definitions so that 'zod' isn't polluting the client bundle where it's not being used directly
  - Move any `zod`ified consents into its own file, so that when importing the type on the client side zod isn't included
  - This saves around 16kb gzipped in the client bundle (~10% of the vendors bundle)!
  - Screenshot 2023-11-27 at 09 43 33

*before*
Screenshot 2023-11-27 at 09 43 48

*after*
Screenshot 2023-11-27 at 09 44 54

Hide the soft opt in if the `useIdapi` flag is detected
  - In the unlikely scenario the user is using IDAPI rather than Okta to register, don't show the toggle as it doesn't support updating the field when using an IDAPI only flow.

### DIFF
--- a/src/server/lib/__tests__/okta/register.test.ts
+++ b/src/server/lib/__tests__/okta/register.test.ts
@@ -13,7 +13,6 @@ import {
 	Status,
 	UserCreationRequest,
 	TokenResponse,
-	RegistrationLocation,
 } from '@/server/models/okta/User';
 import {
 	validateRecoveryToken,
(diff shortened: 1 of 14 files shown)


raphaelkabo (approved):
Makes sense!

PR #2498, opened 2023-12-05
Tracking | `ref` and `refViewId` tracking through Okta and emails

coldlink:
## What does this change?

- Adds `ref` and `refViewId` tracking to the Okta Hosted Sign In Page interception code (OHSIPIC (oh-sip-ic? oh-sy-pic?)) so apps can send page view id and referral url to join tracking between the apps flow and web flow

- Adds `ref` and `refViewId` to the emails we send as part of the Okta flow so that we can track a user between devices/email inbox/urls etc.


cypress/integration/ete-okta/registration_1.2.cy.ts
```diff
@@ -219,7 +219,7 @@ describe('Registration flow - Split 1/2', () => {
 				const newReturnUrl = encodeURIComponent(
 					'https://www.theguardian.com/technology/2017/may/04/nier-automata-sci-fi-game-sleeper-hit-designer-yoko-taro',
 				);
-				cy.visit(`/welcome/${token}?returnUrl=${newReturnUrl}`);
+				cy.visit(`/welcome/${token}&returnUrl=${newReturnUrl}`);
```
raphaelkabo, line 222: Wait, how does this work? Where does the query string start?

coldlink: so buckle up,

the regex matcher on like `216` is `/welcome\/([^"]*)/` which essentially captures everything after the `/welcome` and puts it into the "token" variable, but it's effectively the token + anything after the token, which includes query parameters!

and this change means that the `ref` and `refViewId` is being sent on most emails for tracking purposes, meaning that the query string starts in the "token" variable 😅 

so if you had a url like `https://profile.theguardian.com/welcome/some_token?foo=bar`

then the `token` variable on this line is set to `some_token?foo=bar` not just `some_token`, and the query string is started as part of the token variable

raphaelkabo (approved):

PR #2499, opened 2023-12-11
Dependencies | December 11th 2023 

coldlink:
## What does this change?

Changes from previous https://github.com/guardian/gateway/pull/2493

- Runs `prettier:fix` to fix styling rule changes

Holds back the following:

## Tested

- [x] CODE


kelvin-chappell (commented):
Perhaps put the formatting changes in a separate commit so that they can be added to `.git-blame-ignore-revs`.  Described here: https://moxio.com/blog/ignoring-bulk-change-commits-with-git-blame/

coldlink:
> Perhaps put the formatting changes in a separate commit so that they can be added to `.git-blame-ignore-revs`. Described here: [moxio.com/blog/ignoring-bulk-change-commits-with-git-blame](https://moxio.com/blog/ignoring-bulk-change-commits-with-git-blame/)

good shout, will rebase do so, we've already previously done this in gateway https://github.com/guardian/gateway/blob/main/.git-blame-ignore-revs

kelvin-chappell (approved):
👍

PR #2503, opened 2023-12-18
Dependencies | December 18th 2023

coldlink:
## What does this change?

Changes from previous https://github.com/guardian/gateway/pull/2499

Holds back the following:

## Tested

- [x] CODE


kelvin-chappell (approved):
👍

PR #2517, opened 2023-12-27
Dependencies | December 27th 2023

raphaelkabo:
## What does this change?

Changes from previous https://github.com/guardian/gateway/pull/2503

- Brings our dependencies up to the latest versions where possible. This is a holiday week for most of the team, so only patch and minor dependencies were updated.

## Tested

- [ ] CODE


pvighi (approved):
not really a lot to review other than the numbers going up. If it's been tested to work on `CODE` then 👍

PR #2527, opened 2024-01-08
Dependencies | January 8th 2024

coldlink:
## What does this change?

Changes from previous https://github.com/guardian/gateway/pull/2524

- Moves `@types/supertest` from `dependencies` to `devDependencies`

## Tested

- [x] CODE


kelvin-chappell (approved):
👍

PR #2529, opened 2024-01-09
Fix Sign Out / Sign In issue

coldlink:
## The problem

The `GU_SO` cookie is set when the user signs out, which would tell applications to clear tokens if it exists when they come to check if a user is signed in.

However this cookie is cleared when a user signs in.

This can lead to a problem where if a user signs out, and then immediately signs in again, especially noticeable as a different user, and goes to an application like MMA, then it will appear that the user is signed in as the previous user if they were using MMA before, showing the data for the previous user.

## The solution

1. Check if valid access/id tokens exist, then check if the `GU_SO` cookie exist.
2. Check if the `iat` claim from the access/id token to the `GU_SO` cookie value.
3. If `iat` is after `GU_SO` then all is fine, the user signed in after the last sign out, if the `iat` < GU_SO then the tokens are invalid and for the previous session, so they should be cleared.
4. Make sure the `GU_SO` isn't cleared on sign in

## The PR

- Update the "Web Apps Integration Guide" with the new logic and flowchart
  - Moved to #2530
- Implement the new logic in `loginMiddlewareOAuth`
- Remove the `clearSignOutCookie` method

## Other changes

This logic needs to be replicated/changed in all user facing web applications, namely Gateway (here), MMA, dotcom (through [`identity-auth`](https://github.com/guardian/csnx/tree/main/libs/%40guardian/identity-auth)), and support.

## Tested

- [x] CODE
    - Signed out of an account first to set the `GU_SO` cookie
    - Signed into a new account
    - Visited `/consents` page to get a `GU_ACCESS_TOKEN` and `GU_ID_TOKEN` cookie set
    - Manually modified the `GU_SO` cookie value to be after the `iat` claim of the other two
    - Refreshed the page. Noticed that it did the authorization code flow again to get new tokens with `iat` claims set after the `GU_SO` cookie
    - Refreshed the page again, now with the new `iat` claims after the `GU_SO` cookie value, and no authorization code flow was performed, which meant that the app used the existing tokens
    - **Success!**

### DIFF
--- a/src/server/lib/middleware/login.ts
+++ b/src/server/lib/middleware/login.ts
@@ -73,26 +73,8 @@ export const loginMiddlewareOAuth = async (
 		);
 	}
 
-	// if a user has the GU_SO cookie, they have recently signed out
-	// so we need to clear any existing tokens
-	// and perform the auth code flow to get new tokens
-	// and log the user in if necessary
-	if (req.cookies.GU_SO) {
-		trackMetric('LoginMiddlewareOAuth::SignedOutCookie');
-
-		// clear existing tokens
-		checkAndDeleteOAuthTokenCookies(req, res);
-
-		// perform the auth code flow
-		return performAuthorizationCodeFlow(req, res, {
-			redirectUri: ProfileOpenIdClientRedirectUris.APPLICATION,
-			scopes: scopesForApplication,
-			confirmationPagePath: req.path as RoutePaths, //req.path will always be a RoutePaths
-		});
-	}
-
-	// no
-	// 2. has valid oauth access and id tokens
+	// 1. Check if the reader has a valid access token and id token.
+	// This does not mean that the current reader is definitely signed in, so we have to perform additional checks.
 	const accessTokenCookie = getOAuthTokenCookie(req, 'GU_ACCESS_TOKEN');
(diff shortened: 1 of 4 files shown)


src/server/lib/middleware/login.ts
```diff
@@ -114,6 +95,31 @@ export const loginMiddlewareOAuth = async (
 				scopesForApplication.includes(scope as Scopes),
 			)
 		) {
+			// valid tokens at this point
+			// 2. Check if the user has the `GU_SO` cookie
+			if (req.cookies.GU_SO) {
+				// if the user has a `GU_SO` cookie, compare the value to the `iat` claim in the access and id tokens
+				// if the `GU_SO` cookie (lastSignOut) is newer than the tokens, the user has recently signed out and we need to clear the tokens
+				const lastSignOut = parseInt(req.cookies.GU_SO, 10);
+				const accessTokenIat = accessToken.claims.iat!;
```
raphaelkabo, line 104: I ended up doing [this](https://github.com/guardian/manage-frontend/pull/1288/files#diff-44ef2ef130a808eb6420fdf16f81e941fb80cc6fa14d697c797c7768c317a0abR131) in MMA to handle the existence of `.claims.iat` - maybe you could do similar here to avoid the non-null assertion?

raphaelkabo (approved):
Great work.

PR #2530, opened 2024-01-10
Fix Sign Out / Sign In issue - Documentation Only

coldlink:
## What does this change?

Only updates the documentation for the issue described in #2529 so that we can start linking to the documentation while we update the other user facing applications to use this new flow.

### Problem

The `GU_SO` cookie is set when the user signs out, which would tell applications to clear tokens if it exists when they come to check if a user is signed in.
However this cookie is cleared when a user signs in.

This can lead to a problem where if a user signs out, and then immediately signs in again, especially noticible as a different user, and goes to an application like MMA, then it will appear that the user is signed in as the previous user if they were using MMA before.

### Solution

1. Check if valid access/id tokens exist, then check if the `GU_SO` cookie exist.
2. Check if the `iat` claim from the access/id token to the `GU_SO` cookie value.
3. If `iat` is after `GU_SO` then all is fine, the user signed in after the last sign out, if the `iat` < GU_SO then the tokens are invalid and for the previous session, so they should be cleared.
4. Make sure the `GU_SO` isn't cleared on sign in

### DIFF
--- a/docs/okta/web-apps-integration-guide.md
+++ b/docs/okta/web-apps-integration-guide.md
@@ -159,27 +159,30 @@ Refresh should be attempted by going through the Authorization Code flow again, 
 ```mermaid
 flowchart TD
 
-A[Has GU_SO cookie?]
-A -- Yes --> E[Clear tokens] --> SO[isSignedOut]
-A -- No --> B[Has access/id token and valid?]
-B -- Yes --> SI[isLoggedIn/maybeLoggedIn]
-B -- No --> C[has `GU_U` cookie?]
+A[Has access/id token and valid?]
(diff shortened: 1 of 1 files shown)


kelvin-chappell (approved):
Will we need some margin in the comparison of the value of `iat` and value of `GU_SO` if they're set by different systems?

coldlink:
> Will we need some margin in the comparison of the value of `iat` and value of `GU_SO` if they're set by different systems?

It'll be very unlikely, as both of these are set server-side, and you'd hope that servers are very close together in terms of time syncing.

Generally the only time issues we noticed were when client side browsers/devices didn't match up to the server time, e.g. when getting tokens, so on the clients we include a leeway.

PR #2261, opened 2023-05-15
/subscribe endpoint

tomrf1:
Adds a new endpoint for one-click subscribes from email links.
See [IDAPI PR](https://github.com/guardian/identity/pull/2397) for details of how this works.

I've refactored the existing /unsubscribe functionality, because both subscribing and unsubscribing work the same way.
- `src/server/routes/unsubscribe.ts` is now `src/server/routes/subscriptions.ts`, and handles both endpoints
- `UnsubscribeSuccessPage` is now `SubscriptionSuccessPage`
- `UnsubscribeErrorPage` is now `SubscriptionErrorPage`

And we now use `type SubscriptionAction = 'unsubscribe' | 'subscribe'` to vary behaviour across the handlers/components.

Tested in CODE:
Screenshot 2023-05-15 at 15 22 59

PR #2538, opened 2024-01-18
Show "Subscription Confirmation" page when confirming consents

coldlink:
## What does this change?

Previously when a user was sent an email to confirm their subscription to a consent, they would click on a CTA and either be taken to the `/consents/thank-you` page, or the sign in page.

In either case the user consent was set in the database, but they would not see a confirmation if they were not logged in already, and be shown the sign in page instead, which isn't great UX.

The `/consents/thank-you` page is a legacy identity page on the [`guardian/frontend`](https://github.com/guardian/frontend) repo that hasn't been updated in years, we were planning on removing or changing this page at some point, but had not been prioritised.

So instead we decided on a small change to just show a "Subscription Confirmation" page when a user confirmed their consent, which is the same as the page that was set up in https://github.com/guardian/gateway/pull/2261 for newsletters.

This means that all users get a confirmation page regardless if logged in or not, and we have a link going to MMA that allows a user to manage their email communications!

https://github.com/guardian/gateway/assets/13315440/3a5175e3-0347-428a-9d61-d92ba56f7d95

## Tested

- [x] Cypress E2E
- [x] CODE

### DIFF
--- a/cypress/integration/ete/consent_token/consent_token.2.cy.ts
+++ b/cypress/integration/ete/consent_token/consent_token.2.cy.ts
@@ -1,5 +1,5 @@
 describe('Consent token flow', () => {
-	it('redirects to the success route when supplied a valid token by a logged in user', () => {
+	it('shows the success page when supplied a valid token by a logged in user', () => {
 		cy.createTestUser({
 			isUserEmailValidated: true,
 		}).then(({ emailAddress, cookies }) => {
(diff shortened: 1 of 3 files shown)


tjmw (approved):
This looks great, thank you for making the change!

One quick overall question - is it at all a concern that now when a user lands on `/consent-token/<token>/accept` and the consent is set successfully, they'll remain on this page? This means that if they re-load the page, the consent will be set again I think (to the same thing). Or once a token is "used" does it expire? I can't think of any bad outcomes of doing this, just it's a change from the way things worked before where you were re-directed to a different page.

raphaelkabo (commented):
Lovely work. I have the same question as Tom - and if the token _does_ become invalid after it's used once, refreshing the page means they'll be shown a confusing page telling them they can request a new email to confirm their consent. Maybe it would be better to make the `/subscribe/success` page its own page, and HTTP redirect to it?

coldlink:
> Lovely work. I have the same question as Tom - and if the token _does_ become invalid after it's used once, refreshing the page means they'll be shown a confusing page telling them they can request a new email to confirm their consent. Maybe it would be better to make the `/subscribe/success` page its own page, and HTTP redirect to it?

@raphaelkabo @tjmw 

I made this work the same way as the newsletter subscription flow. The token doesn't actually expire in either flow consents or newsletters. But I'll go and change it as I also agree.

raphaelkabo (approved):
Looks great with the redirect. ✅

PR #2539, opened 2024-01-18
Skip onboarding flow if registration token has app prefix

coldlink:
## What does this change?

There is currently an unhappy path when a user tries to register within the app and clicks on the registration CTA within the email.

The expected behaviour is that when the user clicks the CTA, the native app should intercept the link and open it within the in-app browser flow instead, where everything works as expected for most users.

However some users when the click the CTA on their mobile device, the app doesn't intercept the link and the link opens in the users default browser instead (mobile web). When this happens and a user sets a password they have to go through the whole onboarding flow, and are redirected to the mobile web homepage at the end, which isn't helpful!

While we're unable to fix this bug yet until we investigate a deeplinking CTA or passwordless, we can minimise user frustration by removing the onboarding flow if this happens.

The best way to determine if this is happening is if the recovery token has the app prefix (`al_` for Android Live app, and `il_` for iOS live app), and the `confirmationPage` is pointing to the onboarding flow.

## Tested
- [x] CODE

### DIFF
--- a/src/server/routes/oauth.ts
+++ b/src/server/routes/oauth.ts
@@ -270,6 +270,21 @@ const authenticationHandler = async (
 			return res.redirect(303, authState.queryParams.fromURI);
 		}
 
+		// temporary fix: if the user registered on the app (the token will be prefixed),
+		// and instead of ending up back in the app but in a mobile browser instead,
+		//  where we don't want to show the onboarding,
+		// for now we simply redirect them to the default return url when app prefix is set,
+		// and the confirmation page is one of the consent pages, then we redirect to the default return url,
+		// which is normally the guardian home page, by setting authState.confirmationPage to undefined
+		// this will be fixed when we either use a deep link with a custom scheme, or passwordless OTP flow
+		if (
+			authState.data?.hasAppPrefix &&
+			consentPages.some((page) => page.path === authState.confirmationPage)
+		) {
+			// eslint-disable-next-line functional/immutable-data
+			authState.confirmationPage = undefined;
+		}
+
 		const returnUrl = authState.confirmationPage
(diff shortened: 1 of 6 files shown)


kelvin-chappell (dismissed):
Looks good 👍

src/server/routes/oauth.ts
```diff
@@ -270,6 +270,19 @@ const authenticationHandler = async (
...
 		}
 
+		// temporary fix: if the user registered on the app (the token will be prefixed),
+		// and instead of ending up back in the app but in a mobile browser instead,
+		// for now we simply redirect them to the default return url
+		// which is normally the guardian home page, by setting authState.confirmationPage to undefined
+		// this will be fixed when we either use a deep link with a custom scheme, or passwordless OTP flow
+		if (
+			authState.data?.hasAppPrefix &&
+			authState.confirmationPage === consentPages[0].path
```
raphaelkabo, line 280: I'm a bit confused on the logic here - why do we need to check if `confirmationPage` is the same as the first path in the `consentPages` array? Would it not be more futureproof if we just check that `confirmationPage` matches _any_ of the `consentPages` paths?

coldlink: `consentPages[0].path` is currently a pattern we use elsewhere in gateway (unfortunately), so I decided to use that. But I'll update it so that it could match either.

kelvin-chappell (dismissed):

src/server/routes/oauth.ts
coldlink: Updated!

raphaelkabo (approved):
Very nice! Good with the changes.

PR #2549, opened 2024-01-29
Dependencies | January 29th 2024

coldlink:
## What does this change?

Changes from previous https://github.com/guardian/gateway/pull/2547

- Updates `@swc/*` packages
- Updates `husky` to v`9`

## Tested

- [x] CODE


.husky/pre-commit
coldlink, line 1: file was generated after running `husky init` after upgrading

everytime `husky init` is run, it's automatically generater

.husky/_/husky.sh
coldlink, line 1: file deleted by running `husky init` after upgrading

kelvin-chappell (approved):
👍

PR #2560, opened 2024-02-06
Feast | Set up iOS Feast App links within gateway

coldlink:
## What does this change?

Sibling PR to https://github.com/guardian/identity-platform/pull/703

This updates gateway to make sure to add the new app prefix for the iOS feast app (`il_`) and make sure that it's correctly decrypted too.

Gateway will already be aware that Feast is an iOS app due to this line as the label within Okta will be `ios_feast_app`: https://github.com/guardian/gateway/blob/main/src/server/lib/middleware/requestState.ts#L61. In the future, especially if we're going to be serving different experiences to different apps, we may have to extend this functionality to use the app name instead.

We also need to update the [`apple-app-site-association`](https://github.com/guardian/gateway/blob/main/src/client/.well-known/apple-app-site-association) file so that these new links can be intercepted by the Feast app, but need the correct `appIDs` from the team.

### DIFF
--- a/scripts/okta/__tests__/removePrefixFromToken.test.ts
+++ b/scripts/okta/__tests__/removePrefixFromToken.test.ts
@@ -13,6 +13,10 @@ describe('removePrefixFromToken', () => {
 		expect(removePrefixFromToken('il_test')).toEqual('test');
 	});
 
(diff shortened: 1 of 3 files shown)


kelvin-chappell (approved):
👍

PR #2564, opened 2024-02-08
Passwordless Proof of Concept

coldlink:
👀


coldlink:
We're going to be implementing this after a successful demo.

We're going to start of with passcodes for registration only, and then expand them to sign in and reset password before going full passwordless.

PR #2569, opened 2024-02-15
Registration Platform | Make sure this is updated for social users

coldlink:
## What does this change?

In #2562 we added the ability to get the `registrationPlatform`, i.e. the app the user registered, into the user profile, which is helpful for tracking purposes.

However in hindsight we only set this up for users who created an account with email and password, and overlooked social users.

This PR fixes this by manually updating the `registrationPlatform` for social users on the same screen that they see the registration opt-out newsletters/consents (`/welcome/:social` page)

## Tested
- [x] CODE

### DIFF
--- a/src/server/lib/registrationPlatform.ts
+++ b/src/server/lib/registrationPlatform.ts
@@ -1,0 +1,66 @@
+import { logger } from '@/server/lib/serverSideLogger';
+import { Jwt } from '@okta/jwt-verifier';
+import { getUser, updateUser } from '@/server/lib/okta/api/users';
+import { getApp } from './okta/api/apps';
+
+/**
+ * @name getRegistrationPlatform
+ * @description Get the registration platform (app) based on the appClientId, this will be used to determine the platform the user is registering from, which is helpful for tracking purposes
+ *
+ * @param appClientId - the appClientId to get the app info for, corresponds to an app in Okta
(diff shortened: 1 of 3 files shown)


src/server/lib/registrationPlatform.ts
```diff
@@ -0,0 +1,66 @@
...
+	accessToken: Jwt,
+	appClientId?: string,
+	request_id?: string,
+): Promise<void> => {
+	if (!accessToken) {
+		throw new Error('No access token provided');
+	}
+
+	try {
+		const registrationPlatform = await getRegistrationPlatform(appClientId);
```
pvighi, line 47: is this making an api call or something like that ? if so we might want to delay getting this until after we determined the user doesn't already have a registration platform.

coldlink: This isn't making an API call for the user, it's only making an API call to the Okta Apps API in order to get the label (name of the app). We cache this response so it doesn't do it every time.

coldlink: This code is only used by social users who also don't have a registration platform set, as Okta creates the user during the authentication process through Google/Apple. But I've added a check further down to not update the registration platform for the user if they already have one set.

raphaelkabo (approved):
🚀

PR #2565, opened 2024-02-12
Dependencies | February 12th 2024

coldlink:
## What does this change?

Changes from previous https://github.com/guardian/gateway/pull/2559

- Brings our dependencies up to the latest versions where possible.
- Hopefully fixes flaky recaptcha tests by making sure to stub a `500` error when intercepting recaptcha calls
  - Adds a helpful `cy.interceptRecaptcha` command for this

Holds back the following (as in previous):

* `sharp` - no longer supports yarn 1, will update at the point we move to `pnpm`/`npm`

## Tested

- [x] CODE

PR #2578, opened 2024-02-19
Dependencies | February 19 2024

raphaelkabo:
## What does this change?

Changes from previous https://github.com/guardian/gateway/pull/2565

Holds back the following:

* `sharp` - no longer supports yarn 1, will update at the point we move to `pnpm`/`npm` (as in previous)
* `react`, `testing-library/preact`, and `@testing-library/preact-hooks`, because Yarn wants all of these to be upgraded from their current (up-to-date, according to npm) versions to `'exotic'`, whatever that is. 

Screenshot 2024-02-19 at 14 38 49

## Tested

- [ ] CODE


kelvin-chappell (approved):
👍

coldlink:
So the `'exotic'` thing is because those dependencies are being mapped to a different npm package than its name. This allows us to use Preact instead of React, but keep everything under the React namespace! https://github.com/guardian/gateway/pull/2161

Issue #2567, opened 2024-02-13
Passwordless | Registration only with One Time Passcodes


coldlink:
## Why?

We've now upgraded to Okta Identity Engine. This unlocks a number of new features which previously weren't feasible. One of these is "passwordless", or the idea of eliminating, or reducing the need for passwords.

We also currently have an issue within our OAuth flows where users don't always end up signed in the same browser context that they registered from. This can occur if a user clicks a reset password/registration verification link in their on another browser or device, or if deeplinking/https interception isn't working properly on the user's device.

One way of fixing both these issues it to use One Time Passcodes (OTPs), where instead of a user clicking a link in an email, they have to type a 6 digit number into an input field which performs the same action of verifying the user. This means that the user always stays within the same browser device/context, while also securely verifying the user.

To begin with we're going to only implement OTPs for registration only, with the user still having to set a password. After this we will being rolling out full passwordless.

## Flowchart

what have i done...

```mermaid
flowchart TD;

start[\User comes to Gateway\] -- Register Page --> register;

register[User enters email address] --> register-interact;

register-interact([POST /oauth2/auth_server/v1/interact]) -- Get interction_handle --> register-introspect;

register-introspect([POST /idp/idx/introspect\ninteractionHandle in body]) -- Get and store stateHandle\nin encryptedState\nunless specified, all calls\nuse stateHandle in body --> register-enroll;

register-enroll([POST /idp/idx/enroll]) -- Error --> register-enroll-error;
register-enroll([POST /idp/idx/enroll]) -- Success --> register-enroll-email;

register-enroll-error{{What error was it?}};
register-enroll-error -- unknown error --> register-enroll-error-unknown;
register-enroll-error -- stateHandle\nexpired--> register-enroll-error-expiry;

register-enroll-error-unknown{{An error case which doesn't match any of the others}} --> okta-classic-registration-fallback;
register-enroll-error-expiry{{User took too long.\nDifficult to get here as\nexpiry of stateHandle is\n2 hours here}} -- restart with email --> register-interact;

register-enroll-email([POST /idp/idx/enroll/new\nuserProfile: email, isGuardianUser,\nregistrationPlatform, registrationLocation]);
register-enroll-email -- Error --> register-enroll-email-error;
register-enroll-email --> register-email-sent;

register-enroll-email-error{{What error was it?}};
register-enroll-email-error -- unknown error --> register-enroll-email-error-unknown;
[...]

PR #2580, opened 2024-02-20
Show specific registration complete page for some app users

coldlink:
## What does this change?

In #2539 we fixed an unhappy path when a user tries to register within the app, where they get stuck inside their web browser and get redirected to the onboarding flow. In that PR we skipped the onboarding flow and redirected them to the Guardian homepage.

However this wasn't ideal, as users still weren't signed into the app, this will eventually be solved by using passcodes for registration (see #2567).

In the meantime we're planning on showing them a simple page which tells them to go back to the Guardian app they came from and sign in again.

localhost_6006_iframe html_id=pages-returntoapp--default viewMode=story(iPhone 12 Pro)

Storybook: https://60929d168594f80039336501-vyhifhwooi.chromatic.com/?path=/story/pages-returntoapp--default

## Tested

- [x] Cypress
- [x] CODE - Affected user
- [x] CODE - Non affected user

### DIFF
--- a/cypress/integration/ete-okta/registration_1.2.cy.ts
+++ b/cypress/integration/ete-okta/registration_1.2.cy.ts
@@ -138,7 +138,9 @@ describe('Registration flow - Split 1/2', () => {
 				cy.get('input[name="secondName"]').type('Last Name');
 				cy.get('input[name="password"]').type(randomPassword());
 				cy.get('button[type="submit"]').click();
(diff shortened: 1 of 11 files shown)


raphaelkabo (approved):
This looks very good! ✅

coldlink:
> This looks very good! ✅
> 
> Sidenote: why did Storybook not complain about new stories being added?

I'm not sure, the component is there and it looks fine?

PR #2587, opened 2024-02-26
Dependencies | February 26th 2024

coldlink:
## What does this change?

Changes from previous https://github.com/guardian/gateway/pull/2578

Holds back the following:

* `sharp` - no longer supports yarn 1, will update at the point we move to `pnpm`/`npm` (as in previous)

## Tested

- [x] CODE


kelvin-chappell (approved):
👍

PR #2598, opened 2024-02-29
AB Testing | Send offer emails to new users coming from specific sign in gates

raphaelkabo:
## What does this change?

- Adds two new email templates, for emails which are sent _after_ a user successfully registers _if_ that user has come from one of the two sign-in gates we'll be AB testing on DCR: https://github.com/guardian/dotcom-rendering/pull/10767/files
- Fixes a `tsconfig` issue in Cypress - not clear why this popped up
- Fixes email box padding styles
- Adds some methods to capture the `signInGateId` on `GET /signin/:social` and `POST /register`.
    - In the social case, the ID (if it's one of the ones we care about) is passed through the auth code flow and is accessed in the callback to send the appropriate email.
    - In the email/password case, it's saved as a query param on the 'verify email' link, which passes through to the `POST` set password call, and sends the email if that is successful.

## Tests

- [x] Tested on CODE

### DIFF
--- a/src/email/templates/GuardianLiveOffer/GuardianLiveOffer.tsx
+++ b/src/email/templates/GuardianLiveOffer/GuardianLiveOffer.tsx
@@ -1,0 +1,36 @@
+import React from 'react';
+
+import { Mjml, MjmlBody, MjmlHead, MjmlTitle } from '@faire/mjml-react';
+import { Header } from '@/email/components/Header';
+import { SubHeader } from '@/email/components/SubHeader';
+import { Text } from '@/email/components/Text';
+import { Footer } from '@/email/components/Footer';
+import { Button } from '@/email/components/Button';
+
+export const GuardianLiveOffer = () => {
+	return (
+		<Mjml>
+			<MjmlHead>
+				<MjmlTitle>
+					Your Guardian account and discount code | The Guardian
+				</MjmlTitle>
+			</MjmlHead>
+			<MjmlBody width={600}>
+				<Header />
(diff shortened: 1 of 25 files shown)


src/email/templates/GuardianLiveOffer/GuardianLiveOffer.tsx
```diff
@@ -0,0 +1,36 @@
...
+			<MjmlBody width={600}>
+				<Header />
+				<SubHeader>Welcome to the Guardian</SubHeader>
+				<Text>Thank you for creating an account with the Guardian.</Text>
+				<Text>
+					Here is your Guardian Live discount code, allowing 20% off all
+					livestreamed events.
+				</Text>
+				<Text>
+					<strong>LIVEREG20</strong>
```
coldlink, line 27: Have we checked this code actually works? I just tried applying it to a random event I saw online and it didn't work.

src/server/lib/ophan.ts
```diff
@@ -70,11 +80,48 @@ export const parseComponentEventParams = async (
...
+	return getMatchingSignInGateId(componentEventParams.abTestVariant);
+};
+
+/* AB TEST END */
 
 /**
  * In some cases in DCR and Frontend we send the component type
  * as a lowercase string with no spaces, so we need to convert them
- * to the correct case (upper snake case)
+ * to the correct case (SCREAMING_SNAKE_CASE)
```
coldlink, line 124: 😱 🐍

coldlink (dismissed):

coldlink (approved):

PR #2607, opened 2024-03-04
Add new Feast fields to MDAPI response schema

raphaelkabo:
## What does this change?

The MDAPI user attributes response schema is getting two new fields in this PR: https://github.com/guardian/members-data-api/pull/1045

As our Zod schema for this response is `.strict()`, the Zod parser fails on MDAPI responses which have the new fields. This adds the new fields to the Zod schema.

As a side note, should we remove `.strict()` from the MDAPI response? With it set to strict, we need to know whenever other teams change the response.

## Tests

- [ ] Tested in CODE


coldlink:
Definitely remote the `.strict` check!

coldlink (approved):

PR #2613, opened 2024-03-06
Remove unused code

coldlink:
## What does this change?

- Removes the unused "Magic Link" code, which was never implemented
  - This is to clean up in preparation for passcodes and passwordless
- Removed the `sendForgotPasswordEmail` function
  - This was no longer in use ever since we started sending emails through gateway in #2058
- Removes unused exports after running [ts-prune](https://github.com/nadeesha/ts-prune)
  - In a related note, `ts-prune` has been archived. We should switch to using [knip](https://github.com/webpro/knip) as suggested


raphaelkabo (approved):
Beautiful 🥲

PR #2606, opened 2024-03-04
Dependencies | 4 March 2024

raphaelkabo:
## What does this change?

Changes from previous https://github.com/guardian/gateway/pull/2587

- Brings our dependencies up to the latest versions where possible.

Holds back the following:

* `sharp` - no longer supports yarn 1, will update at the point we move to `pnpm`/`npm` (as in previous)

## Tested

- [x] CODE

PR #2623, opened 2024-03-12
Dependencies | 12th March 2024 

coldlink:
## What does this change?

Changes from previous https://github.com/guardian/gateway/pull/2606

Holds back the following:

* **NEW** Storybook 8 - no longer supports yarn 1, will update at the point we move to `pnpm`/`npm`

## Tested

- [x] CODE


pvighi (approved):
numbers go up so it must be good 🤷

PR #2625, opened 2024-03-12
IDX API / Interaction Code Flow | API Setup + Social Sign In

coldlink:
## What does this change?

The Okta Identity Engine upgrade enables us to use new features, such as passwordless authentication, through their new [Interaction Code Flow](https://developer.okta.com/docs/concepts/interaction-code/) and [IDX API](https://github.com/okta/okta-auth-js/blob/master/docs/idx.md). This also enables us to build our own custom login flows without the need for using the Okta hosted sign in page.

However the IDX API doesn't have its endpoints publicly documented, and instead Okta would prefer we use their own SDKs in order to interface with this API. Unfortunately this isn't viable for us as the SDKs don't perform ideally when using them in a server side environment. So we've taken the option to reverse engineer the IDX API.

We did this by seeing what calls the default Okta hosted sign in page made when authenticating, and seeing what calls we needed to make too.

For social sign in the steps are as follows:

1. Call the `interact` endpoint to start the interaction code flow.
    - The parameters for this call are the same as those provided to the Authorization Code Flow with PCKE
    - But instead make a `POST /oauth2/<auth_server_id>/v1/interact` call, with the parameters encoded in the body as `application/x-www-form-urlencoded`
    - This returns an object with the `interaction_handle` key
2.  Call the `introspect` endpoint with the `interaction_handle` in the body as JSON.
    - `POST /idp/idx/introspect`
    - This returns a whole bunch of stuff, essentially shows all of the authentication options we can perform
    - It also returns a `stateHandle` which is important for identifying the current authentication request, and should be preserved in any future calls to the IDX API
3. From the `introspect` response, in the `remediation` array, find an object with the key `name` and value `redirect-idp`, and the key `type` to be value either `GOOGLE` or `APPLE` depending on which one to authenticate with
4. Once you have this, get the `href` from the same object, and redirect the user to this `href`
5. The user then authenticates with their social provider, and once authenticated redirects back to Okta's `/oauth2/v1/authorize/callback` endpoint
6. This is an Okta controlled endpoint that either creates the user, or signs the user in, and then redirects the user to the callback endpoint defined in the parameters of step 1
7. This callback endpoint receives an `interaction_code` and a `state` (the state was defined in step 1)
8. Redirect the user to `/login/token/redirect?stateToken=${stateToken}`
    - Where `stateToken` is the `stateHandle` from step 2, but only the first part split at the `~`
    - `stateHandle.split('~')[0]`
    - This sets a global okta session, and completes the flow by redirecting the user to the same callback endpoint, but with a `code` parameter (not an `interaction_code`) and the state
9. In this callback endpoint, use a standard OAuth library that supports Authorization Code Flow with PKCE, and complete the flow.
    - The user is authenticated at this point 

This PR does 2 main things
- Sets up the basic framework for the IDX API and Interaction Code flow
- Updates social sign in to use the new flow

The functionality has been set up so that should the IDX API fail for whatever reason, it will fall back to using the old authentication mechanism i.e Okta Classic.

### Set up IDX API

- Set up the new callback URL in our routes definition and OAuth configuration
    - [`d5a77f2` (#2625)](https://github.com/guardian/gateway/pull/2625/commits/d5a77f2e3f28989ed0631de587a306fefb2b8acb)
- Allow the `AuthorizationStateCookie` to preserve the `codeVerifier` (for PKCE), and the `stateToken` (for authentication)
    - [`d5a77f2` (#2625)](https://github.com/guardian/gateway/pull/2625/commits/d5a77f2e3f28989ed0631de587a306fefb2b8acb)
- Set up the `interact` method, schema, and error handling  
    - [`2f89fae` (#2625)](https://github.com/guardian/gateway/pull/2625/commits/2f89faec8338f1bd4eb36f21ad7ef911a545b33e)
- Set up the `introspect` method, schema, and error handling
    - [`c7a049f` (#2625)](https://github.com/guardian/gateway/pull/2625/commits/c7a049fc5edd9fb36a6bb63d503972b828dfe747)
 
### Set up social authentication

[`a12e131` (#2625)](https://github.com/guardian/gateway/pull/2625/commits/a12e131059318141bd1e4581d2af94da72984e7c)

- Set up the `/oauth/authorization-code/interaction-code-callback` route handling
- Allow the `AuthorizationState` to be updated after it's initially generated
    - So we can add the `stateToken` in once thats available to us
- Update the `GET /signin/:social` route to follow the steps mentioned above for social sign in

## Tested
- [x] CODE - Web - Google - Sign In
- [x] CODE - Web - Google - Create account
- [x] CODE - Web - Apple - Sign in
- [x] CODE - Web - Apple - Create account
- [x] CODE - iOS - Google - Sign in
- [x] CODE - iOS - Google - Create account
- [x] CODE - iOS - Apple - Sign in
- [x] CODE - iOS - Apple - Create account
- [x] CODE - Android - Google - Sign in
- [x] CODE - Android - Google - Create account
- [x] CODE - Android - Apple - Sign in
- [x] CODE - Android - Apple - Create account

### DIFF
--- a/src/server/lib/okta/idx/interact.ts
+++ b/src/server/lib/okta/idx/interact.ts
@@ -1,0 +1,168 @@
+import { Request } from 'express';
+import { ResponseWithRequestState } from '@/server/models/Express';
+import { logger } from '@/server/lib/serverSideLogger';
+import { AuthorizationParameters, generators } from 'openid-client';
+import { getConfiguration } from '@/server/lib/getConfiguration';
+import {
+	AuthorizationState,
+	ProfileOpenIdClientRedirectUris,
+	generateAuthorizationState,
+	setAuthorizationStateCookie,
+} from '@/server/lib/okta/openid-connect';
+import {
+	PerformAuthorizationCodeFlowOptions,
+	scopesForAuthentication,
+} from '@/server/lib/okta/oauth';
+import { joinUrl } from '@guardian/libs';
+import { z } from 'zod';
+import { closeCurrentSession } from '@/server/lib/okta/api/sessions';
+import { getPersistableQueryParams } from '@/shared/lib/queryParams';
+import { OAuthError, isOAuthError } from '@/server/models/okta/Error';
+
+const { okta } = getConfiguration();
+
+const interactResponseSchema = z.object({
+	interaction_handle: z.string(),
+});
+export type InteractResponse = z.infer<typeof interactResponseSchema>;
+
+/**
+ * @name interact
+ * @description Okta IDX API/Interaction Code flow - Step 1: Initiates the interaction code flow, and returns an interaction handle which should be used in the next step, specifically `introspect`
+ *
+ * This is very similar to the standard authorization code flow, using similar parameters and the PKCE flow, but the authentication
+ * is different.
+ *
+ * In Gateway we only use the interaction code flow for authentication, in order to avoid using the Okta hosted sign in page. The
+ * standard authorization code flow is used for everything else.
+ *
+ * @param req - Express request
+ * @param res - Express response
+ * @param AuthorizationCodeFlowOptions - Subset of the `PerformAuthorizationCodeFlowOptions` used by our standard authorization code flow, namely the parameters needed for authentication
+ * @returns Promise<[InteractResponse, AuthorizationState]> - Array containing the interaction handle and the authorization state
+ */
+export const interact = async (
+	req: Request,
+	res: ResponseWithRequestState,
+	{
+		confirmationPagePath,
+		closeExistingSession,
+		doNotSetLastAccessCookie = false,
+		extraData,
+	}: Pick<
+		PerformAuthorizationCodeFlowOptions,
+		| 'confirmationPagePath'
+		| 'closeExistingSession'
+		| 'doNotSetLastAccessCookie'
+		| 'extraData'
+	>,
+): Promise<[InteractResponse, AuthorizationState]> => {
+	try {
+		// determines if the current session, should it exist, be closed
+		if (closeExistingSession) {
+			// Okta Identity Engine session cookie is called `idx`
+			const oktaIdentityEngineSessionCookieId: string | undefined =
+				req.cookies.idx;
+			if (oktaIdentityEngineSessionCookieId) {
+				await closeCurrentSession({
+					idx: oktaIdentityEngineSessionCookieId,
+				});
+			}
+		}
+
+		// generate the code verifier and code challenge for PKCE
+		// see https://www.oauth.com/oauth2-servers/pkce/authorization-request/
+		// the `code_verifier` is a cryptographically random string
+		// the `code_challenge` is the `code_verifier` hashed with SHA-256 and then base64url encoded
+		const codeVerifier = generators.codeVerifier(64);
+		const codeChallenge = generators.codeChallenge(codeVerifier);
+
+		// generate and store a "state"
+		// as a http only, secure, signed session cookie
+		// which is a json object that contains a stateParam and the query params
+		// the stateParam is used to protect against csrf
+		// we also set additional data in the state cookie that we need to persist
+		// between the authorization request and the callback
+		const authState = generateAuthorizationState(
+			getPersistableQueryParams(res.locals.queryParams),
+			confirmationPagePath,
+			doNotSetLastAccessCookie,
+			{
+				appPrefix: extraData?.appPrefix,
+				encryptedRegistrationConsents: extraData?.encryptedRegistrationConsents,
+				codeVerifier,
+			},
+		);
+		setAuthorizationStateCookie(authState, res);
+
+		// set up the authorization parameters to send to the /interact endpoint
+		// in order to start the interaction code flow
+		// see https://developer.okta.com/docs/concepts/interaction-code/
+		const authorizationParams: AuthorizationParameters = {
+			// the client_id is the id of the client application
+			client_id: okta.clientId,
+			// only use the interaction code flow callback uri, everything else is handled by standard authorization code flow
+			redirect_uri: ProfileOpenIdClientRedirectUris.INTERACTION_CODE,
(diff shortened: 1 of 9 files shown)


src/server/lib/okta/idx/interact.ts
```diff
@@ -0,0 +1,168 @@
...
+
+		// set up the authorization parameters to send to the /interact endpoint
+		// in order to start the interaction code flow
+		// see https://developer.okta.com/docs/concepts/interaction-code/
+		const authorizationParams: AuthorizationParameters = {
+			// the client_id is the id of the client application
+			client_id: okta.clientId,
+			// only use the interaction code flow callback uri, everything else is handled by standard authorization code flow
+			redirect_uri: ProfileOpenIdClientRedirectUris.INTERACTION_CODE,
+			// only authentication scopes for interaction code flow, everything else is handles by standard authorization code flow
```
raphaelkabo, line 107:
```suggestion
			// only authentication scopes for interaction code flow, everything else is handled by standard authorization code flow
```

src/server/lib/okta/idx/interact.ts
```diff
@@ -0,0 +1,168 @@
...
+			// we send the generated code challenge
+			code_challenge: codeChallenge,
+			// we use the S256 code challenge method, which is the only method supported
+			code_challenge_method: 'S256',
+			// we send the generated stateParam as the state parameter
+			state: authState.stateParam,
+		};
+
+		const authorizationSearchParams = new URLSearchParams(
+			authorizationParams as Record<string, string>,
```
raphaelkabo, line 118: Is this typecast really necessary here?

src/server/lib/okta/idx/interact.ts
```diff
@@ -0,0 +1,168 @@
...
+				body: authorizationSearchParams.toString(),
+			},
+		);
+
+		if (!response.ok) {
+			// Check if the body is likely json using the content-type header
+			const contentType = response.headers.get('content-type');
+			if (contentType && contentType.includes('application/json')) {
+				// if so, parse the body as json, and handle the error
+				const error = await response.json();
```
raphaelkabo, line 137: This _will_ throw in the unlikely scenario where `Content-Type` is JSON, but the body isn't (for instance, if it's an empty string). I would do something like:

```suggestion
				const error = response.json().then(j).catch((e) => new OAuthError(
					{
						error: 'unknown_error',
						error_description: await response.text(),
					},
					response.status,
				));
```

(This may not compile).

src/server/lib/okta/idx/introspect.ts
```diff
@@ -0,0 +1,143 @@
...
+					method: z.string().optional(),
+				}),
+			]),
+		),
+	}),
+});
+export type IntrospectResponse = z.infer<typeof introspectResponseSchema>;
+
+const introspectResponseErrorSchema = z.object({
+	version: z.literal('1.0.0'),
```
raphaelkabo, line 43: What happens with Zod if they bump the version - nothing interesting? Still parses? Does this really need to be a literal?

src/server/lib/okta/idx/interact.ts
coldlink: Unfortunately yes, as the `AuthorizationParams` type doesn't overlap with the URLSearchParams input type, even though it's mostly `Record<string, string>`.

```
Argument of type 'AuthorizationParameters' is not assignable to parameter of type 'string | Record<string, string> | string[][] | URLSearchParams | undefined'.
Type 'AuthorizationParameters' is missing the following properties from type 'string[][]': length, pop, push, concat, and 29 more.ts(2345)
⚠ Error (TS2345)  | 

Argument of type 
AuthorizationParameters
 is not assignable to parameter of type:
| string
| Record<string, string>
| string[][]
| URLSearchParams
| undefined
.
  | Type  AuthorizationParameters  is missing the following properties from type  string[][]  :lengthpoppushconcatand 29 more. -- | --
  ```

File: src/server/lib/okta/idx/introspect.ts [hunk 5]
```diff
@@ -0,0 +1,143 @@
...
+		throw error;
+	}
+};
+
+const handleError = async (response: Response) => {
+	// Check if the body is likely json using the content-type header
+	const contentType = response.headers.get('content-type');
+	if (
+		contentType &&
+		(contentType.includes('application/json') ||
```
raphaelkabo, line 110: Could just check if `Content-Type` includes `'json'`...avoids dealing with them deciding to change to a different serialization dialect down the line (although I suspect this would be the least of our concerns at that point).

src/server/lib/okta/idx/introspect.ts
```diff
@@ -0,0 +1,143 @@
...
+			contentType.includes('application/ion+json'))
+	) {
+		// if so, parse the body as json and see if it matches the schema
+		const json = await response.json();
```
raphaelkabo, line 114: Same thing as above, about potentially parsing an empty string going straight to the `catch` for no real reason.

src/server/lib/okta/idx/introspect.ts
```diff
@@ -0,0 +1,143 @@
...
+		const error = introspectResponseErrorSchema.safeParse(json);
+
+		if (error.success) {
```
raphaelkabo, line 117: I like the idea of `error.success`.

src/server/lib/okta/idx/introspect.ts
coldlink: I'd rather be explicit in the definition, as yeah if they change how they're serialising stuff, we have bigger problems 😓

src/server/lib/okta/idx/introspect.ts
coldlink: It will fail to parse I think, as it's expecting a literal string, so anything else will fail.

Maybe I should instead look for a string that starts with `1.`?

This is assuming they're even semver for the api

coldlink: Maybe I'll add a log if it doesn't match `1.0.0` but will still let the validation pass.

raphaelkabo (approved):
🚀

PR #2626, opened 2024-03-13
Switch to `pnpm` from `yarn` `1.x`

coldlink:
## What does this change?

Switches to use [`pnpm`](https://pnpm.io/) instead of [yarn 1](https://classic.yarnpkg.com/lang/en/) as the package manager for gateway.

pnpm is a self described "fast, disk space efficient package manager". 

The main reason for using pnpm over yarn is that it is much faster compared to yarn, which should speed up development and CI. In most cases installing and adding dependencies is twice as fast compared to yarn.

Also some dependencies no longer support yarn 1, so we can't upgrade them, e.g. `sharp` and storybook.

The large number of changes in this file is because we've removed the `yarn.lock` file, and replaced with a `pnpm-lock.yml`.

Other changes are mostly changed the `yarn` commands to `pnpm` commands, and a fix to some typescript errors in test files.

## Migration

- Checkout this branch, or `main` once this is merged in
- Remove existing `node_modules` folder (`rm -r node_modules`)
- Run `make install` which sets up `pnpm` (which you can also do by running `make setup`) and installs dependencies again
  - You can also run `make dev` if you want to do setup, install, and start the dev server

## Tested
- [x] DEV
- [x] CODE


.github/workflows/cypress-ete-okta.yml
```diff
@@ -56,10 +56,28 @@ jobs:
...
+      - name: Cache pnpm modules and cypress binary
+        uses: actions/cache@v4
+        id: pnpm-cache
+        with:
+          path: |
+            ${{ steps.pnpm-cache-dir.outputs.PNPM_STORE_PATH }}
+            ${{ steps.cypress-cache-dir.outputs.CYPRESS_BIN_PATH }}
+          key: ${{ runner.os }}-pnpm-cypress-${{ hashFiles('**/pnpm-lock.yaml') }}
+          restore-keys: |
+            ${{ runner.os }}-pnpm-cypress-
```
coldlink, line 80: this crazy stuff lets us cache the modules and cypress binaries so they can be shared between all of the different workflows and machines

src/server/lib/__tests__/updateRegistrationLocation.test.ts
```diff
@@ -26,14 +26,15 @@ jest.mock('@/server/lib/serverSideLogger', () => ({
...
-});
+): Request =>
+	({
+		cookies: {
+			GU_geo_country: country,
+		},
+		body: {
+			_cmpConsentedState: consent,
+		},
+	}) as unknown as Request;
```
coldlink, line 37: not sure why this started failing when running unit tests, but using `as` syntax is fine here as its for a test

.github/workflows/chromatic.yml
```diff
@@ -18,12 +18,15 @@ jobs:
...
+        with:
+          version: latest
       - uses: actions/setup-node@v4
         with:
           node-version-file: '.nvmrc'
-          cache: 'yarn'
+          cache: 'pnpm'
       - name: Install dependencies
-        run: yarn
+        run: pnpm install --frozen-lockfile
```
raphaelkabo, line 29: Theoretically, this is unnecessary (https://pnpm.io/cli/install#--frozen-lockfile):

> This setting is true by default in [CI environments](https://github.com/watson/ci-info#supported-ci-tools).

But I don't see any harm in leaving it in!

package.json
```diff
@@ -5,23 +5,22 @@
 	"license": "Apache-2.0",
 	"private": true,
 	"scripts": {
-		"build": "ENVIRONMENT=production webpack --mode production && yarn install --modules-folder ./build/node_modules --production --frozen-lockfile",
-		"build:analyze": "WEBPACK_ANALYZE_BUNDLE=true yarn build",
+		"build": "ENVIRONMENT=production webpack --mode production && pnpm install --prod --node-linker=hoisted --dir=build --ignore-scripts",
```
raphaelkabo, line 8: Are we setting the linker to `hoisted` just because it's safer for production?

coldlink: Yeah it's just the way our production build works. For some reason not using hoisted caused it to fail. Need to investigate whether this is something we can change, but for now this replicates the existing behaviour.

makefile
```diff
@@ -16,111 +16,116 @@ define banner
 endef
 
 # prod
-build: clear clean-build install
-	$(call log, "building production bundles")
-	@yarn build
+build: clear clean-build setup install
+	$(call log, "building production bundles 📦")
```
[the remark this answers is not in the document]
coldlink: :shipit:

.github/workflows/chromatic.yml
coldlink: Ah didn't realise this! But it's probably fine, and better to be explicit too.

raphaelkabo (approved):
👏 👏 👏

PR #2627, opened 2024-03-14
Create braze user for new users

coldlink:
## What does this change?

Users aren't created in braze until the overnight job runs (OAW Dump) or they subscribe to a newsletter, however a use case has come up where a user needs to exist in braze when an account is created, specifically for the Feast team to send service emails to users who create an account.

We created a new endpoint in IDAPI to facilitate this: https://github.com/guardian/identity/pull/2500

This PR uses that new endpoint in Gateway.

It calls this endpoint when a user updated their registration consents, but doesn't have a newsletter they're subscribing too, or when they're a new social user.

If a user subscribed to a newsletter, then we don't call this endpoint as they'll get automatically created in braze. If they don't then we call this endpoint to create them.

## Tested

- [x] CODE

### DIFF
--- a/src/server/lib/idapi/newsletters.ts
+++ b/src/server/lib/idapi/newsletters.ts
@@ -3,6 +3,7 @@ import {
 	APIGetOptions,
 	APIPatchOptions,
 	APIOptionSelect,
+	APIPostOptions,
 } from '@/server/lib/IDAPIFetch';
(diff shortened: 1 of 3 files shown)


tjmw (approved):
Great! This looks good to me (appreciate that's of limited value from a review perspective as I don't know this repo well).

Thank you @coldlink!

PR #2637, opened 2024-03-19
Okta IDX API | API setup for passcode registration

coldlink:
## What does this change?

This PR sets up the rest of the IDX API calls needed for passcode registration. It builds upon the initial setup in https://github.com/guardian/gateway/pull/2625

It does not actually use any of these APIs yet. This will be set up in a separate PR.

In general the IDX API works in the following steps in order to authenticate a user.

1. Start the interaction code flow using the `interact` call (not modified in this PR). This returns an `interaction_handle`, which as far as i can tell is only used to get the `stateHandle` from the next step, and can be discarded. The `interact` calls takes the same parameters as an Authorization Code flow call. So we need to set and preserve anything in the `AuthorizationState` at this point too, so that it can be read by the callback endpoint at the end.
2. Call the `introspect` endpoint with the `interaction_handle` to get the `stateHandle`. The `stateHandle` should be securely preserved and used in all following calls to the IDX API to identify the authentication state. The `introspect` calls returns the start of the authentication process, with a number of `remediation` steps to take to get the user authenticated. We add the `stateHandle` to be preserved in the `AuthorizationState` too.
3. Pick a `remediation` option from the `introspect` call, and call it. This could be, as will be set up for passcode registration, creating a new user by calling the `enroll` endpoint
4. The endpoint that you call will return another object with a `remediation` step within it. Keep checking the `remediation` object for next steps and taking the action it says.
5. Eventually you'll call an endpoint that completes the flow. In this case it will NOT return an `remediation` step, and instead return a `Set-Cookie` header with an `idx` cookie, and a basic `user` object in the response.
6. Since we're taking all these actions server side, we need to extract the `Set-Cookie` header for the `idx` cookie, and make sure this is set on the users response object. However this `idx` cookie weirdly isn't a global session cookie, and can't be used to create OAuth tokens.
7. You have to redirect the user to the `/login/token/redirect?stateToken=${stateHandlePart}` where `stateHandlePart` is the `stateHandle` but split at the `~` character, and take the first part. Weird I know, but we roll with it. This converts the `idx` cookie into a global session cookie, and completes the Interaction Code flow, in the same way as an Authorization Code flow is completed, i.e redirecting to the OAuth callback url with the  `code` and `state` params.
8. The user has now authenticated!

This PR sets up the following IDX API endpoint

- A `shared` file, with lots of methods and options that are used by all of the different IDX API calls
  - This includes a base `fetch` call that handles the call to the IDX API
  - Base Zod schemas and types for the response schemas, remediation schemas, and error schemas
  - Error handling for all the IDX API endpoints
  - Handling for the completion call and redirect to the `/login/token/redirect?stateToken=${stateHandlePart}` endpoint and setting the `idx` cookie
- Updates the `introspect` file
  - To use the new base `idxFetch` method instead of implementing it itself. 
  - And to allow the `introspect` call to use the `stateHandle` as an option in the body alongside the `interaction_handle`
    - This allows us to check the state of the authentication process at any time
- The `enroll` file
  - `enroll` endpoint that starts the process to register a new user
  - `enroll/new` endpoint to create a new user in Okta, this sends the user a passcode
- The `challenge` file
  - A `challenge/answer` endpoint to supply the OTP to that the user would enter (`challengeAnswerPasscode`)
  - A `setPasswordAndRedirect` method, that also calls `challenge/answer` but to set a password instead, but also is the final step in the registration process, so logs the user in and redirects them.
- The `credential` file
  - `credential/enroll` endpoint in order to enroll a new credential for the user, currently only supports `password` but could be expanded in the future
- Updates the `Metrics` so that the metric is automatically based on the IDX API path that we're calling.

## Notes

As mentioned, in this PR none of these new methods/APIs are being used with the exception of `interact` and `introspect` which were previously set up for social sign in (#2625). So there isn't any tests set up yet.

When we come to implementing this API in the frontend we will do this.

### DIFF
--- a/src/server/lib/okta/idx/shared.ts
+++ b/src/server/lib/okta/idx/shared.ts
@@ -39,18 +39,17 @@ export type IdxStateHandleBody<T = object> = T & {
 	stateHandle: IdxBaseResponse['stateHandle'];
 };
 
-const idxErrorSchema = z.object({
-	version: idxVersionSchema,
-	messages: z.object({
-		type: z.literal('array'),
-		value: z.array(
-			z.object({
-				message: z.string(),
-				i18n: z.object({ key: z.string() }).optional(),
-			}),
-		),
-	}),
+const idxErrorObjectSchema = z.object({
+	type: z.literal('array'),
+	value: z.array(
+		z.object({
+			message: z.string(),
+			i18n: z.object({ key: z.string() }).optional(),
+			class: z.literal('ERROR'),
+		}),
+	),
 });
+type IdxErrorObject = z.infer<typeof idxErrorObjectSchema>;
 
 export const idxFetch = async <ResponseType, BodyType>({
 	path,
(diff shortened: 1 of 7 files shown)


src/server/lib/okta/idx/shared.ts
```diff
@@ -0,0 +1,374 @@
...
+					error_description: e.message,
+				},
+				response.status,
+			);
+		});
+
+		// find the error message in the json
+		const error = findErrorMessage(json);
+
+		// if we found and error message
```
raphaelkabo, line 332:
```suggestion
		// if we found an error message
```

PR #2638, opened 2024-03-21
Remove app link for feast app

coldlink:
## What does this change?

The url interception currently isn't working too well for the feast app, so we're removing it from the site association file in order to make sure the app doesn't try to link to the app.

Instead users will see our fallback flow telling them to go back to the feast app and sign in after they register.

## Tested

Unfortunately we can't test this on CODE as Feast doesn't have a CODE environment, but this only changes the `/well-known/apple-app-site-association` file which doesn't affect any app specific code.


kelvin-chappell (approved):
👍

PR #2639, opened 2024-03-21
Passcodes | Set up passcodes for registration

coldlink:
## What does this change?

We've now upgraded to Okta Identity Engine. This unlocks a number of new features which previously weren't feasible. One of these is "passwordless", or the idea of eliminating, or reducing the need for passwords.

We also currently have an issue within our OAuth flows where users don't always end up signed in the same browser context that they registered from. This can occur if a user clicks a reset password/registration verification link in their on another browser or device, or if deeplinking/https interception isn't working properly on the user's device.

One way of fixing both these issues it to use One Time Passcodes (OTPs), where instead of a user clicking a link in an email, they have to type a 6 digit number into an input field which performs the same action of verifying the user. This means that the user always stays within the same browser device/context, while also securely verifying the user.

To begin with we're going to only implement OTPs for registration only, with the user still having to set a password, this is the focus of this PR. After this we will being rolling out passcodes for sign in and password reset too, thus eliminating passwords altogether, or at lease making them optional.

This PR builds on the work previously done in https://github.com/guardian/gateway/pull/2625 where the initial work to set up the IDX API was done, and https://github.com/guardian/gateway/pull/2637 where most of the APIs for passcodes were set up. The focus of this PR is to wire up the registration functionality to use the passcode API, and set up the frontend to handle this.

This PR however does not enable passcodes directly in `PROD`, the `usePasscodesRegistration` query parameter flag must be manually set in order to opt in, and the [switch](https://github.com/guardian/gateway/blob/3e221f52e6068ed8940212bf1cd502f44af7ee5d/src/shared/lib/featureSwitches.ts#L43) has to be enabled in `PROD`, it is currently set to `false`. This allows us to slowly opt in and first test on the CODE environment, followed by PROD behind the flag, followed by enabling it without the flag.

We're currently only enabling passcodes for new users who are registering. Existing users that are attempting to go through the registration flow will fallback to the legacy/current flow without passcodes for now until full passwordless is implemented. The full flowchart can be seen in https://github.com/guardian/gateway/issues/2567.

## The Journey

The steps to enable passcode registration are as follows:

1. User attempts to register with a new email, this calls `POST /register`
2. `POST /register`
	1. Checks if `registrationPasscodesEnabled` feature switch enabled and `usePasscodeRegistration` query parameter is set (this step is temporary before rolling out to all users) 
		1. If it is, attempt passcode registration flow
		2. Otherwise use classic flow
	2. Assuming 1) is stratified, first start the interaction code flow, and get the `interaction_handle` + `authState`
		1. See https://github.com/guardian/gateway/pull/2637 and https://github.com/guardian/gateway/pull/2625 for more details on specifics of what the interaction code flow/IDX API is doing
	3. Introspect the `interaction_handle` to get the `stateHandle`
	4. Encrypt any consents we need to preserve during the registration flow, and update the `authState` with the `stateToken` and `encryptedRegistrationConsents`
	5. Check if we have the `select-enroll-profile` remediation property which means registration is allowed using the IDX API
	6. Call the `/enroll` endpoint to attempt to start the registration process
	7. Call the `/enroll/new` endpoint to attempt to register the user with email
	8. If successful, the user will have been send an email with a passcode by Okta
	9. Set the encrypted state cookie to persist the `email` and `stateHandle` through the flow/between pages
	10. Finally redirect to the email sent page
	11. If at any point there is an error, we log the error and fallback to the classic registration flow
3. User shown the Email Sent page, but it has a passcode input field instead! User grabs the 6 digit passcode from their email. They type it in and click "Submit Passcode". This calls `POST /register/code`
4. `POST /register/code`
	1. Grab the `code` from the request body and read the `stateHandle` from the encrypted state cookie
	2. Check if the `stateHandle` is valid with the `introspect` endpoint
	3. Call the `/challenge/answer` endpoint with the `code` and `stateHandle`
		i. If valid continue
		ii. If invalid, we throw and handle a specific error case for invalid passcode: `api.authn.error.PASSCODE_INVALID`
	4. We need the user to set a password, so we need to enroll them in the "password" factor, find the `passwordAuthenticatorId`
	5. Call the `/credentials/enroll` method with the authenticator id and `methodType` set to `password`
	6. Redirect the user to the `/welcome/password` page where they'll set a password
5. `GET /welcome/password`
	1. This is intercepted as a `GET /welcome/:token` route, so we update the existing handler, which uses the `checkTokenInOkta` from `checkPasswordToken.ts`
	2. Checks if `registrationPasscodesEnabled` feature switch enabled and `usePasscodeRegistration` query parameter is set (this step is temporary before rolling out to all users) 
		1. If it is, attempt passcode registration flow
		2. Otherwise use classic flow, i.e assume the `token` parameter is a token
	3. If it is, read the encrypted state cookie for the email and state handle
	4. Check if the `stateHandle` is valid with the `introspect` endpoint and making sure it has the `enroll-authenticator` object so we know it's in the correct state
	5. Show the set password page
6. User enters a new password and submits, calls `POST /welcome/password`
	1. This is intercepted as a `POST /welcome/:token` route, so we update the existing handler, which uses the `changePasswordInOkta` from `changePassword.ts`
	2. Checks if `registrationPasscodesEnabled` feature switch enabled and `usePasscodeRegistration` query parameter is set (this step is temporary before rolling out to all users) 
		1. If it is, attempt passcode registration flow
		2. Otherwise use classic flow, i.e assume the `token` parameter is a token
	3. If it is, read the encrypted state cookie for the email and state handle
	4. Check if the `stateHandle` is valid with the `introspect` endpoint and making sure it has the `enroll-authenticator` object so we know it's in the correct state
	5. Validate the password field and check if breached
	6. Call the `setPasswordAndRedirect` function, which uses the `/challenge/answer` endpoint to set a password. This sets a global okta session, and completes the flow by redirecting the user to the interaction code callback endpoint, but with a `code` parameter (not an `interaction_code`) and the `state`.
7. In this callback endpoint, use a standard OAuth library that supports Authorization Code Flow with PKCE, and complete the flow. The user is authenticated at this point, and everything else remains the same as what we currently have
8. At any point if there is an error, we catch the error, log it, and fallback to the existing flow.

## Demo

https://github.com/guardian/gateway/assets/13315440/d49a4a24-b2e9-44c8-bfa4-8e668767216e

## Plan

- Merge this PR in behind feature flag and switch
- Perform testing on CODE apps and web
  - The query flag is `usePasscodeRegistration=true`
- Enable switch on PROD
  - PR required
- Perform testing on PROD behind 
- Remove feature flag on PROD
  - Big PR required 

While improving/making refinements.

## Tested?

- [x] Cypress tests
- [x] CODE - New user web
- [x] CODE - Existing user web
- [x] CODE - New user app
- [x] CODE - Existing user app

### DIFF
--- a/cypress/integration/ete-okta/registration_2.6.cy.ts
+++ b/cypress/integration/ete-okta/registration_2.6.cy.ts
@@ -748,7 +748,7 @@ describe('Registration flow - Split 2/2', () => {
 						expect(oktaUser.profile.registrationPlatform).to.eq('profile');
 					});
 
-					cy.url().should('contain', '/consents/newsletters');
+					cy.url().should('contain', '/welcome/review');
 				},
 			);
 		});
--- a/cypress/integration/ete-okta/registration_2.6.cy.ts
+++ b/cypress/integration/ete-okta/registration_2.6.cy.ts
@@ -683,4 +683,320 @@ describe('Registration flow - Split 2/2', () => {
 			});
 		});
 	});
+
+	// TODO: These tests should be merged into the existing registration tests when the passcode registration feature flag is removed
+	context('Passcode registration temporary tests', () => {
+		it('successfully registers using an email with no existing account using a passcode', () => {
+			const encodedReturnUrl =
+				'https%3A%2F%2Fm.code.dev-theguardian.com%2Ftravel%2F2019%2Fdec%2F18%2Ffood-culture-tour-bethlehem-palestine-east-jerusalem-photo-essay';
+			const unregisteredEmail = randomMailosaurEmail();
+			const encodedRef = 'https%3A%2F%2Fm.theguardian.com';
+			const refViewId = 'testRefViewId';
+			const clientId = 'jobs';
+
+			cy.visit(
+				`/register/email?returnUrl=${encodedReturnUrl}&ref=${encodedRef}&refViewId=${refViewId}&clientId=${clientId}&usePasscodeRegistration=true`,
+			);
+
+			const timeRequestWasMade = new Date();
+			cy.get('input[name=email]').type(unregisteredEmail);
+			cy.get('[data-cy="main-form-submit-button"]').click();
+
+			cy.contains('Enter your code');
+			cy.contains(unregisteredEmail);
+			cy.contains('send again');
+			cy.contains('try another address');
+
+			cy.checkForEmailAndGetDetails(unregisteredEmail, timeRequestWasMade).then(
+				({ body, codes }) => {
+					// email
+					expect(body).to.have.string('Your verification code');
+					expect(codes?.length).to.eq(1);
+					const code = codes?.[0].value;
+					expect(code).to.match(/^\d{6}$/);
+
+					// passcode page
+					cy.url().should('include', '/register/email-sent');
+					cy.get('input[name=code]').type(code!);
+
+					cy.get('form')
+						.should('have.attr', 'action')
+						.and('match', new RegExp(encodedReturnUrl))
+						.and('match', new RegExp(refViewId))
+						.and('match', new RegExp(encodedRef))
+						.and('match', new RegExp(clientId));
+
+					cy.contains('Submit passcode').click();
+
+					// password page
+					cy.url().should('include', '/welcome/password');
+					cy.get('form')
+						.should('have.attr', 'action')
+						.and('match', new RegExp(encodedReturnUrl))
+						.and('match', new RegExp(refViewId))
+						.and('match', new RegExp(encodedRef))
+						.and('match', new RegExp(clientId));
+
+					cy.get('input[name="firstName"]').type('First Name');
+					cy.get('input[name="secondName"]').type('Last Name');
+					cy.get('input[name="password"]').type(randomPassword());
+					cy.get('button[type="submit"]').click();
+
+					// test the registration platform is set correctly
+					cy.getTestOktaUser(unregisteredEmail).then((oktaUser) => {
+						expect(oktaUser.status).to.eq(Status.ACTIVE);
+						expect(oktaUser.profile.registrationPlatform).to.eq('profile');
+					});
+
+					cy.url().should('contain', '/consents/newsletters');
+				},
+			);
+		});
+
+		it('successfully registers using an email with no existing account using a passcode and redirects to fromURI', () => {
+			const appClientId = 'appClientId1';
+			const fromURI = '%2FfromURI1';
+
+			// Intercept the external redirect page.
+			// We just want to check that the redirect happens, not that the page loads.
+			cy.intercept(
+				'GET',
+				`https://${Cypress.env('BASE_URI')}${decodeURIComponent(fromURI)}`,
+				(req) => {
+					req.reply(200);
+				},
+			);
+
+			const encodedReturnUrl =
+				'https%3A%2F%2Fm.code.dev-theguardian.com%2Ftravel%2F2019%2Fdec%2F18%2Ffood-culture-tour-bethlehem-palestine-east-jerusalem-photo-essay';
+			const unregisteredEmail = randomMailosaurEmail();
+			const encodedRef = 'https%3A%2F%2Fm.theguardian.com';
+			const refViewId = 'testRefViewId';
+
+			cy.visit(
+				`/register/email?returnUrl=${encodedReturnUrl}&ref=${encodedRef}&refViewId=${refViewId}&appClientId=${appClientId}&fromURI=${fromURI}&usePasscodeRegistration=true`,
+			);
+
+			const timeRequestWasMade = new Date();
+			cy.get('input[name=email]').type(unregisteredEmail);
+			cy.get('[data-cy="main-form-submit-button"]').click();
+
+			cy.contains('Enter your code');
+			cy.contains(unregisteredEmail);
+			cy.contains('send again');
+			cy.contains('try another address');
+
+			cy.checkForEmailAndGetDetails(unregisteredEmail, timeRequestWasMade).then(
+				({ body, codes }) => {
+					// email
+					expect(body).to.have.string('Your verification code');
+					expect(codes?.length).to.eq(1);
+					const code = codes?.[0].value;
+					expect(code).to.match(/^\d{6}$/);
+
+					// passcode page
+					cy.url().should('include', '/register/email-sent');
+					cy.get('input[name=code]').type(code!);
+
+					cy.get('form')
+						.should('have.attr', 'action')
+						.and('match', new RegExp(encodedReturnUrl))
+						.and('match', new RegExp(refViewId))
+						.and('match', new RegExp(encodedRef))
+						.and('match', new RegExp(appClientId))
+						.and('match', new RegExp(fromURI));
+
+					cy.contains('Submit passcode').click();
+
+					// password page
+					cy.url().should('include', '/welcome/password');
+					cy.get('form')
+						.should('have.attr', 'action')
+						.and('match', new RegExp(encodedReturnUrl))
+						.and('match', new RegExp(refViewId))
+						.and('match', new RegExp(encodedRef))
+						.and('match', new RegExp(appClientId))
+						.and('match', new RegExp(fromURI));
+
+					cy.get('input[name="password"]').type(randomPassword());
+					cy.get('button[type="submit"]').click();
+
+					// test the registration platform is set correctly
+					cy.getTestOktaUser(unregisteredEmail).then((oktaUser) => {
+						expect(oktaUser.status).to.eq(Status.ACTIVE);
+						expect(oktaUser.profile.registrationPlatform).to.eq('profile');
+					});
+
+					cy.url().should('contain', decodeURIComponent(fromURI));
+				},
+			);
+		});
+
+		it('passcode incorrect functionality', () => {
+			const unregisteredEmail = randomMailosaurEmail();
+			cy.visit(`/register/email?usePasscodeRegistration=true`);
+
+			const timeRequestWasMade = new Date();
+			cy.get('input[name=email]').type(unregisteredEmail);
+			cy.get('[data-cy="main-form-submit-button"]').click();
+
+			cy.contains('Enter your code');
+			cy.contains(unregisteredEmail);
+			cy.contains('send again');
+			cy.contains('try another address');
+
+			cy.checkForEmailAndGetDetails(unregisteredEmail, timeRequestWasMade).then(
+				({ body, codes }) => {
+					// email
+					expect(body).to.have.string('Your verification code');
+					expect(codes?.length).to.eq(1);
+					const code = codes?.[0].value;
+					expect(code).to.match(/^\d{6}$/);
+
+					// passcode page
+					cy.url().should('include', '/register/email-sent');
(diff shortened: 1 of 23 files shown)


raphaelkabo:
Some questions:

- As a user, what's the experience if there's a token expiry while I'm on the passcode field? What happens if I wait for ages there, then put in the right passcode? What if it's the wrong passcode?
- Is there a situation where a token expiry can occur between the passcode input field and the password input field?
- For simplicity, I would just make `/welcome/password` its own `GET` and `POST` routes, and fail with an error if the feature flag and query parameter aren't present. We wouldn't be letting real users use the query parameter on production anyway, and it'll avoid overloading the `/welcome/:token` routes with logic that doesn't belong in them ('password' can never be a valid token).
-

src/server/lib/okta/idx/challenge.ts
```diff
@@ -9,6 +9,8 @@ import {
 } from './shared';
 import { selectAuthenticationEnrollSchema } from './enroll';
```
raphaelkabo, line 10: It's not from this PR, but these should be absolute imports.

src/server/lib/okta/idx/introspect.ts
```diff
@@ -5,6 +5,7 @@ import {
 	idxBaseResponseSchema,
 	idxFetch,
 } from './shared';
```
raphaelkabo, line 7: This should be an absolute import.

src/server/routes/register.ts
```diff
@@ -110,6 +133,262 @@ router.post(
...
+	const state = res.locals;
+
+	const encryptedState = readEncryptedStateCookie(req);
+
+	if (encryptedState && encryptedState.email && encryptedState.stateHandle) {
+		const html = renderer('/register/email-sent', {
+			requestState: mergeRequestState(state, {
+				pageData: {
+					email: readEncryptedStateCookie(req)?.email,
+					timeUntilTokenExpiry: encryptedState.stateHandleExpiresAt
```
raphaelkabo, line 147: You could be pedantic/avoid confusion and call this `msUntilTokenExpiry`. Just because some OAuth stuff uses seconds instead.

src/server/routes/register.ts
```diff
@@ -110,6 +133,262 @@ router.post(
...
+												}
+											});
+										}
+									});
+								}
+							}
+						})
+						.filter(
+							(id): id is string => typeof id === 'string' && id.length > 0,
+						);
```
raphaelkabo, line 235: This is terrifying. No suggestions to fix it, it just is. :(

src/server/routes/register.ts
```diff
@@ -110,6 +133,262 @@ router.post(
...
+
+				if (!passwordAuthenticatorId.length) {
```
raphaelkabo, line 237: I'm surprised Typescript can promise that `passwordAuthenticatorId` is set to anything after all those nested `flatMap`s and `if`s! Maybe `passwordAuthenticatorId?.length` is safer anyway?

src/server/routes/register.ts
```diff
@@ -110,6 +133,262 @@ router.post(
...
+							requestState: mergeRequestState(res.locals, {
+								queryParams: {
+									returnUrl: res.locals.queryParams.returnUrl,
+									emailSentSuccess: false,
+								},
+								pageData: {
+									email: readEncryptedStateCookie(req)?.email,
+									timeUntilTokenExpiry: encryptedState.stateHandleExpiresAt
+										? new Date(encryptedState.stateHandleExpiresAt).getTime() -
+											Date.now()
```
raphaelkabo, line 273: I wonder if this could be a helper/lib method.

src/client/pages/EmailSent.tsx
```diff
@@ -40,49 +54,144 @@ export const EmailSent = ({
...
-				</MainBodyText>
-			) : (
-				<MainBodyText>
-					Please follow the instructions in this email.
-				</MainBodyText>
+			{hasStateHandle && (
+				<>
+					{email ? (
+						<MainBodyText>
+							We’ve sent an email to <b>{email}</b> with verification
```
raphaelkabo, line 138: This is odd wording ('verification instructions and a verification code') - surely just 'with a verification code' is fine?

src/client/pages/EmailSent.tsx
```diff
@@ -40,49 +54,144 @@ export const EmailSent = ({
...
+							instructions and a verification code.
+						</MainBodyText>
+					) : (
+						<MainBodyText>
+							We’ve sent you an email with verification instructions and a
+							verification code.
```
raphaelkabo, line 144: Similar odd wording here.

src/server/lib/okta/idx/challenge.ts
```diff
@@ -82,13 +106,31 @@ export const setPasswordAndRedirect = async (
...
+		await validateEmailAndPasswordSetSecurely(id);
+	} else {
+		logger.error(
+			'Failed to set validation flags in Okta as there was no id',
+			undefined,
+			{
+				request_id,
+			},
+		);
+	}
```
raphaelkabo, line 132: Does anything rely on this? If the email isn't validated successfully, what happens to the short-lived user session next? What happens to any new user sessions? (Basically, are we happy leaving the error only logged?)

src/server/routes/register.ts
raphaelkabo: A further thought: is there any way at all that we could leverage zod to handle this complex nested object? It feels like a maintenance nightmare otherwise.

cypress/integration/ete-okta/registration_2.6.cy.ts
```diff
@@ -683,4 +683,329 @@ describe('Registration flow - Split 2/2', () => {
 			});
 		});
 	});
+
```
raphaelkabo, line 686: I'll be honest, I didn't pick through these tests with a fine-tooth comb. I'm hoping that any problems will be self-revealing!

cypress/integration/ete-okta/registration_2.6.cy.ts
```diff
@@ -683,4 +683,329 @@ describe('Registration flow - Split 2/2', () => {
...
+			// Test users created via IDAPI-with-Okta do not have the activation
+			// lifecycle run at creation, so they don't transition immediately from
+			// STAGED to PROVISIONED (c.f.
+			// https://developer.okta.com/docs/reference/api/users/#create-user) .
+			// This is useful for us as we can test STAGED users first, then test
+			// PROVISIONED users in the next test by activating a STAGED user. Users
+			// created through Gateway-with-Okta do have this lifecycle run, so if we
+			// rebuild these tests to not use IDAPI at all, we need to figure out a
+			// way to test STAGED and PROVISIONED users (probably by just passing an
+			// optional `activate` prop to a createUser function).
```
raphaelkabo, line 962: This comment doesn't really make sense here (I think it's copied from elsewhere).

raphaelkabo (commented):
I'm very happy (and impressed) with this! I'm not approving it since it needs rebase with `main` anyway, and I've left some comments, but generally speaking I think it's very much ready for a CODE deploy behind a flag.

coldlink:
> Some questions/notes (I think I've answered the first two of these for myself by reading the code but I wanted to double check because they're edge cases):
> 
> * As a user, what's the experience if there's a token expiry while I'm on the passcode field? What happens if I wait for ages there, then put in the right passcode? What if it's the wrong passcode?
> * Is there a situation where a token expiry can occur between the passcode input field and the password input field?
> * For simplicity, I would just make `/welcome/password` its own `GET` and `POST` routes, and fail with an error if the feature flag and query parameter aren't present. We wouldn't be letting real users use the query parameter on production anyway, and it'll avoid overloading the `/welcome/:token` routes with logic that doesn't belong in them ('password' can never be a valid token).

- As a user, what's the experience if there's a token expiry while I'm on the passcode field? What happens if I wait for ages there, then put in the right passcode? What if it's the wrong passcode?
  - Everything should be handled correctly, if there is a passcode expiry at any point this would be caught and the user shown a "link expired" page, thanks to the `timeUntilTokenExpiry` code
  - This also works because we introspect the `stateHandle` on every request to make sure it's still valid, regardless of

src/server/lib/okta/idx/challenge.ts
coldlink: I'll make a separate PR to update any absolute imports!

src/server/lib/okta/idx/introspect.ts
coldlink: I'll make a separate PR to update any absolute imports!

src/server/routes/register.ts
coldlink: I did it this way as `timeUntilTokenExpiry` was already a thing in Gateway, and wanted to avoid a refactor.

src/server/routes/register.ts
coldlink: this is leveraging Zod as best I can 😱

it's just annoying at there are so many states to what the IDX API returns, it's not consistent, so we always have to double check everything.

src/server/routes/register.ts
coldlink: `if (!passwordAuthenticatorId.length) {` is effectively the same as doing `if (passwordAuthenticatorId?.length) {` in this scenario?

src/server/routes/register.ts
coldlink: converted this to a `convertExpiresAtToExpiryTimeInMs` method

src/client/pages/EmailSent.tsx
coldlink: I'll speak to @nickhartyyy about this

src/client/pages/EmailSent.tsx
coldlink: and this

src/server/lib/okta/idx/challenge.ts
coldlink: one of the cases where it should be theoretically impossible to get to, but helpful if we log an error message

on next sign in they'd be sent the "security email" if they were ever to end up in this scenario

cypress/integration/ete-okta/registration_2.6.cy.ts
coldlink: Good spot, I'll update this!

src/server/routes/register.ts
coldlink: hopefully the new function name makes it clearer

src/client/pages/EmailSent.tsx
coldlink: also thinking about this, if we get this merged in, it's easier to get opinions and feedback on wording/structure which is easier to change afterwards anyway

raphaelkabo:
> * For simplicity, I would just make `/welcome/password` its own `GET` and `POST` routes, and fail with an error if the feature flag and query parameter aren't present. We wouldn't be letting real users use the query parameter on production anyway, and it'll avoid overloading the `/welcome/:token` routes with logic that doesn't belong in them ('password' can never be a valid token).
>   
>   * I did it this way for two reasons.
>   * 1. To make it easier to migrate to passcodes in the future without having to rewrite any routes, and to reuse as much code as possible
>   * 2. To make it easier to fallback to the existing code if for some reason something goes wrong

*grumble grumble approved*

raphaelkabo (approved):

PR #2652, opened 2024-03-26
Dependencies | 26th March 2024

coldlink:
## What does this change?

Changes from previous https://github.com/guardian/gateway/pull/2623

- Finally update `sharp` thanks to #2626 

Holds back the following:

- Storybook 8 - While #2626 unblocks this, the upgrade path isn't so easy so will be done in a separate PR at some point

## Tested

- [x] CODE


kelvin-chappell (approved):
👍

PR #2654, opened 2024-03-26
Dependencies | Update to storybook 8

coldlink:
## What does this change?

Updates Gateway to use Storybook 8.

Migration guide: https://storybook.js.org/docs/migration-guide
Full migration guide: https://github.com/storybookjs/storybook/blob/next/MIGRATION.md#from-version-7x-to-800

- Attempted automatic upgrade using `pnpm dlx storybook@latest upgrade`.
  - Removed the `argTypesRegex` in `.storybook/preview.js`, wasn't used, and now removed in storybook anyway
  - Changed `'../src/**/*.stories.mdx'` to `'../src/**/*.mdx'` in `.storybook/main.js`
  - Added `@storybook/addon-webpack5-compiler-babel`
- Manually updated the deprecated `@storybook/addon-styling` to `@storybook/addon-styling-webpack` and `@storybook/addon-themes`
  - Followed the migration guide https://github.com/storybookjs/addon-styling/blob/main/MIGRATION.md
- Manually removed the `babel` option from `.storybook/main.js` and added the preset directly to `babel.config.js`
  - `babel` is only used by storybook
  - The preset, `@emotion/babel-preset-css-prop`, fixes styling issues in storybook when using Emotion 11
  - This also prevents us using `swc` which is also available now in Storybook (`@storybook/addon-webpack5-compiler-swc`), but isn't as big of a deal when using storybook as compared to development
- Change `import { ComponentMeta, ComponentStory } from '@storybook/react';` to `import { Meta, StoryFn } from '@storybook/react';`
  - https://github.com/storybookjs/storybook/blob/next/MIGRATION.md#decoratorfn-story-componentstory-componentstoryobj-componentstoryfn-and-componentmeta-typescript-types

PR #2668, opened 2024-04-02
Dependencies | 2nd April 2024

coldlink:
## What does this change?

Changes from previous #2652 & #2654

Holds back the following:

- *NOTHING* 🤯 

## Tested

- [x] CODE


src/server/lib/__tests__/headers.test.ts
```diff
@@ -28,7 +28,7 @@ jest.mock('@/server/lib/serverSideLogger', () => ({
 	},
 }));
 jest.mock('@aws-sdk/credential-providers');
-jest.mock('@aws-sdk/node-http-handler');
```
coldlink, line 31: `@aws-sdk/node-http-handler` has been deprecated, and instead told us to use `@smithy/node-http-handler` which we had installed already.

raphaelkabo (approved):
🚀

PR #2669, opened 2024-04-02
Fix | Native app registration | Correctly get `isNativeApp` and `appName`

coldlink:
## What does this change?

We noticed today that when registering on any native app, when the user went to set their password, they would be shown the CMP banner, and the onboarding flow navigation bar.

This is because the `isNativeApp` and `appName` parameters were missing in this scenario.

This happened because the `checkTokenInOkta` function would override the value of these parameters when seeing if a prefix was present.

When a prefix is not present, but an `appClientId` query parameter is, then we should derive the app information from this. However the logic in the `checkTokenInOkta` would overwrite this with undefined!

We've fixed this by explicit checking if the `isNativeApp` and `appName` exist on the request state (where they'll have been derived from the `appClientId`, and then if it does not, then try using the prefix, before setting to `undefined`.

This will improve the UX for users registering from within the app!

I also defined a variable called `state` set to `res.locals` which makes the function easier to read, especially where the request state is involved.

<table>
<tr>
 <th> Before
 <th> After
<tr>
 <td>

profile thegulocal com_welcome_TUbtYanF83WtiTUUz6bwtA jz5qn5APsnlgGmWM JKoHWDrY8gGZAT8EU_xT_Uj0ImQ9f2Xb7lnkxh1f1eN1jkVwFxnWOW8mJPiVP87-OslpnPEMho-IkpX765BF_ref=https%3A%2F%2Fprofile thegulocal com%2Fregister%2Femail refViewId=luiewg3vp9sl12

 <td>

Screenshot_20240402_141958

</table>

## Tested
- [x] DEV
- [x] CODE

### DIFF
--- a/src/server/controllers/checkPasswordToken.ts
+++ b/src/server/controllers/checkPasswordToken.ts
@@ -20,12 +20,12 @@ import { getConfiguration } from '@/server/lib/getConfiguration';
 import { validateRecoveryToken as validateTokenInOkta } from '@/server/lib/okta/api/authentication';
 import { trackMetric } from '@/server/lib/trackMetric';
 import { ChangePasswordErrors } from '@/shared/model/Errors';
-import { FieldError } from '@/shared/model/ClientState';
+import { FieldError, IsNativeApp } from '@/shared/model/ClientState';
 import { PersistableQueryParams } from '@/shared/model/QueryParams';
 import { validateReturnUrl } from '@/server/lib/validateUrl';
 import { mergeRequestState } from '@/server/lib/requestState';
 import { decryptOktaRecoveryToken } from '@/server/lib/deeplink/oktaRecoveryToken';
-import { getAppName, getAppPrefix } from '@/shared/lib/appNameUtils';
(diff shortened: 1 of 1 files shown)


src/server/controllers/checkPasswordToken.ts
```diff
@@ -205,24 +206,38 @@ export const checkTokenInOkta = async (
...
+					: 'ios'
+				: undefined;
+
+		// attempt to determine the app name
+		// first checking the pageData, then the prefix, otherwise undefined
+		const appName = state.pageData.appName
+			? state.pageData.appName
+			: prefix
+				? getAppName(prefix)
+				: undefined;
```
raphaelkabo, line 225: This could be...

```suggestion
		const appName = state.pageData.appName ||
		    prefix
	                ? getAppName(prefix)
			: undefined;
```

raphaelkabo (approved):
👍

src/server/controllers/checkPasswordToken.ts
coldlink: Used IIFE instead to make it more readable and explicit.

PR #2670, opened 2024-04-03
Gateway | New designs

raphaelkabo:
## What does this change?

This PR implements a full redesign for Gateway based on new design language developed to better align Gateway with the Guardian Live apps and other redesigns currently being developed at the Guardian.

To view all the design changes, it's best to look at the latest Chromatic build in the reviews and checks box below.

Along with the new designs themselves, we've made some QoL improvements:

- Added a `<Theme />` component which sets CSS variables for light and dark mode colour schemes. These are then implemented throughout the Gateway codebase for consistency.
- Updated our Storybook and Chromatic setup to allow us to run Chromatic tests directly in local Storybook, and test and display light/dark mode theme changes from the Storybook UI. Chromatic will now also automatically test each Storybook story in four modes: light mobile, light desktop, dark mobile, and dark desktop.
- Migrated away from specifying margins for various elements, relying instead on flexbox and the `gap` property to automatically space out items.
- Transitioned to using `remSpace` instead of `space` almost everywhere, to avoid errors creeping in when a number is specified using `space` but we forgot to add `px` after it (`remSpace` automatically adds `rem` to the number; `space`, being just a number, is more useful for JS calculations, which is where we're still using it).
- Simplified our layout widths to two in `src/client/models/Style.ts`: 460px and 980px.
- Improved the logic of the error/success messaging for password fields.


src/client/components/MinimalHeader.tsx
```diff
@@ -0,0 +1,33 @@
...
+`;
+
+const logoStyles = css`
+	height: 2rem;
+	display: flex;
+	svg {
+		fill: var(--color-logo);
+		height: 100%;
+	}
+	${from.desktop} {
```
coldlink, line 20: Desktop here is `980px`, wheres in `@/client/model/Style` it's defined as 940px. This could be causing issues in some of our diffs.

src/client/components/NameInputField.tsx
```diff
@@ -83,6 +84,7 @@ const SecondNameInput = (props: NameInputProps) => {
 			onInvalid={onInvalid}
 			error={errorMessage}
 			defaultValue={props.defaultValue}
+			theme={textInputTheme}
```
coldlink, line 87: I notice this pattern is everywhere.. would it be worth making a new component called `ThemedTextInput` which automatically applies the theme to the source `TextInput` component. And then use the `ThemedTextInput` everywhere the `TextInput` was being used.

This would mean that we wouldn't have to worry about future developers failing to add the `theme` prop everywhere.

raphaelkabo: Yeah, that makes sense! Like for the links.

src/client/components/RegistrationMarketingConsentFormField.tsx
```diff
@@ -1,40 +1,20 @@
...
-`;
-
 type Props = {
 	id: string;
-	label: string;
+	description: string;
 };
 
-export const RegistrationMarketingConsentFormField = ({ id, label }: Props) => {
+export const RegistrationMarketingConsentFormField = ({
```
coldlink, line 9: The `RegistrationMarketingConsentFormField` and `RegistrationNewsletterFormField` are looking basically like the same component with some slightly different props. Could they be combined somehow?

You can always use type narrowing so that depending on the combination of props you provide it will narrow to a specific type.

So it would look something close to this in a single component?

```ts
type MarketingProps = {
	id: string;
	description: string;
  title: undefined;
  imagePath: undefined;
  defaultChecked?: boolean;
};

type NewsletterProps = {
	id: string;
	title: string;
	description: string;
	imagePath?: string;
	defaultChecked?: boolean;
};

type Props = MarketingProps | NewsletterProps

export const RegistrationConsentFormField = ({
  description,
  id,
  title,
  imagePath,
  defaultChecked,
}: Props) => {
  return (
		<ToggleSwitchInput
			id={id}
			title={title}
			defaultChecked={defaultChecked}
			description={description}
			imagePath={imagePath}
		/>
	);
}
```

coldlink (commented):
still reviewing, half way through in number of files, so heres a few more comments

coldlink (commented):
still reviewing, half way through in number of files, so heres a few more comments

src/client/layouts/MinimalLayout.stories.tsx
```diff
@@ -0,0 +1,44 @@
+import React from 'react';
```
coldlink, line 1: The previous `Main` layout had a number of stories detailing all the different ways you could use this component. It's probably worth adding those back in, so it's easier to see how the new `MininalLayout` can be composed.

src/client/layouts/MinimalLayout.tsx
```diff
@@ -0,0 +1,108 @@
...
+import { MinimalLayoutImage } from '@/client/components/MinimalLayoutImage';
+import { LAYOUT_WIDTH_NARROW, LAYOUT_WIDTH_WIDE } from '@/client/models/Style';
+import { MainBodyText } from '@/client/components/MainBodyText';
+import { errorMessageStyles } from '@/client/styles/Shared';
+
+export interface MinimalLayoutProps {
+	children: React.ReactNode;
+	wide?: boolean;
+	pageHeader: string;
+	leadText?: string | React.ReactNode;
```
coldlink, line 26:
```suggestion
	leadText?: React.ReactNode;
```

ReactNodes can already be of type string amongst others!

https://github.com/DefinitelyTyped/DefinitelyTyped/blob/9f855c408dac3c7b3bf0ed9d78242ce073c7aaf1/types/react/index.d.ts#L478

src/client/pages/DeleteAccountBlocked.tsx
```diff
@@ -18,18 +18,22 @@ export const DeleteAccountBlocked = ({ contentAccess = {} }: Props) => {
...
+			pageHeader="Delete your Guardian account"
+			leadText={
+				<MainBodyText>
+					We&apos;ve noticed you have an active membership, recurring
+					contribution or subscription with the Guardian, so you won&apos;t be
+					able to automatically delete your account online. Please follow the
+					instructions below first and try again if you would still like delete
+					your account.
+				</MainBodyText>
+			}
```
coldlink, line 31: I think this should be separated into two paragraphs, as it becomes unwieldy to read?

Something close to this?

Again, it was like this previously, but could be an ideal time to change it.

```suggestion
			leadText={
			  </>
				<MainBodyText>
					We&apos;ve noticed you have an active membership, recurring
					contribution or subscription with the Guardian, so you won&apos;t be
					able to automatically delete your account online.
				</MainBodyText>
				<MainBodyText>
				Please follow the
					instructions below first and try again if you would still like delete
					your account.
					</MainBodyText>
				</>
			}
```

src/client/pages/EmailSent.tsx
```diff
@@ -168,28 +167,15 @@ export const EmailSent = ({
...
-							defaultValue={passcode}
-						/>
+						<PasscodeInput passcode={passcode} fieldErrors={fieldErrors} />
 					</MainForm>
 				</div>
 			)}
 			<InformationBox>
 				<InformationBoxText>
-					Didn’t get the email? Check your spam
+					Didn’t get the email? Check your spam&#8288;
```
coldlink, line 176: ooh so this is a word joiner?

https://www.codetable.net/decimal/8288

src/client/pages/MaintenancePage.tsx
```diff
@@ -1,12 +1,12 @@
...
 import { MainBodyText } from '@/client/components/MainBodyText';
+import { MinimalLayout } from '@/client/layouts/MinimalLayout';
 
 export const MaintenancePage = () => (
-	<MainLayout pageHeader="We’ll be back soon">
+	<MinimalLayout pageHeader="We’ll be back soon">
 		<MainBodyText>
 			Sorry for the inconvenience. We are currently performing some essential
 			maintenance. Please try again later.
 		</MainBodyText>
```
coldlink, line 10: Other places are using the new `leadText` prop for similar things?

Should this use `leadText`?

src/client/pages/NotFoundPage.tsx
```diff
@@ -1,15 +1,15 @@
...
+import { MinimalLayout } from '@/client/layouts/MinimalLayout';
 
 export const NotFoundPage = () => (
-	<MainLayout pageHeader="Sorry – the page does not exist">
+	<MinimalLayout pageHeader="Sorry – this page does not exist">
 		<MainBodyText>
 			You may have followed an outdated link, or have mistyped a URL. If you
 			believe this to be an error, please{' '}
 			<ExternalLink href={locations.REPORT_ISSUE}>report it</ExternalLink>.
 		</MainBodyText>
```
coldlink, line 13: should this also become leadText?

src/client/pages/ResendConsentEmail.tsx
```diff
@@ -12,7 +12,7 @@ interface Props {
 
 export const ResendConsentEmail = ({ token, queryParams }: Props) => {
 	return (
-		<MainLayout pageHeader="Link expired">
+		<MinimalLayout pageHeader="Link expired">
 			<MainBodyText>This link has expired.</MainBodyText>
 			<MainBodyText>To receive a new link, please click below.</MainBodyText>
```
coldlink, line 17: leadText?

.storybook/preview.js
```diff
@@ -1,40 +1,48 @@
...
+} from '@storybook/addon-themes';
 import { Global, css } from '@emotion/react';
 import { FocusStyleManager } from '@guardian/source/foundations';
 import { fontFaces } from '@/client/lib/fonts';
-import { Breakpoints } from '@/client/models/Style';
 import clientStateDecorator from './clientStateDecorator';
 import { neutral } from '@guardian/source/foundations';
+import { Theme } from '@/client/styles/Theme';
+import { LAYOUT_WIDTH, LAYOUT_WIDTH_WIDE } from '@/client/models/Style';
+import { chromaticModes } from './modes';
```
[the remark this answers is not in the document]
raphaelkabo: Yeah, tbf Storybook docs suggest `modes` should be a `.js` file.

src/client/pages/ResendConsentEmail.tsx
raphaelkabo: It looked weird to put both lines in `leadText` and felt wrong in the code to put only one of them in, but maybe yeah, the first line only.

src/client/pages/SignIn.stories.tsx
```diff
@@ -114,3 +106,10 @@ export const WithJobsAndSocialSigninBlocked = (args: SignInProps) => (
 WithJobsAndSocialSigninBlocked.story = {
 	name: 'with Jobs terms and social sign-in blocked',
 };
+
+export const isReauthenticate = (args: SignInProps) => (
```
coldlink, line 110: Stories should use PascalCase.

Just saw this:
Screenshot 2024-06-18 at 09 42 51

src/client/components/PasswordInput.tsx
```diff
@@ -124,12 +110,57 @@ const EyeSymbol = ({
...
+	input {
+		margin-top: 0;
+	}
+	// FOCUS LOGIC
+	// Modern browsers which support :has
+	:has(:focus) {
+		border: 2px solid var(--color-input-text);
+		padding: 0;
+	}
+	// React-based fallback for browsers which don't support :has
```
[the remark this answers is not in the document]
raphaelkabo: But ew, CSS comments are gross. 😛

src/client/pages/SubscriptionSuccess.tsx
```diff
@@ -19,7 +19,9 @@ export const SubscriptionSuccess = ({
 	accountManagementUrl = 'https://manage.theguardian.com',
 }: UnsubscribeSuccessProps) => {
 	return (
-		<MainLayout pageHeader={`${subscriptionActionName(action)} Confirmation`}>
+		<MinimalLayout
+			pageHeader={`${subscriptionActionName(action)} Confirmation`}
+		>
 			<MainBodyText>
```
coldlink, line 25: Could this first paragrah become leadText?

src/client/components/PasswordInput.tsx
```diff
@@ -139,49 +170,59 @@ export const PasswordInput = ({
...
+				</InlineError>
+			)}
+			{supporting && <p css={supportingTextStyles}>{supporting}</p>}
+			<div css={wrapperStyles(hasFocus)}>
+				<div
+					css={[
+						css`
+							flex: 1;
+							align-self: flex-end;
+						`,
```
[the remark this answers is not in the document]
raphaelkabo: I think it's completely unnecessary actually, it's a div with one child element and `flex: 1`.

src/client/components/PasswordInput.tsx
```diff
@@ -139,49 +170,59 @@ export const PasswordInput = ({
...
 					]}
-				/>
-			</div>
+				>
+					<ThemedTextInput
+						onChange={onChange}
+						onFocus={() => setFieldIsFocused(true)}
+						onBlur={() => setFieldIsFocused(false)}
+						label={''}
```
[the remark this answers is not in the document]
raphaelkabo: Nothing we can do here - even if the label prop isn't passed  at all to `<TextInput>`, it still creates an empty label element. The label above has `htmlFor` set to the appropriate input, so in terms of accessibility, the input has two labels, one of them is blank and doesn't have a textLeaf so hopefully screen readers will just ignore it. It's not fantastic, but we'd have to change Source to make label optional to fix this.

src/client/styles/Shared.ts
```diff
@@ -12,30 +12,79 @@ export const disableAutofillBackground = css`
...
-		flex: 1 1;
-		border-bottom: 1px solid ${border.secondary};
-		margin: 8px;
-	}
-`;
-
-// fixed source divider expanding off the side of the container
 export const divider = css`
-	margin-left: 0;
+	// fixed source divider expanding off the side of the container
```
coldlink, line 16:
```suggestion
	/* fixed source divider expanding off the side of the container */
```

src/client/layouts/MinimalLayout.tsx
raphaelkabo: TIL!

src/client/pages/EmailSent.tsx
raphaelkabo: Yes, it's mildly cursed, welcome to 2024 we still need to do weird Unicode tricks

coldlink (commented):
Finished!

src/client/components/RegistrationMarketingConsentFormField.tsx
raphaelkabo: Just realised we don't even need this component at all - the wrapper doesn't add anything useful. Removing it entirely.

src/client/components/ToggleSwitchInput.tsx
```diff
@@ -32,16 +30,32 @@ const labelStyles = css`
...
+	border: 0;
+	margin: 0;
+	border-radius: 4px;
+	padding: ${remSpace[2]};
+	border: ${switchVariables.border}px solid
+		var(--color-toggle-inactive-background);
+	display: grid;
+	grid-template-columns: calc(100% - ${switchComputedWidth}px) ${switchComputedWidth}px;
+
+	&:has(input:focus) {
```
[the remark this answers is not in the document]
raphaelkabo: Yes, great catch!

src/client/layouts/MinimalLayout.stories.tsx
raphaelkabo: Good idea.

src/client/lib/hooks/useRecaptcha.tsx
```diff
@@ -207,7 +208,9 @@ export const RecaptchaWrapper: React.FC<{
 		token,
 		widgetId,
 	]);
-	return <RecaptchaElement id="recaptcha" />;
+	return (
+		<RecaptchaElement id="recaptcha" css={{ marginTop: `-${remSpace[3]}` }} />
```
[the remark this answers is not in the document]
raphaelkabo: I can't now, but I feel like maybe it was necessary one time that reCAPTCHA was visible? Let's delete it and we can always add it back in if necessary.

src/client/pages/SignIn.stories.tsx
raphaelkabo: Weird that this slipped through!

coldlink (approved):
Absolutely audaciously arduous affair.

Approvingly approved.

PR #2695, opened 2024-04-19
Fix Cypress tests to work with new CMP banner

raphaelkabo:
## What does this change?

Recent changes to the CMP banner (namely the addition of a 'No thank you' reject all button) have broken some of our Cypress tests. This updates the tests to work with the new banner layout.


kelvin-chappell (approved):
👍

PR #2709, opened 2024-04-29
Recaptcha | Always show detailed recaptcha message and update copy

coldlink:
## What does this change?

Always show the detailed recaptcha message on a recaptcha related error. This gives additional context to help the user why they are in this scenario. We also update the message to include "Temporarily disable VPNs and content blockers" which we find can block recaptcha.

Since we were previously showing this error message when this error occurred more than once on the same page, we've now changed this to always show the detailed message.

localhost_6006_iframe html_args= globals=viewport_MOBILE id=pages-signin--invalid-recaptcha viewMode=story(iPhone 12 Pro)

## Tested

- [x] CODE


raphaelkabo (approved):
Screenshot 2024-04-29 at 15 12 42

we love to see it ✅

PR #2708, opened 2024-04-29
Dependencies | 29th April 2024

coldlink:
## What does this change?

Changes from https://github.com/guardian/gateway/pull/2706:

- Brings our dependencies up to the latest versions where possible.

Holds back the following:

- Eslint to v9 - needs a configuration change

## Tested

- [x] CODE

PR #2721, opened 2024-05-07
Dependencies | 7th May 2024

coldlink:
## What does this change?

Changes from https://github.com/guardian/gateway/pull/2708:

- Updates Cypress CMP commands to use selectors instead of wording.
  - Recent change on the CMP has caused the wording to change, safer targeting the selectors

Holds back the following:

- Eslint to v9 - needs a configuration change

## Tested

- [x] CODE


raphaelkabo (approved):
🚀

PR #2722, opened 2024-05-08
OAuth | Fix authentication handler delay issues

raphaelkabo:
## What does this change?

- Moves the checks for `legacyIdentityId` introduced in https://github.com/guardian/gateway/pull/2694/ from the main OAuth callback endpoint to the authentication handler, so they only run during authentication activities
- Adds a function to attempt to fix an Okta profile missing a `legacyIdentityId` field by fetching the correct ID from IDAPI.
- Changes the check function to instead attempt to run this fix, and if that succeeds, return fresh tokens which will be used in another invocation of the `authenticationHandler()` function. If the fix or token fetching fails, the user is sent back to `/reauthenticate` with a hopefully useful error, as before.
- The timeout for the fix function is reduced from 30 to 10 seconds, which is under the 30 seconds specified as the TTFB timeout in our Fastly config (this may have been causing users to see 503 errors).
- Adds a Cypress test to check that this fix function works.

Thank you @coldlink for pairing on this with me for 2 hours!

### Flowchart

```mermaid
flowchart TD
    start(["START: Receive a token set"])
    start --> hasId{Does token set\ncontain legacy\nIdentity ID?}
    hasId -- Yes --> returnTokenSet([Return valid token set!])
    hasId -- No --> timedOut{Have we reached\nmaximum wait time?}
    timedOut -- Yes --> returnNull([Return null])
    timedOut -- No --> attemptFix[Attempt to fix Okta\nuser profile]
    attemptFix --> hasFixSucceeded{Did the fix\nsucceed?}
    hasFixSucceeded -- Yes --> refreshTokens[Get new tokens using\nrefresh token]
    refreshTokens -- Pass new tokens --> start
    hasFixSucceeded -- "No - pass\nold tokens" --> start
```

## Tests

- [x] Tested on CODE

### DIFF
--- a/src/server/routes/oauth.ts
+++ b/src/server/routes/oauth.ts
@@ -45,6 +45,7 @@ import { RoutePaths } from '@/shared/model/Routes';
 import { sendGuardianLiveOfferEmail } from '@/email/templates/GuardianLiveOffer/sendGuardianLiveOfferEmail';
 import { sendMyGuardianOfferEmail } from '@/email/templates/MyGuardianOffer/sendMyGuardianOfferEmail';
 import { emailSendMetric } from '@/server/models/Metrics';
+import { fixOktaProfile } from '@/server/lib/okta/fixProfile';
 
 const { baseUri, deleteAccountStepFunction } = getConfiguration();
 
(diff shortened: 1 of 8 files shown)


src/server/routes/oauth.ts
```diff
@@ -85,52 +86,111 @@ const redirectForGenericError = (_: Request, res: ResponseWithRequestState) => {
...
+				waitTime,
+				refreshedTokenSet,
+				newRefreshToken,
 			);
-		}, waitTime);
-	});
+		}
+
+		// Try again but using the existing refresh token
+		return nonBlockingWait(
```
pvighi, line 180: I'm finding it hard to follow this with all the early returns. In what case does execution reach this ?
If the token already has an id then it returns the token so it doesn't get here
if we exceeded the wait time it returns null and we don't get here
if the profile is fixed then do the `nonBlockingWait` on line 170 with the refreshed tokenset 

so is this for the case where the token didn't have the id but and we had an error trying to fetch the id from idapi so we hope that it was a temporary issue and it would work next time ?

It would be good to have a flowchart or something like that somewhere to make sure we don't forget how this works

src/server/routes/oauth.ts
```diff
@@ -174,6 +235,73 @@ const authenticationHandler = async (
...
+					oktaId: sub,
+					email,
+					openIdClient,
+					tokenSet,
+					refreshToken,
+					ip: req.ip,
+					request_id: res.locals.requestId,
+				});
+				if (!validTokenSet) {
+					throw new Error('Failed to get a valid token set');
```
pvighi, line 270: why do we throw an exception in this case but not in the earlier case when there's no `legacy_identity_id` or `refreshToken` ?

src/server/routes/oauth.ts
```diff
@@ -174,6 +235,73 @@ const authenticationHandler = async (
...
 
+		const { user_groups, email_validated, sub, email } =
+			tokenSet.claims() as CustomClaims;
+
+		// We want to ensure that the user has a legacyIdentityId set in their
+		// Okta profile before carrying on. This is surfaced via the legacy_identity_id
+		// claim in the access token.
+		const refreshToken = tokenSet.refresh_token;
+		const { legacy_identity_id } = tokenSet.claims();
+		if (!refreshToken && !legacy_identity_id) {
```
pvighi, line 246: why doesn't this trigger the an error response similar to when the refresh token is there but we cannot fetch a valid token set ?

raphaelkabo: Yeah, I think you're right - this is such a weird edge case that really no user should end up here! So we'll error here.

src/server/routes/oauth.ts
raphaelkabo: Excellent idea, I've added a flowchart to the PR description!

src/server/routes/oauth.ts
```diff
@@ -148,7 +209,8 @@ const authenticationHandler = async (
 	res: ResponseWithRequestState,
 	tokenSet: TokenSet,
 	authState: AuthorizationState,
-) => {
+	openIdClient: OpenIdClient,
+): Promise<void> => {
```
pvighi, line 213: This is not a problem introduced in this PR but as I think this function is too long. 
We should consider if there's opportunities to break off self contained parts of it into separate functions with descriptive names.
Obviously I'm not saying we should do it in this PR as it would delay the merging of this and it would add a lot of noise to the diff

src/shared/model/Routes.ts
```diff
@@ -100,6 +100,7 @@ export type ApiRoutePaths =
 	| '/user/change-email'
 	| '/unsubscribe'
 	| '/subscribe'
+	| '/user'
```
pvighi, line 103: what is this for ?

pvighi (dismissed):
I left a couple of questions but overall it looks good to me

src/server/routes/oauth.ts
raphaelkabo: @coldlink and I were just talking about this. It's way too long!

src/shared/model/Routes.ts
raphaelkabo: For the IDAPI `GET /user?emailAddress=` call.

pvighi (approved):

PR #2727, opened 2024-05-13
OAuth | Refactor IDAPI/Okta account linking flow during authentication

raphaelkabo:
## Accessibility


src/server/lib/okta/fixProfile.ts
```diff
@@ -31,7 +31,7 @@ export const fixOktaProfile = async ({
 		await updateUser(oktaId, {
 			profile: {
 				legacyIdentityId: idapiUser.id,
-				searchPartitionKey: sha256(idapiUser.id),
+				searchPartitionKey: sha256Hex(idapiUser.id),
```
pvighi, line 34: what values was this generating before ? if there are values that are not valid hex then the user will be excluded from the export

coldlink: As discussed in standup, these were accidentally generating base64 encoded values, where they should've been in hex. @raphaelkabo wrote a script to backfill users affected by this.

pvighi: for future reference: we talked about this during standup there's a script being run to fix the users with the search partition in the wrong format

src/server/routes/oauth.ts
```diff
@@ -256,38 +181,51 @@ const authenticationHandler = async (
...
-				const validTokenSet = await fetchValidTokenSet({
+				// We have a refresh token and no legacy IDAPI ID, which means the user is
+				// in one of two states:
+				// 1. They exist in Okta _AND_ IDAPI, but the two accounts aren't linked (i.e.
+				//    the legacyIdentityId field is missing from the Okta user profile).
+				// 2. They existed in Okta but _NOT_ in IDAPI - which will have been fixed
+				//    in the exchangeAccessTokenForCookies call above.
+
+				// Run the fix function to link the two accounts. This won't be needed
+				// in case (2) above, but it's cheaper to just run it than to check first.
```
pvighi, line 192: as far as I know we are not sure why sometimes users don't have idapi ids in okta even though they have idapi accounts. Given that the function the okta hook calls is pretty much the same as we call manually here I'm not 100% sure the user has the idapi id in okta at this point.

pvighi: I was thinking maybe the idapi call returns success in some cases even though the okta function didn't actually update the user profile ?

pvighi (approved):
👍

PR #2728, opened 2024-05-14
Geolocation | Remove GU_geo_country cookie, stop checking CMP for geolocation

raphaelkabo:
## What does this change?

This prepares the ground for adding some new registration newsletter consents in a subsequent PR.

The key changes are:

- We remove all usage of the deprecated `GU_geo_country` cookie, in favour of the `X-GU-Geolocation` HTTP header sent from Fastly. In Cypress, because mocking a header in E2E tests isn't worth the hassle, we mock the geolocation state with the `cypress-mocked-state` cookie we used elsewhere.
- We remove the requirement for CMP consent to be given for geolocation data to be used. For this reason, we will now always set `registrationLocation` in Okta, and always show geolocated newsletters in the onboarding flow. We remove all non-geolocated newsletter mappings and all uses of the `PermissionedGeolocation` type - it's now just `Geolocation`.

## Tests

- [x] Tested in CODE - CMP consented - UK
- [x] Tested in CODE - CMP not consented - UK
- [x] Tested in CODE - CMP consented - Australia
- [x] Tested in CODE - CMP not consented - Australia


src/server/lib/getRegistrationLocation.ts
```diff
@@ -11,20 +15,35 @@ import { RegistrationLocation } from '@/shared/model/RegistrationLocation';
...
+	/**
+	 * Cypress Test START
+	 *
+	 * This code checks if we're running in Cypress
+	 */
+	const runningInCypress = process.env.RUNNING_IN_CYPRESS === 'true';
+	if (runningInCypress) {
+		const cypressMockStateCookie = req.cookies['cypress-mock-state'];
+		return countryCodeToRegistrationLocation(cypressMockStateCookie);
+	}
```
coldlink, line 28: I suggest making this explicitly check for the specific values that you use in the test, rather than a generic value, as i can forsee a scenario where in a test this could be difficult to debug when using a mock state cookie.

```suggestion
	const runningInCypress = process.env.RUNNING_IN_CYPRESS === 'true';
	if (runningInCypress) {
		const cypressMockStateCookie = req.cookies['cypress-mock-state'];
		
		// check if the cookie value is an expected value
		const validCode = ['FR', 'GB', 'US'].some(code => code === cypressMockStateCookie);
		
		// if it is then return the code
		if (validCode) {
			return countryCodeToRegistrationLocation(cypressMockStateCookie);
		}
		
		// otherwise let it fall through to the default check below
	}
```

src/server/routes/register.ts
```diff
@@ -140,7 +139,7 @@ const OktaRegistration = async (
 		return res.sendStatus(418);
 	}
 
-	const { email = '', _cmpConsentedState = false } = req.body;
```
coldlink, line 143: if you're removing the `_cmpConsentedState` here, is it also possible to remove it from the form body if its not being used? think there is a component that adds it in

coldlink (changes requested):

coldlink:
also this commit needs to be reworded, has `wip` in the name

https://github.com/guardian/gateway/pull/2728/commits/2c2b2d33bc7000847f923b797c7b18a2d1792b2e

src/client/components/ConsentsForm.tsx
```diff
@@ -5,8 +5,6 @@ import useClientState from '@/client/lib/hooks/useClientState';
 import { onboardingFormSubmitOphanTracking } from '@/client/lib/consentsTracking';
 import { CsrfFormField } from '@/client/components/CsrfFormField';
 import { buildUrlWithQueryParams } from '@/shared/lib/routeUtils';
-import { CmpConsentedStateHiddenInput } from '@/client/components/CmpConsentStateHiddenInput';
-import { useCmpConsent } from '@/client/lib/hooks/useCmpConsent';
```
coldlink, line 9: I'm assuming `useCmpConsent` is used somewhere else, which is why it hasn't been removed from this PR?

coldlink (approved):

src/client/components/ConsentsForm.tsx
raphaelkabo: Yeah, it's used to determine whether the ad personalization permission gets shown on the onboarding data page.

PR #2731, opened 2024-05-16
Add /unsubscribe-all endpoint

tjmw:
## What does this change?

Adds a new `/unsubscribe-all` endpoint, to be used in the `List-Unsubscribe` header of emails.

According to [the spec](https://datatracker.ietf.org/doc/html/rfc8058) this must be a `POST` endpoint. By default CSRF protection is enabled on all `POST`s, but that's not appropriate in this case as the request is made from an external location. So I've had to modify the middleware to allow `POST`s to this endpoint with no CSRF token.

See also guardian/identity#2514.

## How to test

I've done some end to end testing in CODE. I configured the List-Unsubscribe header in Braze B2C - Dev, and manually posted to the CODE url to simulate the email client making that request. It unsubscribed me from all newsletters and marketing email consents as expected.

## Accessibility

### DIFF
--- a/src/server/lib/idapi/subscriptions.ts
+++ b/src/server/lib/idapi/subscriptions.ts
@@ -119,7 +119,6 @@ export const makeUnsubscribeAllRequest = async (
 			request_id,
 		});
 
-		// Maybe we don't return an error message here? There's nothing user facing
 		throw new IdapiError({ message: 'Unsubscribe all failed', status: 500 });
 	}
 };
--- a/src/server/lib/idapi/subscriptions.ts
+++ b/src/server/lib/idapi/subscriptions.ts
@@ -17,6 +17,11 @@ interface SubscriptionData {
 	timestamp: number;
 }
 
+interface UnsubscribeAllData {
+	userId: string;
+	timestamp: number;
+}
+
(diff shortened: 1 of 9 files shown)


src/server/lib/idapi/subscriptions.ts
```diff
@@ -35,6 +40,19 @@ export const parseSubscriptionData = (data: string): SubscriptionData => {
...
+export const parseUnsubscribeAllData = (data: string): UnsubscribeAllData => {
+	const [userId, timestamp] = data.split(':');
+
+	if (!userId || !timestamp) {
+		throw new Error('Invalid unsubscribe-all data');
+	}
+
+	return {
+		userId,
+		timestamp: +timestamp,
```
raphaelkabo, line 52: The `+` is very elegant! I would have probably done `Number(timestamp)` because it feels more explicit to me that it's making a number, but that's very much a personal preference.

raphaelkabo (approved):
This is great work!

src/server/lib/idapi/subscriptions.ts
tjmw: Yeah it's a good question. I went for consistency with the existing method just above this which also uses `+timestamp`. What do you reckon? (I think we should either update both or update neither).

PR #2739, opened 2024-05-23
Show account linking page for social sign in guest users

coldlink:
## What does this change?

Currently users who are "guest" accounts (i.e users in Okta who are not in the `ACTIVE` state) get blocked from authenticating with a social identity provider by Okta. Okta returns an error to us with the `error=access_denied&error_description=User+status+is+invalid.` query parameters, however this error was not being handled correctly, and users were either being show a "Not Found" or "Error" page, which wasn't proving to be great UX.

Ideally we'd rectify the issue when the user authenticates, however all of social authentication happens through Okta, and all we get back from Okta is either a success (user created and authenticated) or failure (something went wrong). We're still investigating how best we should manage this.

For now we should improve the user experience. In Gateway we already have the "Account Linking" failed page for social users for a different purpose (if the user wasn't in the correct group). This page doesn't have social buttons, only email and password, and an error message explaining that we were unable to sign in with social and to use email and password instead.

This PR now handles the specific error correctly, and we also redirect to this account linking denied page. We also expand the error message to suggest reseting their password, or contacting support. We also handle any other OAuth errors using our generic handler.

<table>
<tr>
 <th>
Screenshot
 <th>
Video
<tr>
 <td>

localhost_6006_iframe html_globals=viewport_MOBILE id=pages-signin--social-signin-blocked viewMode=story(iPhone 14 Pro Max) (1)

 <td>

https://github.com/guardian/gateway/assets/13315440/226408e9-7e9c-44d7-bae5-922bfe8cad4f

</table>

## Tested
- [x] CODE

### DIFF
--- a/cypress/integration/shared/sign_in.shared.ts
+++ b/cypress/integration/shared/sign_in.shared.ts
@@ -368,9 +368,9 @@ export const showsAnErrorMessageAndInformationParagraphWhenAccountLinkingRequire
 					: '/signin?error=accountLinkingRequired';
 				cy.visit(visitUrl);
 				cy.contains(
-					'We cannot sign you in with your social account credentials. Please enter your account password below to sign in.',
+					'We could not sign you in with your social account credentials. Please sign in with your email below.',
 				);
-				cy.contains('Account already exists');
+				cy.contains('Social sign-in unsuccessful');
 			},
 		] as const;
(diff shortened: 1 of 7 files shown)


raphaelkabo (approved):
🦀 🚫

PR #2741, opened 2024-05-24
New Account Flow |  Do not show newsletters page if no newsletters to show

raphaelkabo:
## What does this change?

- If for some reason there aren't any newsletters to show a user on the newsletters page, we just continue immediately on from the newsletters page to the `returnUrl`, returning the user to their article. No point showing an empty page!
- Fixes an errant `.only` in a Cypress test suite.

## Tests

- [x] Tested in CODE


akinsola-guardian (approved):
Seems sensible to me.

PR #2752, opened 2024-05-29
Passcodes | Remove `usePasscodeRegistration` query parameter and make passcode registration default

coldlink:
## What does this change?

*~Note: This PR is currently pointing to #2639 (`mm/passcodes-ete`), once that PR is merged this PR should point to `main` instead~*

*~Note: This PR is currently pointing to #2773 (`mm/passcodes-fixes`), once that PR is merged this PR should point to `main` instead~*

*~Note: This PR is currently pointing to #2779 (`mm/passcode-bugs`), once that PR is merged this PR should point to `main` instead~*

While this PR may seem large at first glance, it's actually pretty straightforward, most of the changes relate to test files.

This PR simply removes the `usePasscodeRegistration` query parameter, which were introduced in #2639, thus making new account creation with passcodes the default.

It also updates the cypress tests that previously relied on the old flow to use passcodes instead, this is what is the largest change with this PR.

A `useOktaClassic` flag is also added in, so we can fallback and/or test the old Okta Classic flow behaviour continues to work, and cypress tests were kept for this.

# Tested

- [x] CODE

### DIFF
--- a/cypress/fixtures/okta-responses/error/idx-enroll-new-existing-user-response.json
+++ b/cypress/fixtures/okta-responses/error/idx-enroll-new-existing-user-response.json
@@ -1,0 +1,94 @@
+{
+	"code": 400,
+	"response": {
+		"version": "1.0.0",
+		"stateHandle": "02.id.state~c.handle",
+		"expiresAt": "2099-12-31T23:59:59.000Z",
+		"remediation": {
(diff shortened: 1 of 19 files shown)


raphaelkabo (approved):
Screenshot 2024-06-20 at 16 43 35

PR #2773, opened 2024-06-11
Passcodes | Fix issues after round one of testing

coldlink:
## What does this change?

During testing passcodes for registration, we found two issues and their solutions:

1. When clicking the "send again" functionality from the "email sent" page, the page would refresh, and the email sent, but there was no UI indication of anything happening.
  -  We now add the `emailSentSuccess` parameter when taking this action, so a UI message appears saying an email has been sent
2. On the link/session expired page (`/welcome/expired`), the form action doesn't know about passcodes, so uses the Okta classic functionality by default
  - We've changed this so that if passcodes are enabled, we use the default passcode registration flow to resolve this.

We also make some copy changes after review, specifically on the email sent page, and the email sent to readers.


akinsola-guardian (approved):
Looks sensible to me.

PR #2774, opened 2024-06-11
Passcodes | Enable `registrationPasscodesEnabled` switch on PROD

coldlink:
## What does this change?

Flick the `registrationPasscodesEnabled` to enabled on `PROD`.

This means that passcodes will be enabled on `PROD`, but will still be behind the `usePasscodeRegistration` query parameter flag until https://github.com/guardian/gateway/pull/2752 is merged in.

This will allow us to do testing in PROD, but not release to users who don't know about the flag.


raphaelkabo (approved):
🚀

PR #2779, opened 2024-06-18
Passcodes | Bug Fixes

coldlink:
## What does this change?

Fixes a few bugs found after testing behind the feature flag. Each one is commit, so can step through in commit order when reviewing.

### `/register/code/expired` not found

If a user takes too long to type a passcode, they should be redirected to an expired page.

The idea initially was to make a new `/register/code/expired` page. This was never implemented in the original PR, so if a user ended up in this scenario they would see a 404 page.

At first we thought we should create a new page under this url, but since this has the exact same functionality as `/welcome/expired` we should just use that one, which is what this commit does.

### Refactor `EmailSent` page into new `PasscodeEmailSent` page and existing `EmailSent` page

The display logic in the `EmailSent` page component was getting super messy and difficult to reason about, more so when passcodes were added in.

It is a sensible decision to make a new EmailSent page component for when we're using passcodes. So we create a `PasscodeEmailSent` component to handle this behaviour.

We also remove the `pageHeader` prop from `EmailSent` component as it was no longer being used.

### Cache headers not being sent

*tl;dr: express ghosts*

We noticed that the `Cache-Control` and `Pragma` headers weren't been sent in the response headers, which is vitally important as we don't want to cache most pages in Gateway, and so we need these headers set to specific values to tell browsers (and Fastly) not to cache.

This meant that browsers fell back on a default caching policy, where new pages would be retrieved from the server, but using forwards/backwards in the browser would use the local cached page instead.

We assumed that the no caching behaviour would be working, as the `noCache` middleware is set on our uncached routes. However investigating further we found that this middleware was not being called!

After much investigation, the culprit turned out to be the `emailTemplate` routes. Why? We don't know. Express ghosts probably. Temporarily removing `emailTemplates` would fix the problem, or adding `emailTemplates` to the uncached routes.

So the solution was the latter, we made `emailTemplates` an `uncachedRoute` too, and the correct headers were being sent on all other routes now! `emailTemplates` don't need to be cached anyway, as it's only used by terraform for getting email templates to use in Okta.

### Passcode Used Message

Users who click back in their browser on the next page after submitting their passcode will see the page again to enter the passcode. Entering the passcode again here will cause errors to be returned, and a user left confused, as the passcode has already been used.

So we want to block users from doing this again, and continue in the journey.

We set a parameter `passcodeUsed` when the passcode is used, and if the user navigates back to the enter passcode page, this parameter is checked. If the `passcodeUsed` is set, we show a new page saying that the code has already been verified and a CTA asking the user to continue.

60929d168594f80039336501-mgoczjxxiv chromatic com__path=_story_pages-passcodeused--defaults globals=viewport_iphone6

### Use expiry on set password page when using passcodes

If a user takes too long to set a password on the password set page, they will get shown an error saying something went wrong.

This happens because the Okta IDX state expired while the user was on the page.

We need to set the `timeUntilTokenExpiry` based on the `expiresAt` from the `stateHandle` on the set password page, so if the token expires while on the set password page, they will be shown the expired page instead.

### CODE
- [x] `/register/code/expired` not found
- [x] Cache headers not being sent
- [x] Passcode Used Message
- [x] Use expiry on set password page when using passcodes

### DIFF
--- a/cypress/integration/ete-okta/registration_2.6.cy.ts
+++ b/cypress/integration/ete-okta/registration_2.6.cy.ts
@@ -952,7 +952,7 @@ describe('Registration flow - Split 2/2', () => {
 
 					// passcode page
 					cy.url().should('include', '/register/email-sent');
-					cy.contains('Email sent');
+					cy.contains('Email with verification code sent');
 
 					cy.get('input[name=code]').type(code!);
 					cy.contains('Submit verification code').click();
--- a/cypress/integration/ete-okta/registration_2.6.cy.ts
+++ b/cypress/integration/ete-okta/registration_2.6.cy.ts
@@ -832,6 +832,48 @@ describe('Registration flow - Split 2/2', () => {
 			);
 		});
 
+		it('passcode used functionality', () => {
+			const unregisteredEmail = randomMailosaurEmail();
+			cy.visit(`/register/email?usePasscodeRegistration=true`);
+
+			const timeRequestWasMade = new Date();
+			cy.get('input[name=email]').type(unregisteredEmail);
+			cy.get('[data-cy="main-form-submit-button"]').click();
+
+			cy.contains('Enter your code');
+			cy.contains(unregisteredEmail);
+			cy.contains('send again');
+			cy.contains('try another address');
+
+			cy.checkForEmailAndGetDetails(unregisteredEmail, timeRequestWasMade).then(
+				({ body, codes }) => {
+					// email
+					expect(body).to.have.string('Your verification code');
+					expect(codes?.length).to.eq(1);
+					const code = codes?.[0].value;
+					expect(code).to.match(/^\d{6}$/);
+
+					// passcode page
+					cy.get('input[name=code]').clear().type(code!);
+					cy.contains('Submit verification code').click();
+
+					cy.url().should('contain', '/welcome/password');
+
+					cy.go('back');
+					cy.url().should('contain', '/register/email');
+					cy.contains('Email verified');
+					cy.contains('Complete creating account').click();
+
+					cy.url().should('contain', '/welcome/password');
+
+					cy.get('input[name="password"]').type(randomPassword());
+					cy.get('button[type="submit"]').click();
+
+					cy.url().should('contain', '/welcome/review');
+				},
+			);
+		});
+
 		it('passcode incorrect functionality', () => {
 			const unregisteredEmail = randomMailosaurEmail();
 			cy.visit(`/register/email?usePasscodeRegistration=true`);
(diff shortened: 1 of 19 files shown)


src/client/pages/PasscodeEmailSent.tsx
```diff
@@ -0,0 +1,170 @@
...
+			if (codeInput) {
+				codeInput.focus();
+			}
+		}
+	}, []);
+
+	return (
+		<MinimalLayout
+			pageHeader="Enter your code"
+			successOverride={showSuccess ? 'Email sent' : undefined}
```
raphaelkabo, line 99: I think we should say 'Verification code sent' here... makes it obvious what the action was.

src/client/pages/PasscodeEmailSent.tsx
```diff
@@ -0,0 +1,170 @@
...
+			errorOverride={
+				recaptchaErrorMessage ? recaptchaErrorMessage : errorMessage
+			}
+			errorContext={recaptchaErrorContext}
+			imageId="email"
+		>
+			{children}
```
raphaelkabo, line 106: What are the `children` for this page? Does it ever have `children`?

cypress/integration/ete-okta/registration_2.6.cy.ts
```diff
@@ -832,6 +832,48 @@ describe('Registration flow - Split 2/2', () => {
...
+					expect(code).to.match(/^\d{6}$/);
+
+					// passcode page
+					cy.get('input[name=code]').clear().type(code!);
+					cy.contains('Submit verification code').click();
+
+					cy.url().should('contain', '/welcome/password');
+
+					cy.go('back');
+					cy.url().should('contain', '/register/email');
```
raphaelkabo, line 863: Why not check that `passcodeUsed` is in the URL here? The other tests check that the page looks right, but it'd be also good to have that checked.

src/client/pages/PasscodeEmailSent.tsx
coldlink: Thats what it used to say, or something along those lines, it was simplified.

src/client/pages/PasscodeEmailSent.tsx
coldlink: it does not, will remove

cypress/integration/ete-okta/registration_2.6.cy.ts
coldlink: because it's not in the url, it's in the encrypted state cookie which we extract into the page data

src/client/pages/PasscodeEmailSent.tsx
coldlink: The normal `EmailSent` component however does use it in another scenario.

src/client/pages/PasscodeEmailSent.tsx
coldlink: Updated to `Email with verification code sent`

raphaelkabo (approved):
Good bugs well caught! 🐛

PR #2780, opened 2024-06-19
Account Review Flow | Remove SC_GU_U cookie use

coldlink:
## What does this change?

@pvighi noticed that Gateway was still making a few calls to IDAPI with the SC_GU_U cookie when it shouldn't be doing so. These were isolated to the `PATCH /users/me/newsletters` and `GET /users/me/newsletters` endpoints in IDAPI.

In Gateway the only place using these endpoints was the account review flow's newsletter page (`/welcome/newsletters`).

The issue occurred because the `loginMiddlewareOAuth` was left out from the `POST /welcome/newsletters` endpoint, meaning that the access token would be undefined on the request state. However since the SC_GU_U cookie was available on the request headers, it would use that to call IDAPI instead, thus causing this issue.

This PR adds the `loginMiddlewareOAuth` back in, and removes any usage of the `SC_GU_U` cookie.

### DIFF
--- a/src/server/routes/welcome.ts
+++ b/src/server/routes/welcome.ts
@@ -351,7 +351,6 @@ router.get(
 		const newsletters = await getUserNewsletterSubscriptions({
 			newslettersOnPage: NewsletterMap.get(geolocation) as string[],
 			ip: req.ip,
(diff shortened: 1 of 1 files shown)


pvighi (approved):
looks good to me

PR #2781, opened 2024-06-20
Validate Return URL | Allowlist certain subdomains to preserve query parameters

coldlink:
## What does this change?

We currently don't preserve query parameter for non `profile` Guardian subdomains when they provide us with a `returnUrl`.

However some domains, such as `support` need access to query parameters for functionality.

This PR adds an allowlist of domains which will preserve query parameter.

### DIFF
--- a/src/server/lib/validateUrl.ts
+++ b/src/server/lib/validateUrl.ts
@@ -7,6 +7,10 @@ const validHostnames = [
 	'.thegulocal.com',
 ];
 
+// valid subdomains
+const validSubdomains = ['profile.', 'support.'];
+
+// invalid paths for profile subdomain
 const invalidPaths = ['/signout'];
 
 const { defaultReturnUri } = getConfiguration();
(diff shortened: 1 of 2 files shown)


akinsola-guardian (approved):
LGTM

src/server/lib/validateUrl.ts
```diff
@@ -27,15 +31,18 @@ export const validateReturnUrl = (returnUrl = ''): string => {
...
 
-		// if the hostname is on the profile subdomain, we also want to check the pathname is valid
-		// and keep any query params
-		if (url.hostname.startsWith('profile.')) {
-			// then check the pathname is valid
-			if (invalidPaths.some((path) => url.pathname.startsWith(path))) {
-				throw 'Invalid pathname';
+		// if valid subdomains are present, we can return the url with the query params
+		if (
+			validSubdomains.some((subdomain) => url.hostname.startsWith(subdomain))
```
jamesgorrie, line 36: If I have a domain that is `support-this-baddy.theguardian.com` or `support-this-naughty.oops.com` - would this still work?

coldlink: It would not, we're specifically looking for the `support.` string at the start of the hostname and `.theguardian.com`

jamesgorrie: Ah, [this was the code I was missing](https://github.com/guardian/gateway/pull/2781/files?diff=split&w=0#diff-ca394c2b701ae7115acc7166b12451ced999603cb59d6a1a0691c8eb6e9a7790L26-L28).

PR #2786, opened 2024-06-24
Passcodes | Improve passcode styling/functionality 

coldlink:
### Passcode Email

- Update the email template to include 4px letter spacing to make it easier to read (based on [Guardian Source `space[1]`](https://guardian.github.io/storybooks/?path=/docs/source_foundations-space--docs))
- Remove bold text under the passcode to make the code the main CTA

## Passcode Input

- Add letter spacing to the passcode input field, for the same reason as above, this time using `remSpace[1]`
- Add client side code to validate the input field, e.g. sanitising the text on input and on paste
- Add server side code to validate the code sent to the server before we send it to Okta as we shouldn't rely on client side only validation
  - Even though Okta validates the input for us, we can avoid making requests to Okta if we check the input format ourselves

## Screenshots

| Email | Page |
|--------|--------|
| 60929d168594f80039336501-rhqwmpyrqw chromatic com_iframe html_args= globals=viewport_mobile id=email-templates-registrationpasscode--passcode viewMode=story(iPhone 12 Pro) | 60929d168594f80039336501-rhqwmpyrqw chromatic com_iframe html_args= globals=viewport_mobile id=pages-passcodeemailsent--with-passcode viewMode=story(iPhone 12 Pro) |

### DIFF
--- a/src/client/components/PasscodeInput.tsx
+++ b/src/client/components/PasscodeInput.tsx
@@ -3,6 +3,7 @@ import { FieldError } from '@/shared/model/ClientState';
 import { TextInputProps } from '@guardian/source/react-components';
 import { css } from '@emotion/react';
 import ThemedTextInput from '@/client/components/ThemedTextInput';
+import { remSpace } from '@guardian/source/foundations';
 
 interface PasscodeInputProps extends Omit<TextInputProps, 'label'> {
 	passcode?: string;
@@ -11,6 +12,7 @@ interface PasscodeInputProps extends Omit<TextInputProps, 'label'> {
 
 const passcodeInputStyles = css`
 	text-align: center;
+	letter-spacing: ${remSpace[0]};
 `;
 
 export const PasscodeInput = ({
--- a/src/client/components/PasscodeInput.tsx
+++ b/src/client/components/PasscodeInput.tsx
@@ -1,6 +1,6 @@
+import React, { useState } from 'react';
 import { FieldError } from '@/shared/model/ClientState';
 import { TextInputProps } from '@guardian/source/react-components';
-import React from 'react';
 import { css } from '@emotion/react';
 import ThemedTextInput from '@/client/components/ThemedTextInput';
 
@@ -14,9 +14,44 @@ const passcodeInputStyles = css`
 `;
 
 export const PasscodeInput = ({
-	passcode,
+	passcode = '',
 	fieldErrors,
 }: PasscodeInputProps) => {
+	/**
+	 * In gateway we normally avoid using client side javascript, but in this case
+	 * we allow it as it an enhancement to the user experience and not required for
+	 * the form to work.
+	 */
+
+	// This function removes all non-numeric characters from the input
+	const sanitiseCode = (code: string) => {
+		return code.replace(/\D*/g, '');
+	};
+
+	// Using an object to store the input value so that we re-render the component
+	// whenever there is an user input, rather than only when the state changes
+	const [input, setInput] = useState({ value: sanitiseCode(passcode) });
+
+	// Update the input value on user input
+	const onChange = (e: React.ChangeEvent<HTMLInputElement>) => {
+		setInput({
+			value: sanitiseCode(e.target.value),
+		});
+	};
+
+	// Update the input value on paste
+	const onPaste = (e: React.ClipboardEvent<HTMLInputElement>) => {
+		const paste = e.clipboardData.getData('text/plain');
+
+		if (!paste) {
+			return;
+		}
+
+		setInput({
+			value: sanitiseCode(paste),
+		});
+	};
+
 	return (
 		<div>
 			<ThemedTextInput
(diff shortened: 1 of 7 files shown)


src/email/components/Text.tsx
```diff
@@ -26,7 +28,7 @@ export const Text = ({
 				padding="0"
 				fontSize={largeText ? '20px' : '17px'}
 				lineHeight="1.35"
-				letterSpacing="-0.02px"
+				letterSpacing={letterSpacing ? letterSpacing : '-0.02px'}
```
raphaelkabo, line 31: Why is _all_ of our text slightly squished?

src/client/components/PasscodeInput.tsx
```diff
@@ -14,9 +14,44 @@ const passcodeInputStyles = css`
...
 }: PasscodeInputProps) => {
+	/**
+	 * In gateway we normally avoid using client side javascript, but in this case
+	 * we allow it as it an enhancement to the user experience and not required for
+	 * the form to work.
+	 */
+
+	// This function removes all non-numeric characters from the input
+	const sanitiseCode = (code: string) => {
+		return code.replace(/\D*/g, '');
```
raphaelkabo, line 30: TIL about `\D`!

src/client/components/PasscodeInput.tsx
```diff
@@ -30,6 +65,9 @@ export const PasscodeInput = ({
 				error={fieldErrors?.find((error) => error.field === 'code')?.message}
 				defaultValue={passcode}
 				cssOverrides={passcodeInputStyles}
+				value={input.value}
+				onChange={onChange}
+				onPaste={onPaste}
```
raphaelkabo, line 72: A paste event should trigger `onChange`, why is it needed separately?

src/server/routes/register.ts
```diff
@@ -176,6 +176,18 @@ router.post(
 			const { stateHandle } = encryptedState;
 
 			try {
+				// validate the code contains only numbers and is 6 characters long
+				// The okta api will validate the input fully, but validating here will prevent unnecessary requests
+				if (!/^\d{6}$/.test(code)) {
+					throw new OAuthError(
+						{
+							error: 'api.authn.error.PASSCODE_INVALID',
```
raphaelkabo, line 184: Are we happy with the error message text being `'Incorrect code'`? Could we make it a bit more user-friendly, similar to the other `RegistrationError`s?

src/email/components/Text.tsx
coldlink: It's because the designs called for this: https://github.com/guardian/gateway/pull/1133 (in figma file https://www.figma.com/file/ZDPonnBXbMhV2smf51Dylc/?node-id=4175%3A126191).

So yeah 🤷🏾

src/server/routes/register.ts
coldlink: I think that's a future UX improvement that should happen. Most users shouldn't get to this scenario anyway as the client side validation will stop them, this is here as a failsafe, and is the same error that would be returned by okta were an invalid token be supplied.

src/client/components/PasscodeInput.tsx
coldlink: I suppose that's true, but having a paste event keeps it more flexible if we were to for example add an auto submit functionality.

raphaelkabo: My concern is that we're needlessly re-rendering twice on paste events - it's a tiny issue really.

raphaelkabo: Also, I don't think we'd want an auto-submit `onPaste`. I think we'd want an auto-submit when `input.value.length === 6`.

src/email/components/Text.tsx
coldlink: Oooh https://github.com/guardian/gateway/pull/1133/files#r756130367

raphaelkabo:
Nick and I think that the new letter-spacing is too wide for the passcode. It looks weirdly like 6 separate numbers, which conversely makes it _harder_ to parse as a single number, and I worry people will think they can't copy it because it looks like it has spaces between it. Can we halve the spacing?

coldlink:
> Nick and I think that the new passcode letter-spacing in the email is too wide. It looks weirdly like 6 separate numbers, which conversely makes it _harder_ to parse as a single number, and I worry people will think they can't copy it because it looks like it has spaces between it. Can we halve the spacing?

Have updated both email and input to use the smaller spacing as I feel that they should be related.

src/client/components/PasscodeInput.tsx
coldlink: I've added the `onPaste` back in, as there was strange behaviour when pasting when only using `onChange`.

Essentially lets say your clipboard has an extra space or another at the start of the string by accident, e.g. ` 123456`. Then when pasting in when only using `onChange` the input value is left at `12345` for some reason, as it seems to only look up to the 6th character before sanitising. When using `onPaste` this doesn't happen and the correct input value of `123456` is appled.

raphaelkabo: Oh, that's super weird. I think that's a good enough reason to keep it in!

raphaelkabo (approved):

src/client/components/PasscodeInput.tsx
coldlink: I wonder if it's something to do with the html validation interfering with the javascript/react validation?

either way, it's fine as it is, gateway doesn't use do much rerendering at all

PR #2787, opened 2024-06-28
AB tests: revert registration offer emails AB test

raphaelkabo:
## What does this change?

Reverts the AB test set up in https://github.com/guardian/gateway/pull/2598, as it's now over. This isn't a straight revert as some of the code we introduced in that PR is now being used elsewhere, so this just removes the AB test code only.

## Testing

- [x] Tested on CODE


coldlink (approved):
🔥

PR #2791, opened 2024-07-03
use the email address from the create test user response

johnduffell:
Traditionally test users have used the first name field to determine whether someone is a test user or not.

However this is less reliable than using email address as the email address field is passed around everywhere whereas there has to be special logic to get test users' first name around to the right places (as for a "real" user, first name would be extra PII)

In some other PRs, changes are proposed to embed the generated token into email addresses.  This means that the email address is generated on the fly, so anything using the create test user endpoint, needs to look in the response to find out what the generated email address is.

--

This PR updates gateway to get the email address out of the response, rather than relying on what was passed in to be correct.  See https://github.com/guardian/identity/pull/2528 for the PR that makes the changes that this PR depends on.

### DIFF
--- a/cypress/support/commands/testUser.ts
+++ b/cypress/support/commands/testUser.ts
@@ -253,7 +253,7 @@ export const createTestUser = ({
 			})
 			.then((res) => {
 				return cy.wrap({
-					emailAddress: finalEmail,
+					emailAddress: res.body.emailAddress as string,
 					cookies: res.body.values as IDAPITestUserResponse,
 					finalPassword,
 				});


cypress/support/commands/testUser.ts
```diff
@@ -253,7 +253,7 @@ export const createTestUser = ({
 			})
 			.then((res) => {
 				return cy.wrap({
-					emailAddress: finalEmail,
+					emailAddress: res.body.emailAddress as string,
```
johnduffell, line 256: this field is added here
https://github.com/guardian/identity/pull/2528/files#diff-dc9b5164ef33bc1fce5db7e899dbb88e983e78c983e09c566eca504b66445efbR37

johnduffell:
> Cypress seems to be failing now!

thanks yes once the PR this depends on is merged, the tests should be able to pass
See https://github.com/guardian/identity/pull/2528 for the PR that makes the changes that this PR depends on.

paulmr:
Hi @johnduffell , we are cleaning up our PR backlog and as this PR is quite old we are going to close it. Please feel free to get in touch if you think it's still relevent and we can re-open and discuss the next steps.

Thanks, -Paul

PR #2792, opened 2024-07-05
Delete |  Block feast users from deleting their account if active subscription

coldlink:
## What does this change?

We noticed that users who had only feast subscriptions could delete their account, this was because we weren't checking if they had a feast subscription when checking if a user is eligible for self deletion.

This PR adds that in.

This does not fix a separate issue where it's possible for a user to delete their account and then purchase a product. That will involve more investigation.


raphaelkabo (approved):
Great catch! 🚀

PR #2796, opened 2024-07-15
Dependencies | 15th July 2024

coldlink:
## What does this change?

Changes from https://github.com/guardian/gateway/pull/2793:

Holds back the following:

- Sentry to v8 - major version upgrade, need to do in a separate PR

## Tested

- [ ] CODE


akinsola-guardian (dismissed):

cypress/integration/ete/consent_token/consent_token.2.cy.ts
```diff
@@ -21,7 +21,8 @@ describe('Consent token flow', () => {
 					timeRequestWasMade,
 					/consent-token\/([^"]*)/,
 				).then(({ token }) => {
-					cy.visit(`/consent-token/${token}/accept`, {
+					// the token has the /accept suffix already appended to it
+					cy.visit(`/consent-token/${token}`, {
```
raphaelkabo, line 25: Could we do this the same way I've got it in the other PR? https://github.com/guardian/gateway/pull/2794/files#diff-a276636b5554d2fffaad0cb0bcfeff0e5b0844b59231f73c1bd6db2e91f7699bR14

Tweaking the regex is more future-proof, I think.

coldlink: thats a much more sensible approach, i'll update this

PR #2799, opened 2024-07-17
Password reset email update

nickhartyyy:
## What does this change?

We need to make it clear to readers that they can retry resetting their password if the link has expired. Many readers contact User Help to say their link has expired, but they don’t realise they can trigger a new one themselves. This change hopes to solve that problem.

### DIFF
--- a/src/email/templates/ResetPassword/ResetPassword.tsx
+++ b/src/email/templates/ResetPassword/ResetPassword.tsx
@@ -6,6 +6,7 @@ import { Header } from '@/email/components/Header';
 import { SubHeader } from '@/email/components/SubHeader';
 import { Text } from '@/email/components/Text';
 import { Footer } from '@/email/components/Footer';
+import { Link } from '@/email/components/Link';
 
 export const ResetPassword = () => {
 	return (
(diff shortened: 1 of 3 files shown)


raphaelkabo (approved):
📧 🚀 🫡

src/email/templates/ResetPassword/ResetPassword.tsx
```diff
@@ -16,7 +16,13 @@ export const ResetPassword = () => {
 			<Text>
 				You&apos;ve asked us to send you a link to reset your password.
 			</Text>
-			<Text noPaddingBottom>This link is valid for 60 minutes.</Text>
+			<Text noPaddingBottom>
+				This link is valid for 60 minutes. If your link has expired, please try{' '}
+				<a href="https://profile.theguardian.com/reset-password">
```
raphaelkabo, line 21: I think we have a specific Link component which we prefer using instead of `<a>`.

nickhartyyy: Thank you!

raphaelkabo (dismissed):
🚀

raphaelkabo (approved):
... 🚀 😡

PR #2806, opened 2024-07-24
Use devx-logging instead of custom Kinesis appender

AshCorr:
## What does this change?

Removes the custom kinesis appender for logging and rely on the [devx-logs](https://github.com/guardian/devx-logs) service provided by the cdk-base role.

This relies on adding some custom cloudformation tags to our EC2 instances which devx-logs uses to discover which service to read logs from and what kinesis stream to send them to.

The downside to this approach is that logs from DEV/local machine aren't sent to Kinesis anymore, however you could use something like [lnav](https://lnav.org/) if you want local log analysis.

## How to test

Deployed to CODE.

Based on the logs we can see the `ShippedBy` tag is now showing that Gateway is using devx-logs:

image

We can also see that custom fields also still work (I tested by adding an extra `port` field to the server start message)

image


coldlink (approved):
🥲 this is so much nicer than what we had

PR #2821, opened 2024-07-31
Alarm Blindness | Add some alarms!

coldlink:
## What does this change?

We weren't reporting on some possibly important alarms, so add them in!

#### `OAuthAuthenticationCallbackInactivityAlarm`
- Checks the successful OAuth Authorization Code flow callback for Authentication metric `OAuthAuthenticationCallback::Success`
- This metric is fired on every successful authentication in Gateway that is used to give the user a session
- This is crucial during all forms of authentication through Gateway, if this alarm is fired something is very wrong with authentication and should be investigated immediately.
- Alarm fires after 20 mins of no metric

#### `OAuthApplicationCallbackInactivityAlarm`
- Checks the successful OAuth Authorization Code flow callback for internal Gateway routes that require the user to be logged in `OAuthApplicationCallback::Success`
- This metric is fired when tokens are successfully received for specific gateway routes that require the user to be logged in, this is mainly the New Account Review flow and corresponding newsletter page.
- This is crucial as if this is failing, users are probably unable to finish setting up their account and logging in after creating an account
- Alarm fires after 1 hour of no metric

#### `DeletionInactivityAlarm`
- Checks if the user has been successfully deleted `OAuthDeleteCallback::Success`
- This metric is fired when the step function for the self service account deletion is successful
- This is a P2 as it's still important to investigate if this process is not working correctly 
- Alarm fires after 6 hours of no metric

#### `UnsubscribeAllInactivityAlarm`
- Checks if the user is able to unsubscribe from all communications using the `UnsubscribeAll::Success` metric
- This is based `List-Unsubscribe` header in an email which signals to email clients to allow a user to one click unsubscribe from emails: https://github.com/guardian/gateway/pull/2731
- This is a P2 as it's important to investigate as it could mean we're not compliant with some email provider services
- Alarm fires after 1 hour of no metric

### DIFF
--- a/cloudformation.yaml
+++ b/cloudformation.yaml
@@ -757,3 +757,36 @@ Resources:
         - !Ref 'TopicSendEmail'
       InsufficientDataActions:
         - !Ref 'TopicSendEmail'
+  DeletionInactivityAlarm:
+    Type: AWS::CloudWatch::Alarm
+    Condition: IsProd
+    Properties:
+      AlarmName: !Sub
+        - '${Priority} - ${App} ${Stage} has had no success self service user deletion in the last 6 hours'
+        - Priority: !FindInMap [StageVariables, AlarmPriorities, P2]
+      AlarmDescription: No one has successfully deleted their account in the last 6 hours.
+      Metrics:
+        - Id: totalDeletionCount
+          Expression: deleteAccountCount
+          Label: 'Total self service user deletions'
+        - Id: deleteAccountCount
+          MetricStat:
+            Metric:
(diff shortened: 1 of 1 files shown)


pvighi (approved):
looks good to me. 
Did you base the thresholds in logged data or something ? I have no idea now frequently some of this things happen

coldlink:
> looks good to me. Did you base the thresholds in logged data or something ? I have no idea now frequently some of this things happen

Yeah I based them on seeing the metrics in Cloudwatch

PR #2823, opened 2024-08-02
Migrate Github SES Send Email role to CDK

AshCorr:
## What does this change?

Moves the `GithubActionsSESSendEmailsRole` from YAML to CDK.


cdk/lib/identity-gateway.ts
```diff
@@ -10,5 +17,30 @@ export class IdentityGateway extends GuStack {
 		new CfnInclude(this, 'IdentityGateway', {
 			templateFile: '../cloudformation.yaml',
 		});
+
+		if (['CODE', 'TEST'].includes(props.stage)) {
+			new GuRole(this, 'GithubActionsSESSendEmailsRole', {
+				roleName: 'GithubActionsSESSendEmailsRole',
```
AshCorr, line 23: Preserve the role name to prevent the ARN from changing.

PR #2830, opened 2024-08-06
Add self service recovery for users in limbo state (not setting password after entering passcode)

coldlink:
## The Issue

If you create an account with email, and verify the passcode, but NOT set a password (e.g. by leaving the page), the account is left in a limbo state. This means that they're unable to log in, or perform recovery on themselves.

These affected users are left in the STAGED state within Okta, which means that they're not a “full” account (the ACTIVE state), and are left unverified. Normally a user would transition from the STAGED to the ACTIVE state by setting a password, which happens if a user inputs a password during the create account process.

For users in a non-ACTIVE state, we have programmatic remediation options for these users if they go through a “reset password” flow. For example for STAGED users [[code](https://github.com/guardian/gateway/blob/a51da8da16f8c65b577005b3b8163dc2018a4a14/src/server/controllers/sendChangePasswordEmail.ts#L141)] normally we attempt to call the Okta “Activate User” method which would create a token we could send to the user. However this fails with an error message “Password reset is not allowed in the user's current status” and the error code “E0000017”, which is why the user sees an error message.  This suggests that the normal remediation option does not work for these users.

## How to fix

Okta support mentioned that these users are left in a state where they’re unable to perform any recovery as they didn’t complete the flow. The way to recover these users is to deactivate these users, and then activate them again

## This PR

In this PR we implement the remediation option that Okta suggested.

Essentially when a user goes through the reset password or register flow, we check what state the user is in.

Users in this limbo state will be either in the STAGED or PROVISIONED state which is determined by if they tried to recover before this PR went in.

This PR adds a check for a specific error which occurs when users are in this specific state.

If this error is detected we perform the remediation, by calling the newly added `deactiveUser` function, we then just call the respective sendEmail function again. This time it will check for the `DEPROVISIONED` state, which as mentioned needs to activate the user again and send them an email. So we simply add a check for `DEPROVISIONED` to the `STAGED` check, as they both activate the user!

We also add logs and metrics so we can monitor these users.

Finally we add cypress tests to test these four scenarios.

Other changes were made to remove a rogue `.only` in a cypress test, and adjust some matrix behaviour.

## Tested

- [x] CODE


cypress/integration/ete/registration_1.2.cy.ts
```diff
@@ -4,8 +4,15 @@ import {
...
 
+const breachCheck = () => {
+	cy.intercept({
+		method: 'GET',
+		url: 'https://api.pwnedpasswords.com/range/*',
+	}).as('breachCheck');
+};
+
 describe('Registration flow - Split 1/2', () => {
-	context.only('Registering with Okta', () => {
```
coldlink, line 8: 🙈  not sure how long this has been here for

cypress/integration/ete/registration_1.2.cy.ts
```diff
@@ -530,7 +537,7 @@ describe('Registration flow - Split 1/2', () => {
 							).then(({ links, body }) => {
 								expect(body).to.have.string('This account already exists');
 								expect(body).to.have.string('Create password');
-								expect(links.length).to.eq(2);
+								expect(links.length).to.eq(3);
```
coldlink, line 540: ...but it did mean we can now fix these tests

src/server/controllers/sendChangePasswordEmail.ts
```diff
@@ -139,31 +140,80 @@ export const sendEmailInOkta = async (
...
+								);
+								throw error;
+							}
+
+							trackMetric(
+								'PasscodePasswordNotCompleteRemediation-ResetPassword-STAGED-Complete',
+							);
+
+							// rerun the sendEmailInOkta function to catch the user in the DEPROVISIONED state
+							return sendEmailInOkta(req, res);
```
pvighi, line 203: The logic should prevent it from happening so it's not necessary,  but we could add a retry counter or some other param as input to `sendEmailInOkta` to detect if we already tried once and make sure infinite loops are not possible.

src/server/controllers/sendChangePasswordEmail.ts
```diff
@@ -173,28 +223,77 @@ export const sendEmailInOkta = async (
...
+								);
+								throw error;
+							}
+
+							trackMetric(
+								'PasscodePasswordNotCompleteRemediation-ResetPassword-PROVISIONED-Complete',
+							);
+
+							// rerun the sendEmailInOkta function to catch the user in the DEPROVISIONED state
+							return sendEmailInOkta(req, res);
```
pvighi, line 281: now that I see this whole logic is duplicated down here for the provisioned users I'm not so sure that it's impossible to go into loops activating and deactivating the user, I'm not familiar enough with these Okta calls

src/server/lib/okta/register.ts
```diff
@@ -61,74 +62,183 @@ const sendRegistrationEmailByUserState = async ({
...
+						request_id,
+					},
 				);
-				throw new OktaError({
-					message: `Okta user reactivation failed: Failed to send email`,
-				});
+
+				// otherwise throw the error to the outer catch block
+				// as it's not handled in the if statement above
+				throw error;
```
pvighi, line 241: It seems like there's a lot of code dupilcation.  I think it makes it harder to read and maintain, but I won't hold a needed fix for that

pvighi (approved):
left some comments but looks like a good fix.

I don't really work in gateway a lot so it might be a deliberate choice, but to me it looks like functions could be broken down into smaller pieces for readability an maintainability (and reusing code)

cypress/integration/ete/new_account_review.3.cy.ts
```diff
@@ -168,6 +168,11 @@ describe('New account review page', () => {
 });
 
 describe('New account newsletters page', () => {
+	beforeEach(() => {
+		cy.intercept('GET', 'https://m.code.dev-theguardian.com/**', (req) => {
+			req.reply(200);
+		});
+	});
```
coldlink, line 175: this fixes an unrelated issue which started occurring in other tests, do to an error on the page it redirected to

since we just want to check the redirect happened, this was added in

src/server/controllers/sendChangePasswordEmail.ts
pvighi: although I see we are already using this pattern for other errors

PR #2831, opened 2024-08-07
Add `loopDetectionFlag` to prevent infinite loop in `PasscodePasswordNotCompleteRemediation` flow

coldlink:
## What does this change?

In https://github.com/guardian/gateway/pull/2830 we fixed an issue with account remediation for users stuck in a specific limbo state. See that PR for more details about the issue.

However since we added a recursive call in the fix, there is an unlikely, but not zero chance that an infinite loop could occur.

This PR adds a `loopDetectionFlag` so we know when we've called the function again recursively, and checking this flag to make sure we throw early should  a loop be detected!


pvighi (approved):
👍

PR #2833, opened 2024-08-07
Passwordless | Add additional IDX API endpoints required

coldlink:
## What does this change?

In order to step towards passwordless authentication using OTPs (One Time Passcodes) being an option for sign in and account recovery at the Guardian, we need to add some addition Okta IDX API endpoints to facilitate this that weren't required when we implemented social authentication and create account with OTPs.

This PR simply sets up these API endpoints, but does not use them anywhere.

Some refactoring was also performed to share similar objects with different endpoints.

### `/identify`

Used to start the authentication process for an existing user. Available after the initial call to the `/introspect` endpoint.

This endpoint takes the `stateHandle` as with other endpoints, which identifies the current request state and context.
It also takes an `identifier` parameter, which is the user's email, and a `rememberMe` property, which is a boolean in our case will always be `true`.

The response object will include a remediation with the `select-authenticator-authenticate` name. This includes a list of authenticators that the user can authenticate with by calling the `/challenge` endpoint. This includes the `email` authenticator, which is required for OTP authentication, and the `password` authenticator, for authentication with a password.

### `/challenge`

Used to initiate a authentication request for a given authenticator.

Takes the `stateHandle`. Also takes an `authenticator` object, which is used to identify the authenticator to use. This has the `id` of the authenticator, and a `methodType` parameter, which is either `email` or `password`.

The response includes a remediation with the `challenge-authenticator` name, which has information on how to verify the given authenticator, which in our case would involve sending a `credentials` object which includes the `passcode` value (which is the value of the authenticator, whether thats a passcode or password), and the state handle, to the `/challenge/answer` endpoint, which has previously been implemented. 

### `/recover`

This endpoint becomes available if selecting the `password` authenticator after calling the `/identify` endpoint.

This is used in order to initiate a password reset flow.

It just takes the `stateHandle` in the body.

The response will include a `authenticator-verification-data` remediation name. Like `/identify` it is a list of authenticators which can be used to verify that the user is able to perform this request. In this case the only available authenticator will be `email`. As with `/identify` the `/challenge` endpoint must be called with the given authenticator.

### DIFF
--- a/src/server/lib/okta/idx/challenge.ts
+++ b/src/server/lib/okta/idx/challenge.ts
@@ -6,8 +6,8 @@ import {
 	idxBaseResponseSchema,
 	idxFetch,
 	idxFetchCompletion,
+	selectAuthenticationEnrollSchema,
 } from './shared';
-import { selectAuthenticationEnrollSchema } from './enroll';
 import { ResponseWithRequestState } from '@/server/models/Express';
 import { validateEmailAndPasswordSetSecurely } from '@/server/lib/okta/validateEmail';
 import { logger } from '@/server/lib/serverSideLogger';
--- a/src/server/lib/okta/idx/challenge.ts
+++ b/src/server/lib/okta/idx/challenge.ts
@@ -1,7 +1,9 @@
 import { z } from 'zod';
 import {
+	AuthenticatorBody,
 	IdxBaseResponse,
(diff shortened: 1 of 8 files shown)


AshCorr (approved):
👌 nice work!

PR #2837, opened 2024-08-09
Dependabot | Group dependency updates

coldlink:
## What does this change?

- Group dependency updates so that it groups together minor and patch updates.
	- Hopefully means less dependency PR spam

- Adds cdk dependency updates

- Run all dependency updates every Monday at 8AM, so they should be ready to review by the time we start work.

- Major dependency updates will have to be manually updated, so this should still be kept and eye on.


AshCorr (approved):
Nice!

PR #2844, opened 2024-08-09
Keep Gateway Clean 🚮 | Add `knip` to replace `ts-prune`

coldlink:
## What does this change?

We used [`ts-prune`](https://github.com/nadeesha/ts-prune) to find unused exports and code within Gateway. However this was only run occasionally, and manually. This has proved helpful in keeping Gateway relatively clean, especially when we forget to remove or update code after adding/removing/refactoring features.

However `ts-prune` has also now been deprecated.

They recommend using [Knip](https://knip.dev/) as a replacement, which self describes as "Find unused files, dependencies and exports in JavaScript and TypeScript projects".

This PR does just that by replacing `ts-prune` with `knip`.

Use `pnpm run knip` to run knip and find any issues it encounters.

However we don't just stop with a like for like replacement.

Firstly `knip` is more configurable, so we set up a `knip.ts` file that determines which files we want to check, and which things we want to ignore. We also configure it to run in `warn` mode by default, so it doesn't exit with an error code if it finds any issues.

Then we cleanup any issues that running `knip` encountered (will add this to `.git-blame-ignore-revs` after merging): [`4aa10a3` (#2844)](https://github.com/guardian/gateway/pull/2844/commits/4aa10a376d53cffce5f87206b3113c617c058721)

Finally it was added to the `pnpm run test` command, so that we'll see any issues when we run tests, but it won't error if it finds issues, just displays them:

```
> gateway@1.0.0 knip /home/runner/work/gateway/gateway
> knip

Unused files (2)
src/server/lib/okta/idx/identify.ts
src/server/lib/okta/idx/recover.ts
Unused exports (2)
challenge                       unknown  src/server/lib/okta/idx/challenge.ts:39:14
selectAuthenticatorValueSchema  unknown  src/server/lib/okta/idx/shared.ts:62:14
```


AshCorr (dismissed):

src/server/models/okta/User.ts
```diff
@@ -117,7 +117,6 @@ export enum Status {
 	PROVISIONED = 'PROVISIONED',
 	RECOVERY = 'RECOVERY',
 	PASSWORD_EXPIRED = 'PASSWORD_EXPIRED',
-	PASSWORD_RESET = 'PASSWORD_RESET',
```
AshCorr, line 120: If this enum doesn't match the return results of the Okta API could we potentially get a type error somewhere if a user is in a `PASSWORD_RESET` state?

AshCorr (approved):

src/server/models/okta/User.ts
coldlink: thankfully the enum isn't used directly by zod, so it shouldn't lead to a type error somewhere

we have error handling which handles cases where the user is in an unknown user state, e.g. 

https://github.com/guardian/gateway/blob/4b58d6ee26ae1c01ba9d66001899167bd49cc4bf/src/server/controllers/sendChangePasswordEmail.ts#L347-L350

PR #2852, opened 2024-08-13
Passwordless | Passcodes for reset password - ACTIVE users

coldlink:
## What does this change?

This PR sets up the first bit of functionality to allow some readers in the ACTIVE state, to reset their password using a passcode sent to their email instead of a link.

If the user is ACTIVE, then they'll be in one of 3 states:
1. ACTIVE users - has email + password authenticator (okta idx email verified)
2. ACTIVE users - has only password authenticator (okta idx email not verified)
3. ACTIVE users - has only email authenticator (SOCIAL users - no password, or passcode only users (not implemented yet))

This PR adds in the ability for users in states 1 and 3 to reset their passwords using a passcode.

This functionality has been added behind the `usePasscodesResetPassword` feature flag.

This allows us to merge this into production and iron out any bugs we find while developing this feature for the other user states. The flag will eventually be removed when all user states for reset password have been migrated to passcodes.

The general IDX API flow for users in these state is:

1. Call the `POST /oauth2/<custom_auth_server>/v1/interact` endpoint  
	- Same parameters auth an authorization code flow call (`/authorize` endpoint)   
		- We only want to allow the `profile` application to do this (this is the gateway one)  
		- And maybe the sample application for testing  
	- This returns a single key in the response body \- `interaction_handle`  
2. Call the `POST /idp/idx/introspect` endpoint with the `interaction_handle` in the post body  
	- This returns a number of things in the response body, effectively all the information needed to generate a login form  
	- The only thing we need from this is the `stateHandle` key, which identifies the current authentication request 
		- There is an `expiresAt` key here, which initially is set to 2 hours in the future  
		- The other things here include information on how to generate a login form, and the social options (and links) available to do this  
	- You can also call this endpoint at any other time with the `stateHandle` in the body to get the current transaction state to see if it’s valid
3. Call the `POST /idp/idx/identify` endpoint with the `email`, `rememberMe=true`, and `stateHandle` in the body  
	- This returns an object very similar to the `/introspect` endpoint but with different things 
		- The `expiresAt` key has now changed to 5 mins 
	- The `remediation` key has everything in it relating to how to resolve the current request 
	- This also lists which authenticators the user has available, and lets us identify the users in each given state
		- If there is both the `password` and `email` authenticator, then the user is in state 1
			- Continue to the next step 
		- If there is only a `password` authenticator, then the user is in state 2
			- This will be implemented in a future PR 
		- If there is only an `email` authenticator then the user is in state 3
			- Set a placeholder password for the user, this will give them the `password` authenticator, and start again from step 1
	- In `remediation.value[1].value` array, there should be 2 authenticators listed, `Email` and `Password`
		- To reset password be need to first select `Password` option  
4. Call `POST /idp/idx/challenge` with `stateHandle` `authenticator:methodType=password,id=password_authenticator_id`  
	- Normally a form to submit a password, however we want to call the recover option  
5. Call `POST /idp/idx/recover` with `stateHandle`  
	- Remediation should have `authenticator-verification-data` with authenticator type `Email`  
6. Call `POST /idp/idx/challenge` with `stateHandle` `authenticator:methodType=email,id=email_authenticator_id`  
	- User should be sent a passcode, recover email  
7. Call the `POST /idp/idx/challenge/answer` with the following shape  
	  - `{“credentials”: { “passcode”: “<passcode>” }, “stateHandle”: “<state_handle>” }` as request body  
	  - Remediation with `reset-authenticator` with how to set a new password, and optionally revoke sessions  
8. Call the `POST /idp/idx/challenge/answer` with the following shape  
	  - `{“credentials”: { “passcode”: “<password>” }, “stateHandle”: “<state_handle>” }` as request body  
	  - This authenticates the code, if valid returns a 200 with the following  
	      - A basic `user` object  
	      - `Set-cookie` with `idx` cookie and value, set this  
9. Redirect the user to `/login/token/redirect?stateToken=${stateHandle.split('~')[0]}`  
	  - The `idx` cookie returned in the previous call doesn’t set a global session, you have to redirect the user to this endpoint for it to be enforced  
	  - The `stateToken` is the `stateHandle` everything before the first `~` character  
	  - This will redirect the user to the callback defined in the `interact` call at the start and completes the interaction code flow

Steps 1 and 2 are provided by the `startIdxFlow` method.

Steps 3 - 6 are provided by the new `changePasswordEmailIdx` method.

Step 7 is `submitPasscode` inside the new `POST /reset-password/code` method 

Steps 8-9 is already provided for us the registration passcode flow, with the exception we add a check for the `reset-authenticator` inside the introspect response to make sure the user is in the correct state.

Here's more information about each individual commit in detail:

---

**Update `changePassword.ts`/`checkPasswordToken.ts` for password (re)set with passcodes**

We update the IDX API handlers in both files to handle the case for password (re)set, which is to add a check for the query parameter (`usePasscodesResetPassword`) and `validateIntrospectRemediation` for the `reset-authenticator` name.

---

**Add `/reset-password/code` `GET` and `POST` methods to handle passcodes**

`GET /reset-password/code` - Essentially the email sent page, but for when using passcodes. Users can end up at this URL after calling `POST /reset-password/code` and the code was incorrect/there was an error.

`POST /reset-password/code` - Endpoint to handle passcode submitted from the email sent page. It tries to submit the passcode and redirect the user to the correct page to set or reset their password.

---

**Add `POST /reset-password/code/resend` to handle resend functionality**

Simple endpoint that just reruns the `sendEmailInOkta` function when the user clicks resend, no need for anything fancy here.

---

**Add `changePasswordEmailIdx` to `sendChangePasswordEmail.ts`**

This set's up the functionality to send an email with a passcode inside the `sendChangePasswordEmail` which is used for all things password reset related.

The `changePasswordEmailIdx` starts the Okta IDX flow and goes through all the steps necessary to send the user an email with a passcode. Currently it only checks for the user with `Status.ACTIVE` and `user.credentials.password` (user has a password). If a user is not in this state it will revert to the legacy password reset flow.

The steps and api calls this function makes are described above.

It will redirect the user to the email sent page with a passcode input if successful, or give an error and revert to the legacy flow.

---

**Add idx/passcode usage metrics for `changePasswordMetric`**

`changePasswordMetric` was only used for the legacy flow, this commit updates this functionality so that we can use this for passcodes also and provide a slightly different metric

---

**Add cypress tests**

As it says on the tin, adds cypress tests to test reset password with passcodes for ACTIVE users with password.

---

**Add passcode reset for users with only email authenticator (i.e SOCIAL users)**

If a user has only the email authenticator, and no password, we set a placeholder password for these users, and then rerun the `changePasswordEmailIdx` method to send a passcode to these users.

---

## How to test

- Deploy this PR to CODE.
- Go to `https://profile.code.dev-theguardian.com/reset-password?usePasscodesResetPassword=true`
- Go through the reset password process, depending on the user type you'll get a different journey
  - In this PR if you want to use passcodes you need to do reset password for an already ACTIVE user with password
  - The easiest way to get a user into this state is to create a new user and go through the whole flow, once you've signed in, sign out, and then do the reset password process

## Tested?

- [x] CODE - ACTIVE user with password, email factors - with flag - passcodes
- [x] CODE - ACTIVE user with password, email factors - without flag - no passcodes
- [x] CODE - ACTIVE user only email factor - with flag - passcodes
- [x] CODE - ACTIVE user only email factor - without flag - no passcodes
- [x] CODE - ACTIVE user only password factor - with flag - no passcodes
- [x] CODE - ACTIVE user only password factor - without flag - no passcodes
- [x] CODE - Other states - with flag - no passcodes
- [x] CODE - Other states - without flag - no passcodes

### DIFF
--- a/src/server/controllers/changePassword.ts
+++ b/src/server/controllers/changePassword.ts
@@ -95,6 +95,12 @@ const oktaIdxApiPasswordHandler = async ({
 					introspectResponse,
 					'enroll-authenticator',
 				);
+			} else {
+				// if we're setting a password, we should be in the reset-authenticator remediation
+				validateIntrospectRemediation(
+					introspectResponse,
+					'reset-authenticator',
+				);
 			}
 
 			// validate the password field
@@ -166,9 +172,14 @@ export const setPasswordController = (
 		const { clientId, useOktaClassic } = res.locals.queryParams;
 
 		// OKTA IDX API FLOW
-		// If the user is using the passcode registration flow, we need to handle the password change/creation.
+		// If the user is using the passcode flow for registration/reset password,
+		// we need to handle the password change/creation.
 		// If there are specific failures, we fall back to the legacy Okta change password flow.
-		if (passcodesEnabled && !useOktaClassic && path === '/welcome') {
+		if (
+			passcodesEnabled &&
+			!useOktaClassic &&
+			(path === '/welcome' || res.locals.queryParams.usePasscodesResetPassword)
+		) {
 			await oktaIdxApiPasswordHandler({
 				req,
 				res,
--- a/src/server/controllers/changePassword.ts
+++ b/src/server/controllers/changePassword.ts
@@ -115,6 +115,9 @@ const oktaIdxApiPasswordHandler = async ({
 				});
 			}
 
+			// track the password change metric
+			trackMetric(changePasswordMetric(path, 'Success', true));
+
 			// Set the password in Okta, redirect the user to set a global session, and then complete
 			// the interaction code flow, eventually redirecting the user back to where they need to go.
 			return await setPasswordAndRedirect({
@@ -133,6 +136,8 @@ const oktaIdxApiPasswordHandler = async ({
 		logger.error('Okta IDX setPassword failure', error, {
 			request_id: state.requestId,
 		});
+
+		trackMetric(changePasswordMetric(path, 'Failure', true));
 
 		if (error instanceof OAuthError) {
 			// case for session expired
@@ -276,7 +281,7 @@ export const setPasswordController = (
 					);
 				}
 
-				changePasswordMetric(path, 'Success');
+				trackMetric(changePasswordMetric(path, 'Success'));
 
 				return await performAuthorizationCodeFlow(req, res, {
 					sessionToken,
@@ -300,7 +305,7 @@ export const setPasswordController = (
 				request_id: res.locals.requestId,
 			});
 
-			changePasswordMetric(path, 'Failure');
+			trackMetric(changePasswordMetric(path, 'Failure'));
 
 			// see the comment above around the success metrics
 			if (clientId === 'jobs' && path === '/welcome') {
--- a/src/server/controllers/checkPasswordToken.ts
+++ b/src/server/controllers/checkPasswordToken.ts
@@ -138,12 +138,18 @@ const oktaIdxApiCheckHandler = async ({
 				req.ip,
 			);
 
+			// validate the introspect response to make sure we're in the correct state
+			// if we're creating a new user (/welcome) we should be in the enroll-authenticator remediation
 			if (path === '/welcome') {
-				// check if the remediation array contains a "enroll-authenticator"	object
-				// if it does, then we know the stateHandle is valid and we're in the correct state
 				validateIntrospectRemediation(
 					introspectResponse,
 					'enroll-authenticator',
+				);
+			} else {
+				// if we're setting a password, we should be in the reset-authenticator remediation
+				validateIntrospectRemediation(
+					introspectResponse,
+					'reset-authenticator',
 				);
 			}
 
@@ -198,13 +204,13 @@ export const checkTokenInOkta = async (
 	const { token } = req.params;
 
 	// OKTA IDX API FLOW
-	// If the user is using the passcode registration flow, we need to check if the user is
-	// in the correct state to be able to change their password.
+	// If the user is using the passcode flow for registration/reset password,
+	// we need to handle the password change/creation.
 	// If there are specific failures, we fall back to the legacy Okta change password flow.
 	if (
 		passcodesEnabled &&
 		!res.locals.queryParams.useOktaClassic &&
-		path === '/welcome' // only check the state handle for registration passcode flow on the welcome page
+		(path === '/welcome' || state.queryParams.usePasscodesResetPassword)
 	) {
 		await oktaIdxApiCheckHandler({
 			path,
--- a/src/server/routes/resetPassword.ts
+++ b/src/server/routes/resetPassword.ts
@@ -8,7 +8,17 @@ import { rateLimitedTypedRouter as router } from '@/server/lib/typedRoutes';
 import handleRecaptcha from '@/server/lib/recaptcha';
 import { sendEmailInOkta } from '@/server/controllers/sendChangePasswordEmail';
 import { mergeRequestState } from '@/server/lib/requestState';
-import { handleAsyncErrors } from '../lib/expressWrappers';
+import { handleAsyncErrors } from '@/server/lib/expressWrappers';
+import { addQueryParamsToPath } from '@/shared/lib/queryParams';
+import {
+	readEncryptedStateCookie,
+	updateEncryptedStateCookie,
+} from '@/server/lib/encryptedStateCookie';
+import { redirectIfLoggedIn } from '@/server/lib/middleware/redirectIfLoggedIn';
+import { convertExpiresAtToExpiryTimeInMs } from '@/server/lib/okta/idx/shared/convertExpiresAtToExpiryTimeInMs';
+import { submitPasscode } from '@/server/lib/okta/idx/shared/submitPasscode';
+import { handlePasscodeError } from '@/server/lib/okta/idx/shared/errorHandling';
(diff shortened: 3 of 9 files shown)


src/server/controllers/changePassword.ts
```diff
@@ -95,6 +95,12 @@ const oktaIdxApiPasswordHandler = async ({
 					introspectResponse,
 					'enroll-authenticator',
 				);
+			} else {
+				// if we're setting a password, we should be in the reset-authenticator remediation
+				validateIntrospectRemediation(
+					introspectResponse,
+					'reset-authenticator',
+				);
```
AshCorr, line 103:
```js
validateIntrospectRemediation(
	introspectResponse,
	path === '/welcome' ? 'enroll-authenticator' : 'reset-authenticator',
);
```

The functions are getting pretty massive and hard to keep it all in my head, might make things easier if we reduced the size of the branches?

non blocking

src/server/controllers/checkPasswordToken.ts
```diff
@@ -138,13 +138,19 @@ const oktaIdxApiCheckHandler = async ({
...
 			if (path === '/welcome') {
-				// check if the remediation array contains a "enroll-authenticator"	object
-				// if it does, then we know the stateHandle is valid and we're in the correct state
 				validateIntrospectRemediation(
 					introspectResponse,
 					'enroll-authenticator',
 				);
+			} else {
+				// if we're setting a password, we should be in the reset-authenticator remediation
+				validateIntrospectRemediation(
```
AshCorr, line 150: same as above suggestion

src/server/controllers/changePassword.ts
```diff
@@ -109,6 +115,9 @@ const oktaIdxApiPasswordHandler = async ({
 				});
 			}
 
+			// track the password change metric
+			changePasswordMetric(path, 'Success', true);
```
AshCorr, line 119: Should something be done with the return value of this function? Unless I'm missing something it seems like all this function does is return a string?

src/server/routes/resetPassword.ts
```diff
@@ -90,6 +100,117 @@ router.get(
...
+	(req: Request, res: ResponseWithRequestState) => {
+		const state = res.locals;
+
+		const encryptedState = readEncryptedStateCookie(req);
+
+		if (encryptedState?.email && encryptedState.stateHandle) {
+			const html = renderer('/reset-password/email-sent', {
+				requestState: mergeRequestState(state, {
+					pageData: {
+						email: readEncryptedStateCookie(req)?.email,
```
AshCorr, line 115:
```suggestion
						email: encryptedState.email,
```

src/server/routes/resetPassword.ts
```diff
@@ -90,6 +100,117 @@ router.get(
...
+			addQueryParamsToPath('/reset-password', res.locals.queryParams),
+		);
+	}),
+);
+
+// Route to resend the email for passcode reset password
+// Essentially the same as POST /reset-password, just start the process again
+router.post(
+	'/reset-password/code/resend',
+	redirectIfLoggedIn,
```
AshCorr, line 208: Do we have a way of preventing spam here? this endpoint feels a bit abusable!

AshCorr (approved):

src/server/controllers/changePassword.ts
coldlink: these are fire and forget promises: https://github.com/guardian/gateway/blob/main/src/server/lib/trackMetric.ts#L24

so we don't care if they succeed/fail and they don't throw either

src/server/controllers/changePassword.ts
coldlink: i might have to more refactors before the next feature is implemented, so i'll keep this in mind!

src/server/routes/resetPassword.ts
coldlink: Ah we have recaptcha protection on the email sent page, and the functionality for resend on the non-passcode flow uses recaptcha for this.

So I'll add that in!

src/server/controllers/changePassword.ts
AshCorr: Shouldn't it be something like this instead then?
 
```suggestion
			trackMetric(changePasswordMetric(path, 'Success', true));
```

AshCorr: As `[changePasswordMetric](https://github.com/guardian/gateway/blob/74e80ef6a07113f143cc2da9a73fed54d1c9e1ff/src/server/models/Metrics.ts#L93-L105)` doesn't seem to actually call `trackMetric` itself.

src/server/controllers/changePassword.ts
AshCorr: Gotcha!

src/server/controllers/changePassword.ts
coldlink: Ah yes!

Very good spot, it should be that, it's what we do elsewhere e.g. for `emailSendMetric`

Will update

src/server/routes/resetPassword.ts
coldlink: Oh and it shouldn't have `redirectIfLoggedIn` anyway, as we do allow logged in users to reset their password, so if this was kept in, then users who were logged in that clicked resend would be redirected to the reauthenticate page, which is not what we want!

PR #2854, opened 2024-08-15
Passwordless | Refactoring multiple Okta IDX API related things

coldlink:
## This PR

This PR refactors, fixes, and adds things related to the Okta IDX API, especially stuff that will be used when we implement passcodes for reset password and sign in.

It's best to review this PR commit by commit, as each one is a separate change.

This description describes each in order of commit.

This has been tested in CODE.

## Changes

---

#### Add helper `ExtractLiteralRemediationNames` type
**refactor**
*commit [0c9acd5](https://github.com/guardian/gateway/pull/2854/commits/0c9acd59bb8ebdac332947e04657501678845b26)*

This is used to extract all possible remediation `name` values into a template literal type containing all the possible values except `string`.

For example:

```ts
// Type to extract all the remediation names from the introspect response
type IntrospectRemediationNames = ExtractLiteralRemediationNames<
	IntrospectResponse['remediation']['value'][number]
>;
```

Will evaluate to:

```ts
type IntrospectRemediationNames = "redirect-idp" | "select-enroll-profile" | "identify"
```

Which is helpful as it means we can avoid using the `string` type when validating the remediation

---

#### Make `IntrospectResponse` remediation a union of all other remediations
**refactor**
*commit [40aa5cc](https://github.com/guardian/gateway/pull/2854/commits/40aa5ccfe17fd0d46b71fb45a5f9a242fd3e018a)*

Since the `introspect` call can be called at any point in the IDX flow to check the current state of the flow, the response returned by the introspect call could be the same as any other IDX API call.

This means that the remediation object could be any remediation object from the other IDX API calls.

This commit updates the remediation object inside the `IntrospectResponse` to be a union of all the other remediation objects from the other IDX API calls.

---

#### Add `reset-authenticator` remediation to `challengeAnswerRemediations`
**feature**
*commit [ed2bb97](https://github.com/guardian/gateway/pull/2854/commits/ed2bb97cc9dc096e4d7e90812963ff919b3a8676)*

This is used by password reset with passcodes. This remediation was missed out in the updated API setup in https://github.com/guardian/gateway/pull/2833

---

#### Add `validateChallengeAnswerRemediation` method
**feature**
*commit [7029c01](https://github.com/guardian/gateway/pull/2854/commits/7029c015f4f79947599d9631a3e155a5a55a773d)*

Validates that the `challenge/answer` response contains a remediation with the given name, throwing an error if it does not. This is useful for ensuring that the remediation we want to perform is available in the challenge/answer response, and the current state is correct.

---

#### Split idx `shared.ts` file into `shared` folder
**refactor**
*commit [5eacb0a](https://github.com/guardian/gateway/pull/2854/commits/5eacb0a61fe07029461887f784b2e26a59f2b3a8)*

The `shared.ts` file was getting very large and unmanageable with all the different things it was doing.

This commit splits the file up into files inside a `shared` folder, each with an individual task making it easier to reason about.

- `convertExpiresAtToExpiryTimeInMs.ts`
  - Just the `convertExpiresAtToExpiryTimeInMs` function
- `idxFetch.ts`
  - Handles all the calls to and responses/errors from the IDX API
- `paths.ts`
  - A list of all the IDX API paths/endpoints
- `schemas.ts`
  - Shared schemas for that all the different IDX API endpoints can inherit from 

---

#### Move `submitPasscode` functionality into shared method
**refactor**
*commit [d7504a0](https://github.com/guardian/gateway/pull/2854/commits/d7504a05cd89bcf8db6bef13818d9d169e949385)*

Create a `submitPasscode` method, and add it to shared folder. This repeating block of code will be repeatedly used to verify passcodes so having a helper method that manages it will prove helpful.

---

#### Create findAuthenticatorId helper method
**refactor**
*commit [d401d12](https://github.com/guardian/gateway/pull/2854/commits/d401d1281a152fd1cf71e6530f11144db2484110)*

To extract a given authenticators id from within the remediation object is a bit of a mess.

Since this functionality will be needed quite often, I've moved this to it's own method.

This method handles all the current possible scenarios where we'll need to find authenticator ids.

Currently in the create account flow it's using the `select-authenticator-enroll` remediation to find the authenticator id for the `password` authenticator.

When we start doing passwordless we'll need more options here. In this commit I've also added the two that are required for password reset using passcodes, specifically `select-authenticator-authenticate` and `authenticator-verification-data`.

Since the shape to find the id can sometimes be different, this also takes into account the possible schema shapes too, so we can hid the complexity from the usages.

---

#### Move passcode error handing into a `handlePasscodeError` function
**refactor**
*commit [57d4da6](https://github.com/guardian/gateway/pull/2854/commits/57d4da6b9557e019a94866349872011f843fee03)*

The passcode error handling will be similar among all routes that will need it. So move it into it's own method.

Any following errors would need to check for `res.headersSent` in case we redirected or rendered a page inside this method.

---

### DIFF
--- a/src/server/lib/okta/idx/shared/schemas.ts
+++ b/src/server/lib/okta/idx/shared/schemas.ts
@@ -1,0 +1,108 @@
+import { z } from 'zod';
+import { logger } from '@/client/lib/clientSideLogger';
+import { trackMetric } from '@/server/lib/trackMetric';
+
+// Schema to check the version of the IDX API, and warn if it's not 1.0.0
+const idxVersionSchema = z.string().refine((val) => {
+	// warn if the version is not 1.0.0
+	if (val !== '1.0.0') {
+		logger.warn('Okta IDX - unexpected version:', val);
+		trackMetric('OktaIDX::UnexpectedVersion');
+	}
+
+	// but pass validation regardless
+	return true;
+});
+
+// Base schema for the IDX API response, everything should inherit from this (using .merge)
+export const idxBaseResponseSchema = z.object({
+	version: idxVersionSchema,
+	stateHandle: z.string(),
+	expiresAt: z.string(),
+});
+export type IdxBaseResponse = z.infer<typeof idxBaseResponseSchema>;
+
+// Base schema for the remediation object in the IDX API response
+// Used in cases where the Okta IDX API during the authentication process
+export const baseRemediationValueSchema = z.object({
+	name: z.string(),
+	type: z.string().optional(),
+	href: z.string().url().optional(),
+	method: z.string().optional(),
+	value: z.unknown().optional(),
+});
+
+// Base type for the body of a request to the IDX API
+// Which should include the stateHandle, and any other data needed using the generic type
+export type IdxStateHandleBody<T = object> = T & {
+	stateHandle: IdxBaseResponse['stateHandle'];
+};
+
+// schema for the select-authenticator-{enroll|authenticate}
+export const selectAuthenticatorValueSchema = z.array(
+	z.union([
+		z.object({
+			name: z.literal('authenticator'),
+			type: z.string(),
+			options: z.array(
+				z.object({
+					label: z.string(),
+					value: z.object({
+						form: z.object({
+							value: z.array(
+								z.object({
+									name: z.enum(['id', 'methodType']),
+									value: z.string(),
+								}),
+							),
+						}),
+					}),
+				}),
+			),
+		}),
+		z.object({
+			name: z.literal('stateHandle'),
+		}),
+	]),
+);
+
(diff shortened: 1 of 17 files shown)


src/server/lib/okta/idx/shared/schemas.ts
```diff
@@ -0,0 +1,108 @@
...
+}>;
+
+// Type to extract all the `name` properties from the remediation value array
+export type ExtractLiteralRemediationNames<T> = T extends { name: infer N }
+	? N extends string
+		? string extends N
+			? never
+			: N
+		: never
+	: never;
```
pvighi, line 108: 🤯

pvighi (approved):
looks good to me assuming it all works in CODE.

PR #2864, opened 2024-08-19
Dependencies | Update Major/Breaking Dependencies | 19th July 2024

coldlink:
## What does this change?

We started using dependabot grouping in https://github.com/guardian/gateway/pull/2837 to update minor/patch dependencies. However this does not update major dependencies, a manual process is still needed for those.

This PR does just that.

It updates:

- `@guardian/eslint-config-typescript`
  - `11.0.0` -> `12.0.0`
- `@guardian/source`
  - `6.1.0` -> `7.0.1`
  - Had to replace the deprecated and now removed typography api with the new typography presets instead, and matched them up to what we had previously where possible
  - https://github.com/guardian/csnx/releases/tag/%40guardian%2Fsource%407.0.0
  - Also had to fix padding on the `CheckboxInput` component as we override Source styles for our own input.
- `@guardian/source-development-kitchen`
  - `6.0.0` -> `7.0.1`

Holds back the following:

- Sentry to v8 - major version upgrade, need to do in a separate PR

## Tested

- [x] CODE


akinsola-guardian (approved):
LGTM

PR #2865, opened 2024-08-20
Passwordless | Refactoring check/change password handlers

coldlink:
## What does this change?

In `changePassword.ts` and `checkPasswordToken.ts` the Okta IDX code is inline with the legacy code.

This isn't ideal for adding new features/maintenance.

This PR moved the Okta IDX API code in each file to it's own methods, `oktaIdxApiPasswordHandler` and `oktaIdxApiCheckHandler` respectfully for `changePassword` and `checkPasswordToken`.

This splits out the Okta IDX code into it's own block, making it easier to manage, especially when we come to add reset password using passcodes.

## Tested

- [x] CODE


akinsola-guardian (approved):
LGTM!

PR #2876, opened 2024-08-29
display request id on unexpected errors

pvighi:
## What does this change?

Display request id on all unexpected errors. Expected errors like 'invalid email or password' should should not display the request id

## Images
Unexpected errors are displayed in 2 different places
in error summaries:
image

in full page errors:
image

## Accessibility


src/client/components/GatewayErrorSummary.tsx
```diff
@@ -0,0 +1,41 @@
...
+				]
+			: props.context;
+
+	const updatedProps = {
+		...props,
+		context: fullContext,
+		message: structuredError?.message,
+		cssOverrides: errorMessageStyles,
+	};
+	return <ErrorSummary {...updatedProps}> </ErrorSummary>;
```
coldlink, line 40: `updatedProps` has no type definition, its preferable to set them directly on the component

```suggestion
	return <ErrorSummary {...props} context={fullContext} message={structuredError?.message} cssOverrides={errorMessageStyles}> </ErrorSummary>;
```

src/client/components/GatewayErrorSummary.tsx
```diff
@@ -0,0 +1,41 @@
...
+	errorContextSpacing,
+	errorMessageStyles,
+} from '@/client/styles/Shared';
+
+export interface GateWayErrorSummaryProps
+	extends Omit<ErrorSummaryProps, 'message' | 'cssOverrides'> {
+	shortRequestId?: string;
+	gatewayError?: GatewayError;
+}
+export default function GatewayErrorSummary(props: GateWayErrorSummaryProps) {
```
coldlink, line 19: In Gateway we generally don't use `default` exports, preferring named exports

```suggestion
export function GatewayErrorSummary(props: GateWayErrorSummaryProps) {
```

or to be even more "Gateway"-like

```suggestion
export const GatewayErrorSummary = (props: GateWayErrorSummaryProps) => {
```

src/client/components/GatewayErrorSummary.stories.tsx
```diff
@@ -0,0 +1,53 @@
...
+const bauError: StructuredGatewayError = {
+	message: 'You entered a field wrong or something',
+	severity: 'BAU',
+};
+
+export const WithStructuredBAUError = () => (
+	<GatewayErrorSummary
+		gatewayError={bauError}
+		context={
+			<p>since this is a BAU error we do not display the request id here</p>
```
coldlink, line 36: These stories probably need to correct styling applied to them

```suggestion
			<p css={[errorContextSpacing, errorContextLastTypeSpacing]}>since this is a BAU error we do not display the request id here</p>
```

src/client/components/GatewayErrorSummary.stories.tsx
```diff
@@ -0,0 +1,53 @@
...
+
+const UnexpectedError: StructuredGatewayError = {
+	message: 'An unexpected error happened unexpectedly',
+	severity: 'UNEXPECTED',
+};
+
+export const WithStructuredUnexpectedError = () => (
+	<GatewayErrorSummary
+		gatewayError={UnexpectedError}
+		context={<p>Since this was not expected we display the request id</p>}
```
coldlink, line 50: And here too

```suggestion
		context={<p css={[errorContextSpacing, errorContextLastTypeSpacing]}>Since this was not expected we display the request id</p>}
```

src/client/components/GatewayErrorSummary.tsx
```diff
@@ -0,0 +1,41 @@
+import React from 'react';
+import {
+	ErrorSummary,
+	ErrorSummaryProps,
+} from '@guardian/source-development-kitchen/react-components';
+
```
coldlink, line 6:
```suggestion
```

src/client/components/GatewayErrorSummary.tsx
coldlink, line 19:
```suggestion
}

export default function GatewayErrorSummary(props: GateWayErrorSummaryProps) {
```

src/client/components/GatewayErrorSummary.tsx
```diff
@@ -0,0 +1,41 @@
...
+export interface GateWayErrorSummaryProps
+	extends Omit<ErrorSummaryProps, 'message' | 'cssOverrides'> {
+	shortRequestId?: string;
+	gatewayError?: GatewayError;
+}
+export default function GatewayErrorSummary(props: GateWayErrorSummaryProps) {
+	const structuredError = asStructuredError(props.gatewayError);
+	const fullContext =
+		props.shortRequestId && structuredError?.severity != 'BAU'
```
coldlink, line 22:
```suggestion
		props.shortRequestId && structuredError?.severity !== 'BAU'
```

src/shared/model/Errors.ts
```diff
@@ -88,3 +88,31 @@ export enum UnsubscribeErrors {
...
+	return error;
+}
+
+// ensures we get a string error message, if the original was a structured error it throws away the rest of the fields
+export function extractMessage(error?: GatewayError): string | undefined {
+	if (typeof error === 'object') {
+		return error.message;
+	}
+	return error;
+}
```
coldlink, line 118: I think it would be easier to check this the other way around, i.e check for a sting first, if it's not a string then it's an object and return the message then.

```suggestion
export function extractMessage(error?: GatewayError): string | undefined {
	if (typeof error === 'string') {
		return error;
	}
	return error.message;
}
```

coldlink (changes requested):

src/shared/model/Errors.ts
pvighi: > I think it would be easier to check this the other way around, i.e check for a sting first, if it's not a string then it's an object and return the message then.
I'm not sure why it's easier to check for string but the only reason I did it this way was because otherwise I have to account for undefined when attempting to return  `error.message`.  I guess I can return `error?.message` instead

coldlink: yeah you'd have to return `error?.message` as `error` might be undefined.

src/server/lib/renderer.tsx
```diff
@@ -140,6 +142,7 @@ export const renderer: <P extends RoutePaths>(
 
 	return `
     <!DOCTYPE html>
+    <!--Request ID: ${clientState.shortRequestId}-->
```
coldlink, line 145: I'd leave a space on either side, so it can't easily be mistake for another tag.
```suggestion
    <!-- Request ID: ${clientState.shortRequestId} -->
```

coldlink (approved):

PR #2877, opened 2024-08-30
Passing query params (ref, refViewId) to subscribe/success

akinsola-guardian:
## What does this change?
This passes the ref, refViewId and listName query parameter to the `/subcribe/success` endpoint. By using the `res.locals.queryParams` in `buildUrlWithQueryParams`

## How to test
This has been tested in CODE in relation to this PR (https://github.com/guardian/identity/pull/2568).


src/server/routes/consentToken.ts
```diff
@@ -26,7 +26,18 @@ router.get(
...
+			return res.redirect(
+				303,
+				buildUrlWithQueryParams(
+					'/subscribe/success',
+					{},
+					{
+						ref: req.query.ref?.toString(),
+						refViewId: req.query.refViewId?.toString(),
+						returnUrl: '',
+					},
```
coldlink, line 38: rather than calling `req.query` directly here, use `res.locals.queryParams`

```suggestion
					res.locals.queryParams,
```

Docs: https://github.com/guardian/gateway/blob/main/docs/development.md#query-params

And info on how to use them server side: https://github.com/guardian/gateway/blob/main/docs/development.md#server

the tl;dr is that we have middleware which parses and checks all query params for us making sure they're all valid, and we make them accessible through the request state `res.locals`, this includes the `ref` and `refViewId` params

coldlink (changes requested):

coldlink (approved):

PR #2881, opened 2024-09-05
Passwordless | `ResetPasswordEmailSentPage` and API refactors

coldlink:
## What does this change?

- Sets up an `ResetPasswordEmailSentPage` component that displays the correct page based on whether we're using passcodes or the legacy link method
  - If we're using passcodes, as defined by `hasStateHandle` variable, then we show the `PasscodeEmailSent` component
    - If the passcode has already been used, then we show the `PasscodeUsed` component
  - If we're not using passcodes, then we show the `EmailSent` component as previously done
- We also move `challengeAuthenticatorSchema` to `shared/schemas.ts` as it's also needed by the `identifyRemediations` type too
- Finally we refactor the `validate<ENDPOINT>Remediation` methods into a single method called `validateRemediation` which takes type generics
  -  Anything that was previously using the full method definition can just use the types instead, e.g. for `validateChallengeAnswerRemediation` it becomes

```ts
export const validateChallengeAnswerRemediation = validateRemediation<
	ChallengeAnswerResponse,
	ChallengeAnswerRemediationNames
>;
```

instead of

```ts
export const validateChallengeAnswerRemediation = (
	challengeAnswerResponse: ChallengeAnswerResponse,
	remediationName: ChallengeAnswerRemediationNames,
) => {
	const hasRemediation = challengeAnswerResponse.remediation.value.some(
		({ name }) => name === remediationName,
	);

	if (!hasRemediation) {
		throw new OAuthError(
			{
				error: 'invalid_request',
				error_description: `Remediation ${remediationName} not found in challenge/answer response`,
			},
			400,
		);
	}
};
```

### DIFF
--- a/src/client/pages/ResetPasswordEmailSentPage.tsx
+++ b/src/client/pages/ResetPasswordEmailSentPage.tsx
@@ -1,0 +1,64 @@
+import React from 'react';
+import useClientState from '@/client/lib/hooks/useClientState';
+import { EmailSent } from '@/client/pages/EmailSent';
+import { PasscodeEmailSent } from '@/client/pages/PasscodeEmailSent';
+import { buildQueryParamsString } from '@/shared/lib/queryParams';
+import { buildUrl } from '@/shared/lib/routeUtils';
+import { PasscodeUsed } from '@/client/pages/PasscodeUsed';
+
+export const ResetPasswordEmailSentPage = () => {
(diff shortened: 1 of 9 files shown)


AshCorr (dismissed):

akinsola-guardian (approved):
LGTM

PR #2882, opened 2024-09-06
Okta | Use `X-Forwarded-For` header to forward user's IP address

coldlink:
## What does this change?

Okta support IP address forwarding if we're running Okta behind an reverse proxy, which we are with Fastly.

https://developer.okta.com/docs/reference/core-okta-api/#ip-address

We can use the `X-Forwarded-For` header to forward the user's IP address to Okta, which should mean that events in the system log, threat detection (based on ip), and rate limiting, should be specific to the user, rather than generic Fastly IP addresses.

This PR adds the IP address and forwards to Okta where possible.

We may also need to make Gateway's public IP addresses in the allowlist in the [okta's network security settings](https://help.okta.com/okta_help.htm?id=ext_Security_Network) as a trusted proxy to forward the user's original IP address with the `X-Forwarded-For` HTTP header: https://github.com/guardian/identity-platform/pull/761

We've also added the IP address to the routing logs, so we can use this to help debug issues by IP address, and tie the request id to an IP address making debugging easier.

e.g.

`{"ip":"::ffff:127.0.0.1","level":"info","message":"GET, /signin","request_id":"283433f8-d6ed-4a5d-9928-a224ba41a75c"}`

## Tested
- [x] DEV
- [x] CODE


src/server/lib/okta/api/headers.ts
```diff
@@ -1,8 +1,19 @@
 import { getConfiguration } from '@/server/lib/getConfiguration';
 
-export const defaultHeaders = {
-	Accept: 'application/json',
-	'Content-Type': 'application/json',
+export const defaultHeaders = (ip?: string) => {
```
coldlink, line 3: This is the main change for the Okta API side, to convert `defaultHeaders` to a function that takes the `ip`

akinsola-guardian (approved):
LGTM.

PR #2889, opened 2024-09-10
Passwordless | Passcodes for reset password - ACTIVE users with only "password" authenticator

coldlink:

The next step was to implement this for users who only had the "password" authenticator, this is being completed in this PR.

Here's a reminder of the possible active user states:
1. ACTIVE users - has email + password authenticator (okta idx email verified)
    - Implemented in https://github.com/guardian/gateway/pull/2852 
2. ACTIVE users - has only password authenticator (okta idx email not verified)
    - Implemented in this PR
    - For a user to get into this state, they would have had to NOT verify the passcode, and then reset their password using the classic flow.
3. ACTIVE users - has only email authenticator (SOCIAL users - no password, or passcode only users (not implemented yet))
    - Implemented in https://github.com/guardian/gateway/pull/2852 

This PR adds the ability for users in state 2 to reset their password with an OTP

For these users the following steps would have to be taken when a user wants to reset their password:

1. Call the `dangerouslySetPlaceholderPassword` method in order to set a placeholder password, this will overwrite any password the user had, but should they complete the flow successfully then that should be fine
2. Call `/identify` and use the placeholder password to authenticate
3. Check that we need to set an `email` authenticator
4. If so call the `/credential/enroll` endpoint to send OTP
5. Have the user input the OTP
6. Verify the OTP, this will put the user into ACTIVE State 1
7. Generate a classic Okta Recovery Token, and redirect the user here to set a password, and sign in

## Tested
- [x] CODE

### DIFF
--- a/src/server/routes/resetPassword.ts
+++ b/src/server/routes/resetPassword.ts
@@ -23,6 +23,8 @@ import {
 	validateChallengeAnswerRemediation,
 } from '@/server/lib/okta/idx/challenge';
 import { OAuthError } from '@/server/models/okta/Error';
+import { forgotPassword } from '@/server/lib/okta/api/users';
+import { buildUrlWithQueryParams } from '@/shared/lib/routeUtils';
 
 // reset password email form
 router.get('/reset-password', (req: Request, res: ResponseWithRequestState) => {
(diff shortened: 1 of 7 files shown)


src/shared/model/EncryptedState.ts
```diff
@@ -13,4 +13,12 @@ export interface EncryptedState {
...
 	// Okta IDX API - Flag to determine if the user has used a passcode
 	passcodeUsed?: boolean;
+	// Okta IDX API - State of the user in the Okta determines if we can send passcodes to the user when resetting the password
+	// 0 - user doesn't exist - not currently used in any code
+	// 1 - ACTIVE - user has email + password authenticator (okta idx email verified)
+	// 2 - ACTIVE - user has only password authenticator (okta idx email not verified)
+	// 3 - ACTIVE - user has only email authenticator (okta idx email verified) - not currently used in any code
+	// 4 - user not in ACTIVE state - not currently used in any code
+	// Only including states that we actually currently set on the EncryptedState cookie
+	userState?: 1 | 2;
```
akinsola-guardian, line 23: Could it be more-readable if it used an enum or set of defined strings? or does this come from Okta?

coldlink: Have updated it to be more clear, see [`c000429` (#2889)](https://github.com/guardian/gateway/pull/2889/commits/c000429a5e375c0ddf4c60983f1a1628c4287ea1)

src/server/routes/resetPassword.ts
```diff
@@ -150,16 +152,58 @@ router.post(
 				const challengeAnswerResponse = await submitPasscode({
 					passcode: code,
 					stateHandle,
-					introspectRemediation: 'challenge-authenticator',
+					introspectRemediation:
+						// if the user is in the `2` state, then they are trying enroll the email authenticator
```
AshCorr, line 156:
```suggestion
						// if the user is in the `ACTIVE_PASSWORD_ONLY` state, then they are trying enroll the email authenticator
```

src/server/routes/resetPassword.ts
```diff
@@ -150,16 +152,58 @@ router.post(
...
-							'Invalid challenge/answer response - got a complete login response',
+					if (encryptedState.userState !== 'ACTIVE_PASSWORD_ONLY') {
+						throw new OAuthError({
+							error: 'invalid_response',
+							error_description:
+								'Invalid challenge/answer response - got a complete login response',
+						});
+					}
+
+					// if the user was in the `2` state and they were trying to reset their password
```
AshCorr, line 177:
```suggestion
					// if the user was in the `ACTIVE_PASSWORD_ONLY` state and they were trying to reset their password
```

AshCorr (dismissed):

AshCorr (approved):

akinsola-guardian (approved):
LGTM

PR #2890, opened 2024-09-11
Use AsyncLocalStorage to enrich logs with Request information

AshCorr:
## What does this change?

Uses Node's somewhat newish [`AsyncLocalStorage`](https://nodejs.org/api/async_context.html#class-asynclocalstorage) (its been Stable since Node 16) to make Request and Response information implicitly available to the logger. For those used to the Scala world this is somewhat like the `implicit` keyword, and for those used to the React world this is somewhat similar to [`useContext`](https://www.w3schools.com/react/react_usecontext.asp) hook.

This allows us to attach information about the request to every log message without having to pass parameters such as `request_id` all the way down the call tree.

In addition to logging the `request_id` I've also started logging the the user IP, which was started in https://github.com/guardian/gateway/pull/2882 but only done on the request logger messages.

## How it works

The way this works is that you create a AsyncLocalStorage instance and invoke the `run` function to create a new "store", any code that is ran inside the run function is then able to access said store.

```js
const state = new AsyncLocalStorage<string>();

state.run("Hello World!", () => {
    setTimeout(() => {
        // Logs "Hello World!"
        console.log(state.getStore());
   }, 5000);
});
```

The coolest feature of `AsyncLocalStorage` is that promises that run asynchronously within the `run` context are also able to access the same context. This is great for us as it means we can create a "store" for each request.

As we use Express all we need to do is create a new middleware which instantiates a `AsyncLocalStorage` and calls `run` with the `next` middleware as the target. Since under the hood Express middlewares form a chain of function calls, as long as 1 middleware has access to the store all of the following middlewares will also get access to the same store.

## What we should probably avoid

Using implicit context can quickly make refactoring and testing a nightmare, as functions that use `context` essentially have hidden parameters now that can make their behaviour unpredictable when not used in the correct scope.

Because of this I suggest we only use `context` for paths that are not business critical, for example logging and _potentially_ error pages/messages (@pvighi could be helpful for some of your work?). 

For example, for Okta IP rate limiting we could get the IP from the request context but since this is a business critical path we should probably just pass the IP down the call tree explicitly to make it clear how and where it is being used.

## How to Test

Ran in CODE.

image

## Previous Work

We previously started using AsyncLocalStorage in DCR for logging: https://github.com/guardian/dotcom-rendering/pull/7670

## How to review

The core part of this PR, where the request context is implemented, can be found in https://github.com/guardian/gateway/pull/2890/commits/eef49518f0602ac970c6dd3e9b2bb62cb75445b0

The following commit is mostly just removing `request_id` from the call tree.

### DIFF
--- a/src/server/lib/middleware/requestContext.ts
+++ b/src/server/lib/middleware/requestContext.ts
@@ -1,0 +1,36 @@
+import { NextFunction, Request, Response } from 'express';
+import { AsyncLocalStorage } from 'node:async_hooks';
+
+type LoggerContext = {
+	request: Request;
+	response: Response;
+};
+
+export const requestContext = new AsyncLocalStorage<LoggerContext>();
+
+/**
+ * This middleware handles creating a new context store for each request.
+ *
+ * I recommend only using this store for for logging and debugging,
+ * as relying on it can quickly spiral into hard to refactor and hard to test
+ * spaghetti code.
+ *
+ * @param request - The Request object to be stored in the Store
+ * @param response - The Response object to be stored in the Store
+ * @param next - The next middleware to be called and wrapped in the Stores scope
+ */
+export const requestContextMiddleware = (
+	request: Request,
+	response: Response,
+	next: NextFunction,
+) => {
+	// Express middlewares form a function call chain under the hood, so wrapping one middleware in a Store
+	// will make it available to all subsequent middlewares in the chain.
+	requestContext.run(
+		{
+			request,
+			response,
+		},
+		next,
+	);
+};
--- a/src/server/lib/middleware/requestContext.ts
+++ b/src/server/lib/middleware/requestContext.ts
@@ -1,12 +1,12 @@
 import { NextFunction, Request, Response } from 'express';
 import { AsyncLocalStorage } from 'node:async_hooks';
 
-type LoggerContext = {
-	request: Request;
-	response: Response;
+type RequestContext = {
+	requestId?: string;
+	ip?: string;
 };
 
-export const requestContext = new AsyncLocalStorage<LoggerContext>();
+export const requestContext = new AsyncLocalStorage<RequestContext>();
 
 /**
  * This middleware handles creating a new context store for each request.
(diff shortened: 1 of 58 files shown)


src/server/lib/middleware/requestContext.ts
```diff
@@ -0,0 +1,36 @@
+import { NextFunction, Request, Response } from 'express';
+import { AsyncLocalStorage } from 'node:async_hooks';
+
+type LoggerContext = {
+	request: Request;
+	response: Response;
+};
```
coldlink, line 7: Rather than storing the entire express request and response. I think it's potentially better to limit what's available here, to avoid misuse further down the line.

If this is being used only for logging at the moment, then it should only include what's relevant for that, namely the `requestId` and `ip`, and we can always update if we need to log anything extra.

src/server/lib/middleware/requestContext.ts
```diff
@@ -0,0 +1,36 @@
...
+
+export const requestContext = new AsyncLocalStorage<LoggerContext>();
```
coldlink, line 9: Whats the maximum storage size of an `AsyncLocalStorage`? Does it keep growing for every request? Or is node smart enough to cleanup/garbage collect once the request is over?

Just worried about this filling up potentially over a period of time, and then the instance crashing and restarting because it was out of memory.

src/server/lib/serverSideLogger.ts
```diff
@@ -25,6 +26,22 @@ class ServerSideLogger extends BaseLogger {
...
+		const context = requestContext.getStore();
+
+		const extraFieldsWithContext = {
+			...extraFields,
+			...(context
+				? {
+						in_request_context: true,
+						request_id:
+							context.request.headers?.['x-request-id'] ?? 'no-request-id',
+						ip: context.request.ip ?? 'no-ip',
```
coldlink, line 38: My personal opinion would be for it to be `undefined` or an empty string if the request id or ip doesn't exist. But this is just preference.

src/server/lib/middleware/requestContext.ts
coldlink: The documentation suggests that it is cleared up when the request and hence function chain is over, which is great!

And also this StackOverflow answer: https://stackoverflow.com/a/78392911

src/server/lib/middleware/requestContext.ts
AshCorr: Mm, yea thats probably a good point!

src/server/lib/serverSideLogger.ts
AshCorr: I'm ambivalent! Lets go with the less code way!

coldlink (approved):

coldlink (approved):

PR #2891, opened 2024-09-11
Passwordless | YMIAR (Yet more IDX API Refactors)

coldlink:
## What does this change?

The next step was to implement this for users who only had the "password" authenticator, this is being completed in https://github.com/guardian/gateway/pull/2889, see more information about those users in that PR.

In order to support some of these features we refactor and add some features in the IDX API implementation in Gateway.

Each commit is independent of each other, and rather than reviewing this PR as a whole, the reviewer is asked to go through the PR commit by commit, and review this PR that way.

This is done to avoid me potentially opening a number of smaller PRs with each individual change, which would be difficult to organise.

Each commit has additional context attached to it, and I've also laid out below what each commit achieves.

---

## Changes

---

**Use correct resendEmailAction**
*Type: Fix*

We were using the wrong resend email action on the reset password page, where clicking "resend" would actually use the "register/create account" (`POST /register/email-sent/resend`) resend action instead of the correct one (`POST /reset-password/resend`)

---

**Update challenge/answer API type/variable naming**
*Type: Refactor*

Updates the variables in `src/server/lib/okta/idx/challenge.ts` related to the `/challenge/answer` endpoint to better reflect the API, and to bring it in line with naming elsewhere in the IDX API implementation within Gateway

---

**Remove `idx` cookie from `idxFetch` methods**
*Type: Refactor*

While testing the IDX API flow, it turns out we don't actually need to get the `idx` cookie from the `idxFetchCompletion` method.

What actually happens is that the `idx` cookie will be automatically set by Okta when we redirect the user to the `/login/token/redirect` endpoint which completes the authentication process.

By removing this code, we can make the `idxFetch` methods easier to reason about, and actually `idxFetchCompletion` is removed in a later commit altogether

---

**Make `challenge/answer` API response either `ChallengeAnswerResponse` or `CompleteLoginResponse`**
*Type: Refactor/Feature*

When using the IDX API, the `challenge/answer` endpoint returns different things depending on the context. This could also potentially include a `CompleteLoginResponse`.

This would happen, for example, if using the IDX API to sign in a user with a password or OTP, where no additional steps are required after calling the `challenge/answer` endpoint.

This is required for https://github.com/guardian/gateway/pull/2889 as we need to authenticate the user with a password in order to determine whether we need to enroll the user in the email authenticator or not.

So this commit refactors usages of the `challengeAnswer` method to make sure we check the correct response type we're expecting in a particular scenario.

---

**Add `submitPassword` method**
*Type: Feature*

This is similar to the `submitPasscode` method, but takes a password instead.

It submit a password to Okta to answer the password challenge using the `challenge/answer` endpoint and return the response.

Validation can be enabled as needed through the params, `validatePasswordLength` and `validateBreachedPassword` (both default to false) where if validation fails then the method will throw.

This is required for https://github.com/guardian/gateway/pull/2889 as we need to authenticate the user with a password in order to determine whether we need to enroll the user in the email authenticator or not.

---

**Add parameter to return placeholder password if true**
*Type: Refactor/Feature*

> [!WARNING]
> Since this is passing around a password, even a placeholder one, care should be taken how it is used. Thankfully the name of the function, `dangerouslySetPlaceholderPassword`, already helps with that, but making the point here too.

In some cases, e.g. for ACTIVE users with only the "password" authenticator (state 2), we actually need to set a placeholder password for these users that is then used to authenticate the user in order for them to be able to verify their email.

To do this we need to return the placeholder password so we can use it in the Okta IDX authentication process.

---

**Add `Authenticators` type**
*Type: Feature*

Adds the `Authenticators` type which can either be `email` or `password`.

This is a helper type, which is used in a few places makes it easier to identify the authenticators that users can currently have.

---

**Add `validateChallengeRemediation` method for the `challenge` API response**
*Type: Feature*

`validateChallengeRemediation` is a function that validates that the challenge response contains a remediation with the given name and authenticator.

This is slightly more complicated compared to the other `validateRemediation` methods, as we don't just want to check if the response has a specific remediation name, we also want to check if it has a specific authenticator, and possibly also a given resend/recover method.

---

**Remove `signInRedirect` and `status` from `EncryptedState` as no longer used and update comments accordingly**
*Type: Refactor*

Clean up the `EncryptedState` cookie by removing unused things, and updating comments to better reflect the usages.

---

**Create `oktaIdxCreateAccount` controller**
*Type: Refactor*

Move the code for creating a user using the Okta IDX API out of the `OktaRegistration` method, and into it's own controller instead.

This makes the code easier to reason about, and separates the IDX flow from the classic flow!

Also leaves a `TODO` about updating this in the future once we have finished developing passcodes for sign in and reset password to come back and make sure that user behaviour is the same regardless of if their a new or existing user when going through the create account process.

---

**Add `validateEnrollRemediation` for `enrollResponse`**
*Type: Feature*

As we've done elsewhere, add an validate remediation for the EnrollResponse, making it easier to check whether we're in the correct state.

---

**Remove `idxFetchCompletion` and update `setPasswordAndRedirect` to use `submitPassword` instead**
*Type: Refactor*

In Okta IDX API, it is difficult to determine the exact scenarios where we get a `CompleteLoginResponse`, and thus when to use the `idxFetchCompletion` method vs the `idxFetch` method.

We previously refactored `challenge/answer` to be either `ChallengeAnswerResponse` or `CompleteLoginResponse`, and added the `isChallengeAnswerCompleteLoginResponse` and `isChallengeAnswerResponse` type guards which should be used to check which response we recieved.

We also made a new `submitPassword` function, which abstracts away everything to do with submiting the password to the Okta API and any additional validation.

We can then update the `setPasswordAndRedirect` method to use `submitPassword` instead of the `idxFetchCompletion` method, and use the new `getLoginRedirectUrl` to determine where the user needs to go to get a global session set after authenticating.

---

### DIFF
--- a/src/server/lib/okta/idx/challenge.ts
+++ b/src/server/lib/okta/idx/challenge.ts
@@ -141,7 +141,7 @@ export type ChallengeAnswerResponse = z.infer<
 >;
 
 // Body type for the challenge/answer request - passcode can refer to a OTP code or a password
-type ChallengeAnswerPasswordBody = IdxStateHandleBody<{
+type ChallengeAnswerBody = IdxStateHandleBody<{
 	credentials: {
 		passcode: string;
 	};
@@ -157,13 +157,13 @@ type ChallengeAnswerPasswordBody = IdxStateHandleBody<{
  * @param ip - The ip address
  * @returns Promise<ChallengeAnswerResponse> - The challenge answer response
  */
-export const challengeAnswerPasscode = (
+export const challengeAnswer = (
 	stateHandle: IdxBaseResponse['stateHandle'],
-	body: ChallengeAnswerPasswordBody['credentials'],
+	body: ChallengeAnswerBody['credentials'],
 	request_id?: string,
 	ip?: string,
 ): Promise<ChallengeAnswerResponse> => {
-	return idxFetch<ChallengeAnswerResponse, ChallengeAnswerPasswordBody>({
+	return idxFetch<ChallengeAnswerResponse, ChallengeAnswerBody>({
 		path: 'challenge/answer',
 		body: {
 			stateHandle,
@@ -220,7 +220,7 @@ export const setPasswordAndRedirect = async ({
 	ip,
 }: {
 	stateHandle: IdxBaseResponse['stateHandle'];
-	body: ChallengeAnswerPasswordBody['credentials'];
+	body: ChallengeAnswerBody['credentials'];
 	expressReq: Request;
 	expressRes: ResponseWithRequestState;
 	path?: string;
@@ -228,7 +228,7 @@ export const setPasswordAndRedirect = async ({
 	ip?: string;
 }): Promise<void> => {
 	const [completionResponse, redirectUrl] =
-		await idxFetchCompletion<ChallengeAnswerPasswordBody>({
+		await idxFetchCompletion<ChallengeAnswerBody>({
 			path: 'challenge/answer',
 			body: {
 				stateHandle,
--- a/src/server/lib/okta/idx/challenge.ts
+++ b/src/server/lib/okta/idx/challenge.ts
@@ -8,6 +8,8 @@ import { setupJobsUserInOkta } from '@/server/lib/jobs';
 import { trackMetric } from '@/server/lib/trackMetric';
 import { sendOphanComponentEventFromQueryParamsServer } from '@/server/lib/ophan';
 import {
+	CompleteLoginResponse,
+	completeLoginResponseSchema,
 	idxFetch,
 	idxFetchCompletion,
 } from '@/server/lib/okta/idx/shared/idxFetch';
@@ -162,14 +164,17 @@ export const challengeAnswer = (
 	body: ChallengeAnswerBody['credentials'],
 	request_id?: string,
 	ip?: string,
-): Promise<ChallengeAnswerResponse> => {
-	return idxFetch<ChallengeAnswerResponse, ChallengeAnswerBody>({
+): Promise<ChallengeAnswerResponse | CompleteLoginResponse> => {
+	return idxFetch<
+		ChallengeAnswerResponse | CompleteLoginResponse,
+		ChallengeAnswerBody
+	>({
 		path: 'challenge/answer',
 		body: {
 			stateHandle,
 			credentials: body,
 		},
-		schema: challengeAnswerResponseSchema,
+		schema: challengeAnswerResponseSchema.or(completeLoginResponseSchema),
 		request_id,
 		ip,
 	});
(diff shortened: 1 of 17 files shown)


src/server/lib/okta/idx/challenge.ts
```diff
@@ -86,6 +87,92 @@ export const challenge = (
...
+ * @name validateChallengeRemediation
+ * @description Validates that the challenge response contains a remediation with the given name and authenticator.
+ * @param challengeResponse - The challenge response
+ * @param remediationName - The name of the remediation to validate
+ * @param authenticator - The authenticator type - email or password
+ * @param checkForResendOrRecover - Whether to check for resend (email) or recover (password) functionality depending on the authenticator
+ * @param useThrow - Whether to throw an error if the remediation is not found (default: true)
+ * @returns boolean - Whether the remediation was found in the response
+ */
+export const validateChallengeRemediation = (
```
AshCorr, line 108: Some tests for these `validate` functions might be a good call at some point!

src/server/routes/register.ts
AshCorr, line 1: Love the separation!

src/server/routes/register.ts
```diff
@@ -337,153 +353,179 @@ export const OktaRegistration = async (
...
-			// check if we have the `select-enroll-profile` remediation property which means registration is allowed
-			const introspectEnrollProfileCheck =
-				introspectResponse.remediation.value.find(
-					({ name }) => name === 'select-enroll-profile',
-				);
+		// check if we have the `select-enroll-profile` remediation property which means registration is allowed
+		const introspectEnrollProfileCheck =
+			introspectResponse.remediation.value.find(
+				({ name }) => name === 'select-enroll-profile',
+			);
```
AshCorr, line 371: Are we able to use `validateRemediation` here?

src/server/routes/register.ts
```diff
@@ -337,153 +353,179 @@ export const OktaRegistration = async (
...
-					// will implement when full passwordless is implemented
-				}
-			}
+	// OKTA IDX API FLOW
+	// Attempt to register the user with Okta using the IDX API
+	// and specifically using passcodes.
+	// If there are specific failures, we fall back to the legacy Okta registration flow
+	if (passcodesEnabled && !useOktaClassic) {
+		// try to start the IDX flow to create an account for the user
+		await oktaIdxCreateAccount(req, res);
```
AshCorr, line 523: 👌

AshCorr (approved):

src/server/routes/register.ts
coldlink: Ah it's something I missed! Have added an `validateEnrollRemediation` method.

There was also a place above in the code where I added `validateIntrospectRemediation` too!

PR #2900, opened 2024-09-17
Add listName to queryParams

akinsola-guardian:
## What does this change?

Adds listName as optional PersistableQueryParams to ensure it's passed on via /consent-token/:token/accept to /subscribe/success/.

Example of query param
`/subscribe/success?listName=afternoon-update&ref=https%3A%2F%2Fwww.theguardian.com%2Femail-newsletters&refViewId=m168924jlazsbd612jw1&returnUrl=https%3A%2F%2Fm.code.dev-theguardian.com`

This has been tested locally using Hoppscotch and in CODE.


src/server/lib/queryParams.ts
```diff
@@ -70,6 +71,7 @@ export const parseExpressQueryParams = (
 			componentEventParams || bodyParams.componentEventParams,
 		encryptedEmail,
 		error,
+		listName: listName || bodyParams.listName,
```
coldlink, line 74: Is there any possible scenario that the `listName` would come in a request body (req.body) and not query params?

As I don't think this parameter is set anywhere as a field on a form directly?

```suggestion
		listName: listName,
```

akinsola-guardian: No, it's only coming via query parameters. Thanks for this. This has been updated.

coldlink (approved):

PR #2902, opened 2024-09-18
Passwordless | Passcodes for reset password - non-ACTIVE users

coldlink:
## What does this change?

In https://github.com/guardian/gateway/pull/2889 we implemented passcode for reset password for the remaining active users, i.e the ones with only the "password" authenticator.

This PR adds the ability for all except non-existent users to reset their passcodes using passwords.

For users in a non-`ACTIVE` state, we should first get them into one of the `ACTIVE` states in order to send them an email which allows them to reset their password with a passcode.

The best way to do this is to first `deactivate` the user, which works on all user states and puts them into the `DEPROVISIONED` state.

Then we can activate the user, which will put them into the `PROVISIONED` state and return us a recovery token.

We then use the recovery token to set a placeholder password for the user, which transitions them into the `ACTIVE` state and then we can call `changePasswordEmailIdx` method again to send the user a passcode as they'll be in one of the `ACTIVE` states, namely the one with the "email" and "password" authenticator, or the one with only the "password" authenticator.

From there the existing functionality for `ACTIVE` users takes over and we send the users a passcode!

We also modify the `validateEmailAndPasswordSetSecurely` method to do the exact reverse depending on a flag. The `flagStatus` parameter was added in to set these flags `emailValidated` and `passwordSetSecurely` to `false` if `flagStatus` is provided as `false`. `flagStatus` defaults to `true` to retain existing behaviour.

A whole lotta cypress tests are included to make sure this functionality works.

## Tested
- [x] CODE

### DIFF
--- a/src/server/lib/okta/validateEmail.ts
+++ b/src/server/lib/okta/validateEmail.ts
@@ -10,12 +10,18 @@ import { trackMetric } from '@/server/lib/trackMetric';
  *
  * @param {string} id accepts the Okta user ID, email address (login) or login shortname (as long as it is unambiguous)
  * @param {string} ip The IP address of the user
+ * @param {boolean} flagStatus Default is true. If true, the flags are set to true, if false, the flags are set to false
  * @returns {Promise<UserResponse>} Promise that resolves to the user object
  */
-export const validateEmailAndPasswordSetSecurely = async (
-	id: string,
-	ip?: string,
-): Promise<UserResponse> => {
+export const validateEmailAndPasswordSetSecurely = async ({
+	id,
+	ip,
+	flagStatus = true,
+}: {
+	id: string;
+	ip?: string;
+	flagStatus?: boolean;
+}): Promise<UserResponse> => {
 	try {
 		const timestamp = new Date().toISOString();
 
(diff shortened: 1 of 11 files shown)


pvighi (dismissed):
nice test coverage for all the cases!

pvighi (dismissed):

src/server/lib/okta/validateEmail.ts
```diff
@@ -10,22 +10,33 @@ import { trackMetric } from '@/server/lib/trackMetric';
...
  * @returns {Promise<UserResponse>} Promise that resolves to the user object
  */
-export const validateEmailAndPasswordSetSecurely = async (
-	id: string,
-	ip?: string,
-): Promise<UserResponse> => {
+export const validateEmailAndPasswordSetSecurely = async ({
+	id,
+	ip,
+	unsetFlags = false,
```
AshCorr, line 19:
```suggestion
	setFlags = true,
```

Perhaps just default to true and avoid the [negation on line 30](https://github.com/guardian/gateway/pull/2902/files#diff-8e0a6ee150799c6cc4fd0826332e5b6b5500e0eaef2899cad2c42dccccbc7bc8R30), I find it a little bit confusing 😅

coldlink: Yeah that's a good point! I'll see what I can conjure up.

coldlink: Updated it to `flagStatus` instead, which makes more sense than "set"/"unset"!

AshCorr (approved):

PR #2908, opened 2024-09-23
Dependencies | 23rd September 2024

coldlink:
## What does this change?

Updates dependencies where possible since #2899

Also runs [`knip`](https://knip.dev/) to remove unused exports.
 
Holds back the following:

- Sentry to v8 - major version upgrade, need to do in a separate PR i.e #2903
- Storybook - Issues with preact/react in 8.3.x will investigate separately

## Tested

- [x] CODE


akinsola-guardian (approved):
LGTM

PR #2915, opened 2024-09-24
Passwordless | Reset password with passcodes handle case for non-existent users

coldlink:
## What does this change?

In https://github.com/guardian/gateway/pull/2889 we implemented passcode for reset password for the remaining active users, i.e the ones with only the "password" authenticator.

And in https://github.com/guardian/gateway/pull/2902 we implemented passcode for reset password for existing users in the non-ACTIVE state.

There is one more group of users that could go through the reset password process, and these are users who currently don't have a Guardian account.


src/server/controllers/sendChangePasswordEmail.ts
```diff
@@ -812,6 +812,31 @@ export const sendEmailInOkta = async (
...
 			error.code === 'E0000007'
 		) {
+			// if we're using passcodes, then show the email sent page with OTP input
+			// even if the user doesn't exist
+			if (passcodesEnabled && usePasscodesResetPassword) {
+				// set the encrypted state cookie to persist the email and stateHandle
+				// which we need to persist during the passcode reset flow
+				setEncryptedStateCookie(res, {
+					email,
+					stateHandle: `${crypto.randomUUID()}`, // use a random non-existent stateHandle, so encrypted value is different
```
coldlink, line 822:
```suggestion
					stateHandle: `${crypto.randomUUID()}`, // use a random non-existent stateHandle, so encrypted value is different
```

generate a random string which is the same length as a real state handle

akinsola-guardian (approved):
Thanks for taking us through this. LGTM

PR #2925, opened 2024-09-30
Okta | Fix users without `isGuardianUser` flag, but in `GuardianUser-All` group

coldlink:
## What does this change?

We found a scenario where the `isGuardianUser` flag on the user object wasn't being set when the user created an account using a social provider, but were added to the `GuardianUser-All` group.

This is similar to what we previously identified with the `emailValidated` flag, and the `GuardianUser-EmailValidated` group. See https://github.com/guardian/gateway/pull/1682 for more context on that.

This PR updates the logic that after authentication, we check for this mismatch by identifying users who don't have this flag set (`is_guardian_user`), but are in the `GuardianUser-All` group and fix them by updating the user's profile with the `isGuardianUser` flag.

## Tested

- [ ] CODE

### DIFF
--- a/src/server/routes/oauth.ts
+++ b/src/server/routes/oauth.ts
@@ -52,6 +52,7 @@ interface OAuthError {
 interface CustomClaims extends IdTokenClaims {
 	user_groups?: string[];
 	email_validated?: boolean;
+	is_guardian_user?: boolean;
 }
 
 /**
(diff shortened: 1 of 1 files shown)


pvighi (approved):
looks good to me

coldlink:
Closing as this is no longer needed.

PR #2931, opened 2024-10-08
Passwordless | Make IDX API default for sign in with password flows!

coldlink:
## What does this change?

As it says on the tin!

In https://github.com/guardian/gateway/pull/2926 we implemented using the Okta IDX API for sign in with a password behind the `useIdxSignIn` feature flag.

This PR makes the Okta IDX API the default option when a user attempts to sign in with a password.

We do this by removing the check for the `useIdxSignIn` query parameter.

We also update the code, so that if the `useOktaClassic` query parameter is provided, it uses the legacy behaviour instead (sends an email with a link).

Most of this PR is updating the tests to check and handle both behaviours.

## Tested

- [x] CODE

### DIFF
--- a/cypress/fixtures/okta-responses/error/idx-challenge-answer-password-401-response.json
+++ b/cypress/fixtures/okta-responses/error/idx-challenge-answer-password-401-response.json
@@ -1,0 +1,267 @@
+{
+	"code": 401,
+	"response": {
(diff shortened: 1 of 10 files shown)


AshCorr (approved):
🥳

PR #2943, opened 2024-10-16
Passwordless | Refactors/Fixes before sign in with passcodes

coldlink:
## What does this change?

In #2942 we are implementing Sign In with Passcodes.

In interest of making that PR smaller, we've separated out a number of refactors/fixes into this PR, so they can be safely reviewed and merged in before being used by the metioned PR.

## Changes

**Add `shortRequestId` to `PasswordEmailSent` component inside the `UnvalidatedEmailEmailSentPage`**
*fix*

The `shortRequestId` was missing from the `PasswordEmailSent` component inside the `UnvalidatedEmailEmailSentPage`, so unknown errors would have been shown errors without the request id for debugging.

This commit adds that in the `shortRequestId` prop to make sure errors get shown with the request id if required

---

**Add ability to set individual flags inside the `validateEmailAndPasswordSetSecurely` function**
*refactor*

When signing in with passcodes we need to be able to set the `emailValidated` flag independently from the `passwordSetSecurely` flag, this commit updated the function to be able to do this

---

**Create `oktaIdxShared.ts` controller file to share code**
*refactor*

For sign in with passcodes, we'll be sharing some code from the `sendChangePasswordEmail` controller, so move the shared code into it's own file to make it easier to manage.

---

**Create `signInControllers.ts` controller file to handle sign in behaviour in the idx api**
*refactor*

Moves the logic for the sign in with idx behaviour into it's own file, allowing us to separate out the logic related to the idx api when performing sign in

---

**Refactor `dangerouslySetPlaceholderPassword`**
*refactor*

Update `dangerouslySetPlaceholderPassword` to be a named export (like all the other functions in gateway).

Refactor some of the code into a `dangerouslySetPlaceholderPasswordUsingRecoveryToken` which takes in a recovery token which is used to set a placeholder password, as the token isn't always generated using `dangerouslyResetPassword`, but could also come from `activateUser`.

## Tested

- [x] CODE


src/server/lib/okta/validateEmail.ts
```diff
@@ -1,54 +1,83 @@
...
+ * @description Use this method to update the emailValidated and passwordSetSecurely flags for a user.
  *
  * @param {string} id accepts the Okta user ID, email address (login) or login shortname (as long as it is unambiguous)
  * @param {string} ip The IP address of the user
  * @param {boolean} flagStatus Default is true. If true, the flags are set to true, if false, the flags are set to false
+ * @param {boolean} updateEmail Default is true. If true, the emailValidated flag is updated to the flagStatus
+ * @param {boolean} updatePassword Default is true. If true, the passwordSetSecurely flag is updated to the flagStatus
  * @returns {Promise<UserResponse>} Promise that resolves to the user object
  */
 export const validateEmailAndPasswordSetSecurely = async ({
```
AshCorr, line 18: non blocking, does this function name still make sense? Feels like its doing a lot more than just validating a user!

src/server/controllers/oktaIdxShared.ts
```diff
@@ -0,0 +1,257 @@
...
+			},
+			req.ip,
+		);
+
+		// Unset the emailValidated and passwordSetSecurely flags
+		await validateEmailAndPasswordSetSecurely({
+			id: user.id,
+			ip: req.ip,
+			flagStatus: false,
+		});
```
AshCorr, line 249: Might be missing something, but this seems very similar to what [dangerouslySetPlaceholderPassword](https://github.com/guardian/gateway/blob/b9e2e18eb61217c41da2e7ea485173a89597a727/src/server/lib/okta/dangerouslySetPlaceholderPassword.ts#L36C16-L36C49) is doing?

AshCorr (approved):

pvighi (dismissed):

src/server/controllers/oktaIdxShared.ts
coldlink: that's a good spot! I'll refactor that.

AshCorr (approved):

PR #2948, opened 2024-10-21
Dependencies | 21st October 2024

coldlink:
## What does this change?

Updates dependencies where possible since #2935
 
Holds back the following:

- Eslint to v9 - needs a configuration changes

## Tested

- [x] CODE


cdk/package.json
```diff
@@ -12,15 +12,15 @@
     "diff": "cdk diff --path-metadata false --version-reporting false"
   },
   "devDependencies": {
-    "@guardian/cdk": "59.5.2",
+    "@guardian/cdk": "59.5.3",
```
AshCorr, line 15: This version of CDK is broken! See https://github.com/guardian/cdk/pull/2484

coldlink: I'll revert!

AshCorr (approved):

cdk/package.json
coldlink: Updated 😄

PR #2952, opened 2024-10-21
Documentation | Okta IDX API and Interaction Code flow

coldlink:
## What does this change?

Add a whole bunch of documentation about the Okta Interaction Code Flow, and the Okta IDX API.

This includes information on how it works, why we're using it, it's usecase in Gateway too. 

We also include reasoning as to why we're betting on passwordless (currently one-time passcodes).

It also documents all of the individual Okta IDX endpoints we're currently using within Gateway, as well as describing the flow on how the IDX API is set up for sign in (with passwords and passcodes), create account, and reset password flows.

This PR doesn't update the existing (some of it outdated) documentation, which will be done in a follow up PR.

The best way to review this PR would be to click "View File" when viewing the diff, in order to get a formatted view of the changes.

### DIFF
--- a/docs/okta/idx/README.md
+++ b/docs/okta/idx/README.md
@@ -1,57 +1,94 @@
-# Okta IDX API
+# Okta IDX API and Identity Engine
 
-Status: WIP
+## Identity Engine
 
-_More information about the Okta IDX API will be added soon_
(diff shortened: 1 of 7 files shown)


AshCorr (approved):
LGTM! Thanks for documenting all of this!

AshCorr (approved):

PR #2957, opened 2024-10-30
Documentation | Bring existing documentation up to date.

coldlink:
## What does this change?

In https://github.com/guardian/gateway/pull/2952 we updated documentation to add information about the Okta Interaction Code Flow, and the Okta IDX API.

However the existing documentation was out of date or referred to the legacy API/flows.

This PR goes through all the existing documentation and updates and simplify where appropriate, with changes not just related to the IDX API, but other changes/missing information too.


docs/gateway-flows.md
```diff
@@ -1,16 +1,20 @@
 # Flow diagrams for common Gateway routes
 
-These flow diagrams are a WIP, mostly used to help visualise some complex functions in Gateway for the purposes of mocking Okta flows in testing.
+**_Note: These diagrams are outdated and describe behaviour for the older Okta Classic API. For more up to date flows using the newer Okta IDX API, see the [IDX Documentation](./okta/idx/README.md)_**
```
pvighi, line 3: if they are outdated why do we keep them ?

AshCorr (approved):

docs/gateway-flows.md
coldlink: Because the code hasn't been removed, and can still be accessed if using `useOktaClassic=true` or if using tokens from the Classic Authentication API which is still used in some cases, i.e activation emails from the guest user endpoint.

pvighi (approved):
looks great. 
Some of the general Okta information doesn't seem too Gateway-specific to me so maybe it might make more sense to have it somewhere else and only link it from here, but I'm against leaving it here if you prefer that

PR #2990, opened 2024-12-02
Dependencies | 2nd December 2024

coldlink:
## What does this change?

Updates dependencies where possible since #2985
 
Holds back the following:

## Tested

- [x] CODE

## Notes

`src/client/components/ToggleSwitchInput.tsx` was changed due to prettier update


akinsola-guardian (approved):
LGTM

PR #3000, opened 2024-12-09
Pressreader Editions | Native App / OS Detection

coldlink:
## What does this change?

Unlike other apps, within Okta, the Pressreader Editions app for both Android and iOS is set up as a single application, this was done to mirror the already existing Editions app.

However we need to know whether the user is using the iOS or Android version of the app (`isNativeApp` variable) to show app specific design variants and hide the CMP banner, as well as displaying the name of the app if the user is signed in already (`appName` variable).

Thankfully we already use [`Bowser`](https://github.com/bowser-js/bowser) to parse the `user-agent` header we get from the client.

So we can use a combination of the `appLabel` (`editions_pressreader`) and the user agent to determine whether we're on iOS or Android, and derive the correct app name!

## Tested

- [x] CODE

### DIFF
--- a/src/shared/lib/appNameUtils.ts
+++ b/src/shared/lib/appNameUtils.ts
@@ -1,4 +1,6 @@
 import { isOneOf } from '@guardian/libs';
+import Bowser from 'bowser';
+import { IsNativeApp } from '@/shared/model/ClientState';
 
 /**
  * list of app prefixes and labels that is used by native apps to determine
@@ -22,6 +24,7 @@ export const apps = [
 	['ios_live_app', 'il_'],
 	['android_feast_app', 'af_'],
 	['ios_feast_app', 'if_'],
+	['editions_pressreader', 'ed_'],
 ] as const;
 
 type AppLabel = (typeof apps)[number][0];
@@ -34,7 +37,7 @@ export const isAppPrefix = isOneOf(appPrefixes);
 const appLabels = apps.map(([label]) => label);
 export const isAppLabel = isOneOf(appLabels);
 
-export type AppName = 'Guardian' | 'Feast';
+export type AppName = 'Guardian' | 'Feast' | 'Editions';
 
 /**
  * @name getAppPrefix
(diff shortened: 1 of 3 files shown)


akinsola-guardian (dismissed):
LGTM

src/shared/lib/appNameUtils.ts
```diff
@@ -60,5 +69,28 @@ export const getAppName = (
...
+		case 'editions_pressreader':
+			return 'Editions';
+	}
+};
+
+/**
+ * @name getNativeAppFromBowser
+ * @description Use a Bowser instance to determine if the browser is a native os.
+ * @param browser - Bowser instance
+ * @returns {IsNativeApp} - 'android' or 'ios' if the browser is a native app, otherwise undefined
```
pvighi, line 82: is this strictly true ? wouldn't this just return whether you are using a mobile device? even if you don't have the app installed this would return true if you use your phone browser, right ? Do we not care about this distinction ?

coldlink: Have updated both the name and comment to make it clearer.

src/shared/lib/appNameUtils.ts
```diff
@@ -60,5 +69,28 @@ export const getAppName = (
...
+
+/**
+ * @name getIsNativeAppFromBowser
+ * @description Use a Bowser instance to determine if the browser is a native os.
+ * @param browser - Bowser instance
+ * @returns {IsNativeApp} - 'android' or 'ios' if the browser matches Android or iOS, otherwise undefined
+ */
+export const getIsNativeAppFromBowser = (
+	browser: Bowser.Parser.Parser,
+): IsNativeApp | undefined => {
```
pvighi, line 86: I still think this is slightly confusing because it mentions Native when the code doesn't have anything to do with native app. Is Native os a term people use ? because the way I understand it it's  just the client apps that can be native or running on a browser

But I guess the return type is `IsNativeApp`.  Also it's pretty strange to me that a type with that name is not a boolean but that is not introduced in this PR.

In any case this is all about naming so I'll approve and leave it up to you if you are happy to merge it like this

pvighi (approved):

PR #3023, opened 2025-01-06
Dependencies | 6th January 2025

coldlink:
## What does this change?

Updates dependencies where possible since #3008
 
Holds back the following:

- `@types/react` - Stick to v18 types, until we're ready to upgrade preact

## Tested

- [x] CODE


akinsola-guardian (approved):
LGTM

PR #3025, opened 2025-01-08
Passcodes | Use new sign in ux for passcodes behind flag

coldlink:
## What does this change?

In #2942 we implemented sign in using a one time passcode behind a feature flag (query parameter) `?usePasscodeSignIn=true`.

The UX in that example wasn't particularly ideal, where we used a toggle to allow the reader to switch between sign in via passcodes or sign in via passwords.

In this PR we change the UX to remove the toggle, and instead when using the feature flag, we make passcodes the default option.

In this scenario, the reader has to just enter their email address and click "Continue with email". We then send the user a passcode, and show the passcode input page as before. However we've added an additional button here which allows the reader to "Sign in with password instead".

This button will take them to a sign in page with the password input, where they can use a password to sign in.

<table>
<tr>
<th>Passcode</th>
<th>Password</th>
</tr>
<tr>
<td>

https://github.com/user-attachments/assets/b342cd0b-0d54-4976-9de5-56dfee08ff6c

</td>
<td>

https://github.com/user-attachments/assets/3800d929-3067-4ce1-805f-000ee2e65370

</td>
</tr>
</table>

This PR also sets up an AB test for passcodes, which rather than used for AB testing, we use it to release passcodes to 10% of the audience so we can test out the flow and look for any issues that may occur.

## Tested

- [x] CODE

### DIFF
--- a/src/client/pages/PasscodeEmailSent.tsx
+++ b/src/client/pages/PasscodeEmailSent.tsx
@@ -7,8 +7,10 @@ import { MinimalLayout } from '@/client/layouts/MinimalLayout';
 import { PasscodeInput } from '@/client/components/PasscodeInput';
 import { EmailSentInformationBox } from '@/client/components/EmailSentInformationBox';
 import { EmailSentProps } from '@/client/pages/EmailSent';
+import { buildUrl } from '@/shared/lib/routeUtils';
+import ThemedLink from '@/client/components/ThemedLink';
 
-type TextType = 'verification' | 'security' | 'generic';
+type TextType = 'verification' | 'security' | 'generic' | 'signin';
 
 type Props = {
 	passcodeAction: string;
@@ -19,6 +21,7 @@ type Props = {
 	timeUntilTokenExpiry?: number;
 	noAccountInfo?: boolean;
 	textType?: TextType;
+	showSignInWithPasswordOption?: boolean;
 };
 
 type PasscodeEmailSentProps = EmailSentProps & Props;
@@ -29,6 +32,7 @@ type Text = {
 	sentTextWithEmail: string;
 	sentTextWithoutEmail: string;
 	securityText: string;
+	passcodeInputLabel: string;
 	submitButtonText: string;
 };
 
@@ -43,6 +47,7 @@ const getText = (textType: TextType): Text => {
 					'We’ve sent you a temporary verification code. Please check your inbox.',
 				securityText:
 					'For your security, the verification code will expire in 30 minutes.',
+				passcodeInputLabel: 'Verification code',
 				submitButtonText: 'Submit verification code',
 			};
 		case 'security':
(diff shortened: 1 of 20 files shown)


src/client/pages/PasscodeEmailSent.tsx
```diff
@@ -153,8 +172,28 @@ export const PasscodeEmailSent = ({
...
 			</MainForm>
+			{showSignInWithPasswordOption && (
+				// <LinkButton
+				// 	cssOverrides={secondaryButtonStyles()}
+				// 	priority="tertiary"
+				// 	href={`${buildUrl('/signin/password')}${queryString}`}
+				// >
+				// 	Sign in with password instead
+				// </LinkButton>
+				<>
```
AshCorr, line 189: Still need this fragment?

src/client/pages/SignIn.tsx
```diff
@@ -175,14 +142,13 @@ export const SignIn = ({
...
 	isReauthenticate = false,
 	shortRequestId,
-	defaultView = 'passcode',
-	currentView = defaultView,
 	usePasscodeSignIn = false,
+	hideSocialButtons = false,
 }: SignInProps) => {
 	// status of the OTP checkbox
-	const [selectedView, setSelectedView] = React.useState(
+	const [selectedView] = React.useState(
```
AshCorr, line 149: Does this still need to be a React state if its readonly?

src/shared/model/experiments/tests/passcode-signin-test.ts
```diff
@@ -0,0 +1,25 @@
+import { ABTest } from '@guardian/ab-core';
+
+export const passcodeSignInTest: ABTest = {
+	id: 'PasscodeSignInTest', // This ID must match the Server Side AB Test
+	start: '2025-01-13',
+	expiry: '2025-01-31', // Remember that the server side test expiry can be different
+	author: 'mahesh.makani@theguardian.com',
+	description:
+		'Testing the release of one time passcodes as the default sign in option',
+	audience: 0.1, // 10% (1 is 100%)
```
AshCorr, line 10: So does this mean that once this PR is merged 10% of users will start seeing the new UX or is there something else that has to be toggled first?

AshCorr (dismissed):

src/client/pages/SignIn.tsx
coldlink: Ah yes, good point! I'll make it static.

src/shared/model/experiments/tests/passcode-signin-test.ts
coldlink: Yeah, it will go out to 10% of users when the PR is merged in! The switch is already set to on.

src/client/pages/PasscodeEmailSent.tsx
coldlink: Nope! And I don't need the comment above anymore either.

AshCorr (approved):

PR #3033, opened 2025-01-16
Dependencies | 20th January 2025

coldlink:
## What does this change?

Updates dependencies where possible since #3029
 
Holds back the following:

## Tested

- [ ] CODE


akinsola-guardian (approved):
LGTM

PR #3035, opened 2025-01-16
Passcodes | Release sign in with passcodes to all users

coldlink:
## What does this change?

Releases passcodes for sign in to all of the audience following on from https://github.com/guardian/gateway/pull/3034

We also remove the AB test and the `usePasscodeSignIn` query parameter which would've been needed in order to view the sign in with passcode page.

We add a `usePasswordSignIn` query parameter flag, which preserves the previous behaviour of sign in with a password directly on the initial sign in page. This is useful as it means tests can still continue to run without having to be integrated with passcodes for sign in.

## Tested

- [x] DEV
- [x] CODE


src/server/controllers/signInControllers.ts
```diff
@@ -400,13 +400,11 @@ export const oktaIdxApiSignInController = async ({
...
-	const usePasscodeSignInFlag =
-		res.locals.queryParams.usePasscodeSignIn ||
-		res.locals.abTestAPI.isUserInVariant('PasscodeSignInTest', 'variant');
+	// we're using passwords if the `usePasswordSignIn` query parameter is set
+	const usePasscodeSignInFlag = !res.locals.queryParams.usePasswordSignIn;
 
 	try {
-		// only attempt to sign in with a passcode if the user currently has the query parameter set
-		// this should be removed when we're ready to enable this for all users
+		// only attempt to sign in with a passcode if we've enabled passcodes for all users
```
pvighi, line 407: is this comment correct? it seems like this variable means that the user has not forced password sign in with the parameter, right ?

src/server/controllers/signInControllers.ts
```diff
@@ -417,6 +415,8 @@ export const oktaIdxApiSignInController = async ({
 			}
 		}
 
+		// Otherwise use password sign in
```
pvighi, line 418: if the usePasswordSignIn flag is not set but there's no passcode in the request body then it would attempt password sign in ? is this the desired behaviour ?

src/server/controllers/signInControllers.ts
```diff
@@ -400,13 +400,11 @@ export const oktaIdxApiSignInController = async ({
 	// get the email and password from the request body
 	const { email = '', password = '', passcode } = req.body;
 
-	const usePasscodeSignInFlag =
-		res.locals.queryParams.usePasscodeSignIn ||
-		res.locals.abTestAPI.isUserInVariant('PasscodeSignInTest', 'variant');
+	// we're using passwords if the `usePasswordSignIn` query parameter is set
+	const usePasscodeSignInFlag = !res.locals.queryParams.usePasswordSignIn;
```
pvighi, line 404: keeping the `usePasscodeSignInFlag` might be confusing now that it means that the `usePasswordSignIn`  was NOT set

src/server/controllers/signInControllers.ts
coldlink: I've updated the comments on this page to make it clearer about whats going on!

src/server/controllers/signInControllers.ts
coldlink: That's right. It only attempts to use passcodes if it sees the `passcode` in the request params, which is just a hidden flag saying to use the passcode flow rather than the password flow

src/server/controllers/signInControllers.ts
coldlink: I've added comments and updated the name to make it clearer about what's going on!

src/server/controllers/signInControllers.ts
```diff
@@ -397,17 +397,22 @@ export const oktaIdxApiSignInController = async ({
 	req: Request;
 	res: ResponseWithRequestState;
 }) => {
-	// get the email and password from the request body
+	// get the email and password from the request body if using passwords
+	// or the "passcode" parameter is a hidden input, which is to determine if the
+	// user is signing in with a passcode and not an actual passcode value
```
pvighi, line 402: > // user is signing in with a passcode and not an actual passcode value

What is the difference between 'passcode` and `passcode value` ?

cypress/integration/ete/delete.4.cy.ts
```diff
@@ -4,7 +4,7 @@ describe('Delete my account flow in Okta', () => {
 		cy.visit(
 			`/signin?returnUrl=${encodeURIComponent(
 				`https://${Cypress.env('BASE_URI')}/welcome/review`,
-			)}`,
+			)}&usePasswordSignIn=true`,
```
pvighi, line 7: does this mean we are not testing the default sign in method now ?

pvighi (approved):
added some comments but in general it looks good to me

src/server/controllers/signInControllers.ts
coldlink: It's not a "code" like "123456", but a key/value set to passcode coming from: https://github.com/guardian/gateway/blob/f28db8f39c863b4831140bff873f7d150eed3fdf/src/client/pages/SignIn.tsx#L210

cypress/integration/ete/delete.4.cy.ts
coldlink: We are! It's just in other test files that have already been added. Adding `usePasswordSignIn` to tests just means we don't have to update the authentication specific tests, e.g. for delete my account.

PR #3045, opened 2025-01-27
Passwordless | Add countdown timer to "send again" functionality

coldlink:
## What does this change?

A number of users are complaining that they're getting multiple passcodes to their email, each time with a different code, and the original code no longer valid.

We think this is due to delays in the user receiving the email

So rather than allow the user to resend the passcode straight away, we added a countdown timer so that they can only do it after a specificed amount of time, currently set to 30 seconds.

We've only enabled this for passcode users, as the same problem doesn't occur with links.

We also update the `MainForm` component to take a `disabled` value.

A new hook provides the countdown functionality on the client side: `useCountdownTimer`
This hook takes a `durationInSeconds` parameter to start the countdown, and returns the `timeRemaining` in seconds, and an `isComplete` boolean.

This is then integrated withing the `EmailSentInformationBox` component, which houses the "send again" functionality. Where it takes a new `sendAgainTimerInSeconds` prop in order to control it.

In turn the same thing was applied to `PasscodeEmailSent` component which uses the `EmailSentInformationBox`, followed by the pages that use the `PasscodeEmailSent` component.

Finally in order to set the timer, rather than harcode the value everywhere, we add it to the `ClientState` and `PageData` objects. This allows us to only set it in one place (through `getRequestState`), but also lets us modify the value specifically for tests, i.e Cypress. A `getPasscodeSendAgainTimer` helper function helps with this.

<table>
<tr>
<th>Video</th>
</tr>
<tr>
<td>

https://github.com/user-attachments/assets/33077a80-269d-4974-b250-efad3e569281

</td>
</tr>
</table>

## Tested

- [x] DEV
- [x] CODE

### DIFF
--- a/src/server/lib/passcodeSendAgainTimer.ts
+++ b/src/server/lib/passcodeSendAgainTimer.ts
@@ -1,0 +1,15 @@
+import { Request } from 'express';
+
+export const getPasscodeSendAgainTimer = (req: Request): number => {
+	// if we're running in cypress, we want to use the cypress mock state cookie to set the timer for testing
+	const runningInCypress = process.env.RUNNING_IN_CYPRESS === 'true';
+	if (runningInCypress) {
+		const cypressMockStateCookie = req.cookies['cypress-mock-state'];
+
+		if (cypressMockStateCookie && !isNaN(+cypressMockStateCookie)) {
+			return +cypressMockStateCookie;
+		}
+	}
+
+	return 30; // 30 seconds
+};
--- a/cypress/integration/ete/registration_1.2.cy.ts
+++ b/cypress/integration/ete/registration_1.2.cy.ts
@@ -24,6 +24,8 @@ const existingUserSendEmailAndValidatePasscode = ({
 	expectedEmailBody?: 'Your one-time passcode' | 'Your verification code';
 	additionalTests?: 'passcode-incorrect' | 'resend-email' | 'change-email';
 }) => {
+	cy.setCookie('cypress-mock-state', '1'); // passcode send again timer
+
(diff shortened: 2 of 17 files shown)


src/server/lib/passcodeSendAgainTimer.ts
```diff
@@ -0,0 +1,15 @@
+import { Request } from 'express';
+
+export const getPasscodeSendAgainTimer = (req: Request): number => {
+	// if we're running in cypress, we want to use the cypress mock state cookie to set the timer for testing
+	const runningInCypress = process.env.RUNNING_IN_CYPRESS === 'true';
+	if (runningInCypress) {
+		const cypressMockStateCookie = req.cookies['cypress-mock-state'];
```
pvighi, line 7: Some questions to understand the cypress tests better:

- does `cypress-mock-state` hold the value for the counter specifically  or it just a value used for anything ?
- Where is it set ?

pvighi, line 7: maybe a silly question because I don't know cypress that well:

is there any reason for this to be called `cypress-mock-state` and not something that refers to the timer ?

pvighi (approved):

src/server/lib/passcodeSendAgainTimer.ts
coldlink: It's because it's a pattern we use already in Gateway to customise behaviour on the sever based on Cypress tests.

So we set a `cypress-mock-state` cookie in other tests that also modifies behaviour is other ways only when it exists. Which is useful for testing edge case scenarios.

PR #3046, opened 2025-01-30
Fix - Metrics that were missing or werent firing

coldlink:
## What does this change?

After launching passcodes and looing at the metrics dashboard, I noticed a couple that were missing or not firing.

This PR fixes that.

Firstly `OktaPasswordSignInFlow` metric no longer needs a `usePasscode` check, as this is always the case now, and hadn't been firing since https://github.com/guardian/gateway/pull/3035

Next `OAuthAuthorization::Success` had been missing since May 2024 when it had been accidentally removed in https://github.com/guardian/gateway/pull/2722


akinsola-guardian (approved):
LGTM

PR #3052, opened 2025-02-04
Passcodes | Auto submit form on paste or change when valid

coldlink:
## What does this change?

Currently the reader has to manually submit the code after typing it in or pasting it in.

To make it faster for the reader, this PR updates the functionality to automatically submit the code when it reaches 6 digits long or pasting it in.

We added checks to make sure that the auto-submission is disabled for specifc scenarios
- If there is an error on the field
- If the reader hasn't previously interacted with the field
- If the value isn't a 6 digit number

It is also disabled in scenarios where `formRef` isn't passed in.

Speaking of which, a minor refactor had to be performed on the `MainForm` to make `formRef` a parameter instead. This allows us to pass in a `ref` for the `MainForm`, so it can be used in other components, specifically in this scenario the `PasscodeInput` component.

Finally `PasscodeEmailSent` page was updated to create the new `formRef` and pass it to both `MainForm` and `PasscodeInput`

<table>
<tr>
<th>Previous</th>
<th>Auto submit - onChange</th>
<th>Auto submit - onPaste</th>
<th>Auto submit - disabled after error</th>
</tr>
<tr>
<td>

https://github.com/user-attachments/assets/a33c9752-95c3-44c0-8573-e7c5268358e5

</td>
<td>

https://github.com/user-attachments/assets/217a3543-8f86-4cd4-a203-372a4134cd93

</td>
<td>

https://github.com/user-attachments/assets/361b648d-c492-41f7-a3af-ab687f66dc40

</td>
<td>

https://github.com/user-attachments/assets/7cdd726b-bf84-48fa-b0ed-28a6e9cca59f

</td>
</tr>
</table>

## Tested

- [x] DEV
- [x] CODE
  - [x] Chrome Web
  - [x] Chrome Mobile
  - [x] Firefox Web
  - [x] Firefox Mobile 
  - [x] Safari Web
  - [x] Safari Mobile

### DIFF
--- a/src/client/components/PasscodeInput.tsx
+++ b/src/client/components/PasscodeInput.tsx
@@ -1,4 +1,4 @@
-import React, { useState } from 'react';
+import React, { useEffect, useState } from 'react';
 import { FieldError } from '@/shared/model/ClientState';
 import { TextInputProps } from '@guardian/source/react-components';
 import { css } from '@emotion/react';
@@ -8,6 +8,7 @@ import { remSpace } from '@guardian/source/foundations';
 interface PasscodeInputProps extends TextInputProps {
 	passcode?: string;
 	fieldErrors?: FieldError[];
+	formRef: React.RefObject<HTMLFormElement | null>;
 }
 
 const passcodeInputStyles = css`
@@ -19,13 +20,13 @@ export const PasscodeInput = ({
 	label,
 	passcode = '',
 	fieldErrors,
+	formRef,
 }: PasscodeInputProps) => {
 	/**
 	 * In gateway we normally avoid using client side javascript, but in this case
 	 * we allow it as it an enhancement to the user experience and not required for
 	 * the form to work.
 	 */
-
 	// This function removes all non-numeric characters from the input
 	const sanitiseCode = (code: string) => {
 		return code.replace(/\D*/g, '');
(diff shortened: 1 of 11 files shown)


src/client/components/PasscodeInput.tsx
```diff
@@ -8,6 +8,7 @@ import { remSpace } from '@guardian/source/foundations';
 interface PasscodeInputProps extends TextInputProps {
 	passcode?: string;
 	fieldErrors?: FieldError[];
+	formRef?: React.RefObject<HTMLFormElement | null>;
```
AshCorr, line 11: Would this ever be undefined?

src/client/pages/PasscodeEmailSent.tsx
```diff
@@ -112,6 +112,7 @@ export const PasscodeEmailSent = ({
 	const [recaptchaErrorMessage, setRecaptchaErrorMessage] = useState('');
 	const [recaptchaErrorContext, setRecaptchaErrorContext] =
 		useState<ReactNode>(null);
+	const formRef = createRef<HTMLFormElement>();
```
AshCorr, line 115: Should we use `useRef` instead as we're using function components 
 as per 
https://react.dev/reference/react/createRef

image

AshCorr (approved):

src/client/pages/PasscodeEmailSent.tsx
coldlink: Ah good spot! I'll try and refactor it!

src/client/components/PasscodeInput.tsx
coldlink: The way it's currently being used, it's always defined, so will refactor that.

AshCorr (approved):

PR #3054, opened 2025-02-05
fix(passcodes): user enumeration on invalid passcodes

coldlink:
Okta will only allow a maximum of 5 attempts for passcodes before giving an errorm where the user will be redirected to the sign in page/

However if a user doesn’t exist in Okta, we fake this functionality to always show an passcode error, however we don’t keep track of the number of attempts.

This means that you can work out if an account exists or not, by attempting to submit a passcode more that 5 times on the sign in/password reset flow. If you’re able to submit the code more than 5 times, then the user doesn’t exist in Okta, if you’re not able to submit more than 5 times then the user exists in Okta.

Essentially we need to replicate this behaviour for user’s who don’t exist in Okta, by keeping track of how many times the code was submitted.

This commit does this by adding and incrementing a `passcodeFailedCount` on the `EncryptedState` cookie. Then when we reach 5 attempts (or more), we redirect to the expired page, in order to replicate this behaviour for all scenarios.

## Accessibility


coldlink:
Replaced by https://github.com/guardian/gateway/pull/3056 and https://github.com/guardian/gateway/pull/3057

PR #3055, opened 2025-02-05
Fix | Show correct error page on password incorrect

coldlink:
## What does this change?

When using the password option instead of passcodes, this occurs on the `/signin/password` page now.

However we made a mistake in the error handling ternary operator to determine the error page.

We've fixed it so that `usePasscode` is falsy, then go to `/signin/password` as the error page, as the user is using password to sign in.

## Tested

- [x] DEV
- [x] CODE


pvighi (approved):
good job adding a test case for this as well

PR #3057, opened 2025-02-06
Passcodes | Update expired functionality for create account/reset password

coldlink:
## What does this change?

Previously when the passcode had expired, or had been incorrect multiple times, we'd redirect users to the older `/welcome/expired` page for create account (register), and `/reset-password/expired` for reset password.

The problem with these pages is that the copy and UX was more suited to when a link rather than code had expired/invalidated.

Instead we now take the same approach as taken on the Sign In page, where we redirect back to the initial page the user took the action on, but show an error with context saying their code had expired, and to receive a new one instead.

We also add a new errors grouping, specifically for passcodes, called `PasscodeErrors`.

<table>
<tr>
<th>Previous behaviour - Reset password</th>
<th>Previous behaviour - Create account</th>
<th>New behaviour - Reset password</th>
<th>New behaviour - Create account</th>
</tr>
<tr>
<td>

https://github.com/user-attachments/assets/d9b98661-0d27-4d1b-adb9-cbf523290685

</td>
<td>

https://github.com/user-attachments/assets/da710802-6f68-432c-a94f-7b8c29bba26c

</td>
<td>

https://github.com/user-attachments/assets/9eaf00ab-b4fd-4f5d-87d0-ff8e7ca967c4

</td>
<td>

https://github.com/user-attachments/assets/5592ceb1-879b-4312-9bfe-dbda7eef4658

</td>
</tr>
</table>

## Tested

- [x] DEV
- [x] CODE


src/shared/model/Errors.ts
```diff
@@ -15,6 +15,11 @@ export enum GenericErrors {
 	DEFAULT = 'Sorry, something went wrong. Please try again.',
 }
 
+export const PasscodeErrors = {
+	PASSCODE_INVALID: 'Incorrect code',
+	PASSCODE_EXPIRED: 'Your code has expired',
+} as const;
+
```
coldlink, line 22: TODO: refactor the rest of the errors to use `as const` instead of enums

PR #3058, opened 2025-02-07
Passwordless | Move sign in with password instead option

coldlink:
## What does this change?

Recently we deployed sign in with passcodes to all users.

While we noticed that overall sign in numbers remained consistent, we did get a few user complaints saying that they did not like how the "sign in with password" option was only available after getting the one time code sent to their email. Other users complained that they thought they could input a password into the passcode field box and sign in with a password directly.

To improve the UX, we've made some minor adjustments to the sign in flow.

Firstly we added a "sign in with a password instead" link to the sign in landing page, which takes the email from the email input, and redirects the user to `/signin/password` with the email option prefilled, or directly to the page if there is no email. We also moved the "sign in with a password instead" option on the passcode input page to the information box, as an additional option if a user is having trouble with passcodes.

We belive that this should alleviate some of the issues users are having, even as we're prioritising sign in with passcodes.

We also rename the passcode CTA from "Continue with email" to Sign in with email" instead.

<table>
<tr>
<th>Sign In page</th>
<th>Passcode page</th> 
</tr>
<tr>
<td>

https://github.com/user-attachments/assets/64e343b1-adf7-46c2-bdd4-571ddc0497ee

</td>
<td>

https://github.com/user-attachments/assets/88e49500-451c-4bcd-ba11-d59f5eeb3307

</td>
</table>

## Tested

- [x] DEV
- [x] CODE

### DIFF
--- a/src/client/components/EmailSentInformationBox.tsx
+++ b/src/client/components/EmailSentInformationBox.tsx
@@ -31,6 +31,7 @@ type EmailSentInformationBoxProps = Pick<
 	>;
 	setRecaptchaErrorMessage: React.Dispatch<React.SetStateAction<string>>;
 	sendAgainTimerInSeconds?: number;
+	showSignInWithPasswordOption?: boolean;
 };
 
 const sendAgainFormWrapperStyles = css`
@@ -50,6 +51,7 @@ export const EmailSentInformationBox = ({
 	queryString,
 	shortRequestId,
 	sendAgainTimerInSeconds,
+	showSignInWithPasswordOption,
 }: EmailSentInformationBoxProps) => {
 	const timer = useCountdownTimer(sendAgainTimerInSeconds || 0);
 
@@ -89,9 +91,17 @@ export const EmailSentInformationBox = ({
 				)}
 				{changeEmailPage && (
 					<>
-						, or{' '}
+						,{!showSignInWithPasswordOption ? <> or </> : <> </>}
 						<ThemedLink href={`${changeEmailPage}${queryString}`}>
 							try another address
+						</ThemedLink>
+					</>
+				)}
+				{showSignInWithPasswordOption && (
+					<>
+						, or{' '}
+						<ThemedLink href={`${buildUrl('/signin/password')}${queryString}`}>
+							sign in with a password instead
 						</ThemedLink>
 					</>
 				)}
--- a/src/client/pages/SignIn.tsx
+++ b/src/client/pages/SignIn.tsx
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, { useEffect } from 'react';
 import {
 	extractMessage,
 	GatewayError,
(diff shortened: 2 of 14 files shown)


src/client/components/EmailSentInformationBox.tsx
```diff
@@ -89,12 +91,20 @@ export const EmailSentInformationBox = ({
 				)}
 				{changeEmailPage && (
 					<>
-						, or{' '}
+						,{!showSignInWithPasswordOption ? <> or </> : <> </>}
```
pvighi, line 94: sorry I'm having a bit of trouble understanding the intent (or maybe the syntax) here:

is the intention that if it has to show both links it would say 
`try another address or sign in with a password instead`
but if it only has to one it would just say `or try another address / or sign in with a password instead` ?

coldlink: the latter option!

coldlink: you can see in storybook on how it behaves: https://60929d168594f80039336501-dlorewwtdy.chromatic.com/?path=/story/components-emailsentinformationbox--with-show-sign-in-with-password-option&globals=viewport:mobile

src/client/pages/SignIn.tsx
```diff
@@ -145,7 +146,23 @@ export const SignIn = ({
...
+	useEffect(() => {
+		if (typeof window !== 'undefined' && focusPasswordField && email) {
+			const passwordInput: HTMLInputElement | null =
+				window.document.querySelector('input[name="password"]');
+
+			if (passwordInput) {
+				passwordInput.focus();
+			}
+		}
+	}, [focusPasswordField, email]);
```
AshCorr, line 164: Would [`autofocus`](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/autofocus) work here?

For some reason my screen reader isn't announcing the "password" field when its getting `.focus()`ed, not sure if thats a side effect from it being called by `useEffect` or if im just using the screen reader wrong, but I wonder if using the browsers built in autofocus it might work better.

coldlink: I'll refactor to use `autofocus` anyway!

As it avoids a `useEffect` and any client side code, as the `autofocus` attribute can be applied on the server side render of the page!

AshCorr (approved):

src/client/components/EmailSentInformationBox.tsx
pvighi: is a comma before 'or' correct ?
It looks weird to me but sadly I don't have an english degree to know this kind of things for sure

coldlink: In generally I've always used a comma before the or in the last item in a list, so it separated from the other items in the same way, like an array.

See `Oxford comma` in the Guardian style guide: https://www.theguardian.com/guardian-observer-style-guide-o#:~:text=what%20we%20use-,Oxford%20comma,-a%20comma%20before

PR #3071, opened 2025-02-18
Remove Sentry

coldlink:
## What does this change?

Sentry for [gateway](https://the-guardian.sentry.io/projects/gateway/?project=6138895) hasn't been working for a long time.

Rather than fixing this, we determined with the team that we weren't getting any useful help from using sentry to debug client side errors even when it was working. It also takes up a large proportion of the `vendors` bundle in a production build, around 100kb (30kb gzipped), or roughly a 1/3 of total size of the bundle.

Taking both of these into account we decided to remove sentry from Gateway.

I've replaced it with the [`log`](https://github.com/guardian/csnx/tree/main/libs/%40guardian/libs/src/logger) method from `@guardian/libs`, so we can at least get debug/client side logging this way.

## Tested

- [x] CODE


package.json
```diff
@@ -63,7 +63,6 @@
 		"@testing-library/react-hooks": "npm:@testing-library/preact-hooks",
 		"@types/compression": "^1.7.5",
 		"@types/cookie-parser": "^1.4.8",
-		"@types/eslint__js": "^8.42.3",
```
coldlink, line 66: Also removed this as `knip` (`pnpm run knip`) suggested that it's not used anywhere

PR #3081, opened 2025-02-24
Dependencies | 24th February 2025

coldlink:
## What does this change?

Updates dependencies where possible since #3068 and #3071
 
Holds back the following:

## Tested

- [x] CODE


akinsola-guardian (approved):

src/server/lib/okta/api/apps.ts
```diff
@@ -35,6 +35,7 @@ export const getApp = async (id: string): Promise<AppResponse> => {
 		headers: { ...defaultHeaders(), ...authorizationHeader() },
 	}).then(handleAppResponse);
 
+	// eslint-disable-next-line functional/immutable-data -- we need to cache the app, so  we need to mutate it here
```
coldlink, line 38: caused by the `eslint-plugin-functional` major update

PR #3084, opened 2025-02-28
Fix | NoScript | Clashing styles with other components

coldlink:
## What does this change?

The styling on the error messages on some forms was broken, e.g. for incorrect email/password.

This was due to the `<noscript>` tag styles sharing styles with the error message component on the page.

Our assumption is that `<noscript>` has some weird behaviour with `emotion` (css-in-js), as it's trying to execute js inside of the `<noscript>` tag, which wouldn't work, so no styles are generated.

This means that anything else sharing styles with the noscript tag, would also fail, as emotion would think the shared styles have already been generated, e.g. the error component.

To fix this we have to separate out the styles used by the noscript tag so that all the styles are only used by that, and not share any styles with other components. Thanks to SSR, if the user has javascript disabled, all the styles will still get rendered, and the component looks okay.

As it's now no longer sharing any styles with other components, the other components also look okay too, which we're assuming is that it's not sharing styles with the noscript tag!

<table>
<tr>
<th> Before - Error
<th> After - Error
<th> Before - NoScript
<th> After - NoScript
<tr>
<td>

Screen Shot 2025-02-28 at 11 11 48

<td>

Screen Shot 2025-02-28 at 11 11 51

<td>

Screen Shot 2025-02-28 at 11 22 42

<td>

Screen Shot 2025-02-28 at 11 22 52

</table>

## Tested

- [x] CODE

### DIFF
--- a/src/client/components/NoScript.tsx
+++ b/src/client/components/NoScript.tsx
@@ -1,44 +1,116 @@
 import React from 'react';
-import { ErrorSummary } from '@guardian/source-development-kitchen/react-components';
 import locations from '@/shared/lib/locations';
 import { SUPPORT_EMAIL } from '@/shared/model/Configuration';
+import { css } from '@emotion/react';
 import {
-	errorContextSpacing,
-	errorContextLastTypeSpacing,
-	errorMessageStyles,
-} from '@/client/styles/Shared';
+	remSpace,
+	size,
+	space,
+	textSans17,
+	textSansBold17,
+} from '@guardian/source/foundations';
+import { error as errorColors } from '@guardian/source/foundations';
+import { SvgAlertRound } from '@guardian/source/react-components';
 
-// don't use any non-default html tags, or react components,
-// inside the context within the noscript tag, as this
-// has the possibility to break styling outside of the noscript tag
-export const NoScriptContext = () => (
-	<>
-		<p css={errorContextSpacing}>
-			We use JavaScript to provide a seamless and secure authentication
-			experience. Please{' '}
-			<a
-				href="https://www.whatismybrowser.com/guides/how-to-enable-javascript/"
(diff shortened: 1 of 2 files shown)


src/client/components/NoScript.tsx
```diff
@@ -1,44 +1,116 @@
...
+ *
+ * Note: This component shouldn't used any shared styles or components as this causes issues with with any other components that are rendered on the page.
+ * Which we're guessing is related to css-in-js and noscript interactions.
+ * So all styles must be defined in this file, so that they are SSR rendered and don't clash with other components.
+ *
+ * The styles are based on the ErrorSummary component from the Source design system: https://github.com/guardian/csnx/blob/main/libs/%40guardian/source-development-kitchen/src/react-components/summary/ErrorSummary.tsx
+ */
 export const NoScript = () => {
 	return (
 		<noscript>
```
AshCorr, line 86: Would wrapping <noscript> in its own [CacheProvider](https://emotion.sh/docs/cache-provider) seperate emotions rendering of the different styles for you instead of having to duplicate them?

coldlink: Ooh I'll give that a go!

coldlink: So it seems SSR doesn't work when you use a cache provider in client code.

My guess is we'd have to use the [advanced approach](https://emotion.sh/docs/ssr#advanced-approach) for SSR, and use a separate cache for normal and no script.

At which point it's a lot of refactoring to get that approach to work 😓

AshCorr: Ahhh... yea... Thats a shame, sorry to make you go down a rabbit hole!

coldlink: That's alright! It seemed a good solution!

PR #3091, opened 2025-03-04
refactor(typescript): use `erasableSyntaxOnly` flag for typechecking

coldlink:
In TypeScript 5.8 a `tsc` flag was introduced called [`--erasableSyntaxOnly`](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-5-8.html#the---erasablesyntaxonly-option).

When this flag is enabled, TypeScript will error on most TypeScript-specific constructs that affect runtime behaviour.

This means things like `enum` declarations, `namespace`s and `module`s with runtime code, parameter properties in classes, and non-ECMAScript `import =` and `export =` assignments don't work.

This is useful as it it allows for a step towards ["type stripping"](https://nodejs.org/api/typescript.html#type-stripping) which allows runtimes like nodejs to run typescript files by simply stripping any inline types without performing type checking or any other code transformation. Essentially allowing the runtimes to run typescript files without any build/transpilation steps!

The more immediate benefit for us is to enable developers to use typescript as a simple layer over javascript, rather than relying on typescript code that outputs generated code at compilaton such as enums. This should improve the developer experience and avoid common pitalls. By stripping types, it means the JavaScript output is as similar as possible to the originally authored TypeScript code.

See https://satanacchio.hashnode.dev/everything-you-need-to-know-about-nodejs-type-stripping as a useful blog post talking about this.

## Accessibility


src/client/models/Font.ts
```diff
@@ -5,34 +5,36 @@ export interface FontFace {
...
-	SANS = 'fonts/guardian-textsans/noalts-not-hinted/GuardianTextSans',
-}
+export const FontFamilyPath = {
+	TITLEPIECE:
+		'fonts/guardian-titlepiece/noalts-not-hinted/GTGuardianTitlepiece',
+	HEADLINE: 'fonts/guardian-headline/noalts-not-hinted/GHGuardianHeadline',
+	EGYPTIAN:
+		'fonts/guardian-textegyptian/noalts-not-hinted/GuardianTextEgyptian',
+	SANS: 'fonts/guardian-textsans/noalts-not-hinted/GuardianTextSans',
+};
```
AshCorr, line 40:
```suggestion
} as const;
```

src/shared/model/Errors.ts
```diff
@@ -1,97 +1,104 @@
...
+		'Your account has been created but there was a problem signing you in.',
+} as const;
 
 // shown at the top of the change password page when something goes wrong
-export enum ChangePasswordErrors {
-	GENERIC = 'There was a problem changing your password, please try again.',
-}
+export const ChangePasswordErrors = {
+	GENERIC: 'There was a problem changing your password, please try again.',
+};
```
AshCorr, line 45:
```suggestion
} as const;
```

AshCorr (dismissed):

AshCorr (approved):

PR #3092, opened 2025-03-04
Switch to `@swc/jest` from `ts-jest`

coldlink:
## What does this change?

We've been using [swc](https://swc.rs/) for a [long time](https://github.com/guardian/gateway/pull/2132) in Gateway to improve build times!

However we've been failing to utilise the same benefit when it comes to our unit testing.

This PR replaces [`ts-jest`](https://kulshekhar.github.io/ts-jest/) with the swc equivalent, [`@swc/jest`](https://swc.rs/docs/usage/jest).

As you can see from the table below, this is about a 3x speed up to running our unit tests (using `pnpm run test:unit`)!

<table>
<tr>
<th> ts-jest
<th> @swc/jest
<tr>
<td>

```
Test Suites: 48 passed, 48 total
Tests:       2 skipped, 436 passed, 438 total
Snapshots:   3 passed, 3 total
Time:        16.483 s
```

</td>

<td>

```
Test Suites: 48 passed, 48 total
Tests:       2 skipped, 436 passed, 438 total
Snapshots:   3 passed, 3 total
Time:        5.147 s
```

</td>
</table>

It also speeds up tests for `cdk` from ~4-5s to ~1s too! For `cdk` we could eventually move to use `swc` to compile cdk, but that's beyond the scope of this PR.

While it was a mostly drop in replacement, it did allow us to uncover some previous issues which were hidden.

Firstly we can now use `"jsx": "preserve",` instead of `"jsx": "react",` in our `tsconfig.json`. This allows us to use swc to compile our typescript react code rather than using the typescript compiler (`tsc`)!
This has the added benefit where we can now remove the `import React from 'react'` statement from the top of files. This has been a bugbear of mine, as this is only needed if we use tsc to compile code, where the only place we were doing this was for jest.
We'll remove these statements in another PR.

We also found a couple of `ReferenceError` issues, which have been solved in the comments below!

## Tested

- [x] CI
- [x] CODE


src/server/lib/okta/idx/challenge.ts
```diff
@@ -287,104 +275,6 @@ export const challengeResend = (
...
- * @param stateHandle - The state handle from the previous step
- * @param password - The password to set
- * @param expressReq - The express request object
- * @param expressRes - The express response object
- * @param introspectRemediation - The remediation object name to validate the introspect response against
- * @param path - The path of the page
- * @param ip - The ip address
- * @returns Promise<void> - Performs a express redirect
- */
-export const setPasswordAndRedirect = async ({
```
coldlink, line 302: This function was moved to `submitPasscode.ts` as it was causing circular dependencies, causing jest to error with `ReferenceError: Cannot access 'x' before initialization`

It doesn't make sense for this function to be in the challenge file anyway!

src/server/lib/__tests__/middleware/csrf.test.ts
```diff
@@ -11,16 +11,18 @@ const testToken =
...
+// Mock crypto module
 jest.mock('crypto', () => ({
 	createHmac: () => ({
 		update: () => ({
 			digest: digestSpy.mockReturnValue('hash'),
 		}),
 	}),
-	randomBytes: randomBytesSpy.mockReturnValue(
-		Buffer.from(testToken, 'base64url'),
-	),
```
coldlink, line 23: This was changed as we were getting an

`ReferenceError: Cannot access 'randomBytesSpy' before initialization`

error.

src/client/__tests__/MainForm.test.tsx
```diff
@@ -1,9 +1,7 @@
 /**
  * @jest-environment jsdom
  */
-/** @jsx jsx */
 
-import { jsx } from '@emotion/react';
```
coldlink, line 6: we no longer need this now that we use `"jsx": "preserve",` in our tsconfig!

PR #3107, opened 2025-03-18
Dependencies | 18th March 2025

coldlink:
## What does this change?

Updates dependencies where possible since #3100
 
Holds back the following:

## Tested

- [ ] CODE


cdk/package.json
```diff
@@ -16,12 +16,12 @@
...
     "@guardian/tsconfig": "^1.0.0",
-    "@swc/core": "^1.11.8",
+    "@swc/core": "^1.11.11",
     "@swc/jest": "^0.2.37",
     "@types/jest": "^29.5.14",
     "@types/node": "22.13.10",
-    "aws-cdk": "2.1003.0",
-    "aws-cdk-lib": "2.182.0",
+    "aws-cdk": "2.1004.0",
+    "aws-cdk-lib": "2.184.1",
```
AshCorr, line 24: I think `@guardian/cdk` specifically wants `aws-cdk` 2.178.2 based on its [peer dependencies](https://github.com/guardian/cdk/blob/main/package.json#L71-L72). Imo I feel like the @guardian/cdk package should relax its peer dependency range though as any incompatability issues would be highlighted by snapshot testing...

AshCorr (approved):

PR #3111, opened 2025-03-19
Set the toggle to true on New Account Review

akinsola-guardian:
## What does this change?
Shows the advertisng and profiling sections by default for non ad-free users.

## How to test

Tested in CODE.

### DIFF
--- a/cypress/integration/ete/new_account_review.3.cy.ts
+++ b/cypress/integration/ete/new_account_review.3.cy.ts
@@ -61,9 +61,7 @@ describe('New account review page', () => {
 				);
 				cy.contains('Advertising is a crucial source of our funding');
 
-				// Flip the switches so personalised advertising is on
-				// and profiling is off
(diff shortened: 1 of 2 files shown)


cypress/integration/ete/new_account_review.3.cy.ts
```diff
@@ -86,7 +86,7 @@ describe('New account review page', () => {
 									expect(consent.consented).to.be.true;
 								}
 								if (consent.id === 'personalised_advertising') {
-									expect(consent.consented).to.be.true;
+									expect(consent.consented).to.be.false;
```
coldlink, line 89: This was fine as it was.

The problem is on line 66 of the test file, where we click on the advertising checkbox.

With the change in this PR, we don't need that click anymore, as advertising is on by default, and we click to turn it off, whereas before advertising was off by default and we had to click to turn it on.

So we should remove this click, and update the comment

```suggestion
									expect(consent.consented).to.be.true;
```

coldlink (approved):

PR #3124, opened 2025-03-31
Add user-benefits-api

akinsola-guardian:
## What does this change?

This PR adds functionality to set the required product benefit cookies upon user sign-in and session refresh.

It retrieves the user’s product benefits by making a request to the `/benefits/me` endpoint of the user-benefits-api. The response returns an array of benefit identifiers associated with the user.

Based on the returned benefits, the following cookies are conditionally set: `gu_user_benefits_expiry`, `gu_hide_support_messaging`, `gu_allow_reject_all`, `GU_AF1`

This PR builds on prior work in Dotcom, referencing:
https://github.com/guardian/dotcom-rendering/blob/a0eff55ee7fedf08b785557fceaa0f4f7265e0df/dotcom-rendering/src/client/userFeatures/user-features.ts
https://github.com/guardian/support-service-lambdas/blob/main/modules/product-benefits/src/schemas.ts

This change introduces support for the new user-benefits-api, with the aim of eventually replacing reliance on the members-data-api.

https://docs.google.com/document/d/19ExTKgMhCJbrSLD9GFTeNKy3ArjoS2GR8wVc2ONbNXo/edit?tab=t.0


src/shared/lib/user-benefits-api.ts
```diff
@@ -0,0 +1,21 @@
+import { z } from 'zod';
```
coldlink, line 1: I was playing around in zod, and came across this better way of making the schema!

```ts
import z from 'zod';

// list of benefits we're looking for
const BenefitEnum = z.enum(['adFree', 'allowRejectAll', 'hideSupportMessaging']);

const responseSchema = z.object({
  benefits: z
    .array(z.string()) // Accept any array of strings
    .transform((values) => values.filter((val) => BenefitEnum.safeParse(val).success)) // Filter out unknown values based on the BenefitEnum
});

// Example usage
const response = {
  benefits: ['adFree', 'allowRejectAll', 'unknownBenefit'] // Contains an invalid value
};

const parsedResponse = responseSchema.parse(response); // Will remove 'unknownBenefit'
console.log(parsedResponse); // { benefits: [ 'adFree', 'allowRejectAll' ] }
```

akinsola-guardian: Thanks. I've implemented this

src/server/lib/user-benefits-api/user-benefits.ts
```diff
@@ -0,0 +1,37 @@
...
+import {
+	UserBenefitsResponse,
+	userBenefitsResponseSchema,
+} from '@/shared/lib/user-benefits-api';
+import { logger } from '../serverSideLogger';
+import { joinUrl } from '@guardian/libs';
+
+const { userBenefitsApiUrl } = getConfiguration();
+
+export const getUserBenefits = async ({
```
coldlink, line 11: I suggesting adding a comment about where this API lives, e.g. see the way the comments for the members-data-api is, where we link to the API, and a short description about the API:

https://github.com/guardian/gateway/blob/34311cdf8524f247a11841fbf7bdf9568f5829ea/src/server/lib/members-data-api/user-attributes.ts#L10-L24

src/server/lib/user-features.ts
```diff
@@ -1,13 +1,20 @@
...
 // port of some functionality from https://github.com/guardian/frontend/blob/main/static/src/javascripts/projects/common/modules/commercial/user-features.ts
 
 const { baseUri } = getConfiguration();
 
 const domain = `${baseUri.replace('profile.', '').split(':')[0]}`;
 
+export const AD_FREE_USER_COOKIE = 'GU_AF1';
+export const HIDE_SUPPORT_MESSAGING_COOKIE = 'gu_hide_support_messaging';
+export const ALLOW_REJECT_ALL_COOKIE = 'gu_allow_reject_all';
+export const USER_BENEFITS_EXPIRY_COOKIE = 'gu_user_benefits_expiry';
```
coldlink, line 16:
```suggestion
const AD_FREE_USER_COOKIE = 'GU_AF1';
const HIDE_SUPPORT_MESSAGING_COOKIE = 'gu_hide_support_messaging';
const ALLOW_REJECT_ALL_COOKIE = 'gu_allow_reject_all';
const USER_BENEFITS_EXPIRY_COOKIE = 'gu_user_benefits_expiry';
```

I don't think these need to be an export?

src/server/lib/user-features.ts
```diff
@@ -42,3 +60,47 @@ export const setUserFeatureCookies = async ({
 		});
 	}
 };
+
+export const createCookie = ({
```
coldlink, line 64:
```suggestion
const createCookie = ({
```

src/server/lib/user-features.ts
```diff
@@ -42,3 +60,47 @@ export const setUserFeatureCookies = async ({
...
+	daysTillExpiry: number;
+}) => {
+	const tmpDate = new Date(Date.now() + daysTillExpiry * 24 * 60 * 60 * 1000);
+	res.cookie(name, tmpDate.getTime().toString(), {
+		domain,
+		expires: tmpDate,
+	});
+};
+
+export const persistUserBenefitsCookies = ({
```
coldlink, line 80:
```suggestion
const persistUserBenefitsCookies = ({
```

src/shared/lib/user-benefits-api.ts
```diff
@@ -0,0 +1,18 @@
...
+	'adFree',
+	'allowRejectAll',
+	'hideSupportMessaging',
+]);
+
+export const userBenefitsResponseSchema = z.object({
+	benefits: z
+		.array(z.string()) // Accept any array of strings
+		.transform((values) =>
+			values.filter((val) => BenefitEnum.safeParse(val).success),
```
coldlink, line 14:
```suggestion
			values.filter((val): val is z.infer<typeof BenefitEnum> => BenefitEnum.safeParse(val).success))
```

The way you've implemented it wont actually make the type correct, it is still being parsed as `z.string()`. Using the suggestion I have parses the type correctly to

```ts
type ResponseType = { benefits: ("adFree" | "allowRejectAll" | "hideSupportMessaging")[]; }
```

src/shared/lib/user-benefits-api.ts
```diff
@@ -0,0 +1,18 @@
+import { z } from 'zod';
+
+// list of benefits we're looking for
+const BenefitEnum = z.enum([
+	'adFree',
+	'allowRejectAll',
+	'hideSupportMessaging',
+]);
```
coldlink, line 8: Can we get a comment detailing exactly where we're getitng this list from, e.g. is there something in DCR/user-benefits-api repo we can point to?

src/server/lib/user-features.ts
```diff
@@ -42,3 +60,47 @@ export const setUserFeatureCookies = async ({
...
+	name,
+	res,
+	daysTillExpiry,
+}: {
+	name: string;
+	res: Response;
+	daysTillExpiry: number;
+}) => {
+	const tmpDate = new Date(Date.now() + daysTillExpiry * 24 * 60 * 60 * 1000);
+	res.cookie(name, tmpDate.getTime().toString(), {
```
coldlink, line 74: Is this value coming from somewhere, e.g. DCR?
Can we get an comments explaining where this value is coming from.

src/server/lib/user-features.ts
```diff
@@ -29,6 +36,17 @@ export const setUserFeatureCookies = async ({
 		accessToken,
```
coldlink, line 36: Not related to this line, but can you update the comment/description of this function `setUserFeatureCookies` to include we're also pulling from the user-benefits-api now?

src/server/lib/user-features.ts
coldlink: e.g. as we've done previously for the old way of getting the ad-free cookie:

https://github.com/guardian/gateway/blob/34311cdf8524f247a11841fbf7bdf9568f5829ea/src/server/lib/user-features.ts#L34-L42

src/server/lib/user-features.ts
```diff
@@ -42,3 +60,47 @@ export const setUserFeatureCookies = async ({
...
+			res,
+			daysTillExpiry: 7,
+		});
+	}
+	if (userBenefits?.benefits?.includes('allowRejectAll')) {
+		createCookie({ name: ALLOW_REJECT_ALL_COOKIE, res, daysTillExpiry: 7 });
+	}
+	if (userBenefits?.benefits?.includes('adFree')) {
+		createCookie({ name: AD_FREE_USER_COOKIE, res, daysTillExpiry: 2 });
+	}
```
coldlink, line 99: For each of these benefits, can we add a comment that link to the code in the DCR repo, again like we did previosuly for members-data-api:

https://github.com/guardian/gateway/blob/34311cdf8524f247a11841fbf7bdf9568f5829ea/src/server/lib/user-features.ts#L34-L42

src/server/lib/user-features.ts
```diff
@@ -42,3 +60,47 @@ export const setUserFeatureCookies = async ({
...
+
+	createCookie({
+		name: USER_BENEFITS_EXPIRY_COOKIE,
+		res,
+		daysTillExpiry: 7,
+	});
```
coldlink, line 105: And a comment for this cookie as well to link to it's correstonding definition in DCR!

I also think that we should only add this cookie only if we have at least 1 benefit we're setting, if we're not setting any benefits, then I don't think we need to add this?

src/server/lib/user-features.ts
coldlink: I think we should also add a comment to the corresponding secion in the DCR repo saying that if new cookies/benefits are added/removed there, that this section in gateway should also be updated:

https://github.com/guardian/dotcom-rendering/blame/a0eff55ee7fedf08b785557fceaa0f4f7265e0df/dotcom-rendering/src/client/userFeatures/user-features.ts#L38-L49

coldlink (changes requested):
Looking good so far!

Here's my initial suggestions for changes.

Also if we only need the user-benefits-api now, do we still need to keep the members-data-api code and the way we set the ad-free cookie there?

coldlink:
Also can you update the PR description to include a more detailed description about *what* is changing, and also the *why* we're making this change.

The initial members-data-api pr is an okay example: https://github.com/guardian/gateway/pull/2016

src/server/lib/user-features.ts
akinsola-guardian: This need to be set once the request o user benefits is made. 

In a scenario where a user doesn't have any benefits. We still want to drop the cookie so it doesn't make a request every time they view the site. The user benefits expires after 24 hours. https://github.com/guardian/dotcom-rendering/blob/a0eff55ee7fedf08b785557fceaa0f4f7265e0df/dotcom-rendering/src/client/userFeatures/user-features.ts#L38

coldlink (approved):
Great work!

PR #3127, opened 2025-04-03
Passwordless | Remove set password on account creation

coldlink:
## What does this change?

Removes the need for the user to set a password on account creation, thus making all new accounts passwordless! Once the user verifies their account with a passcode, they're signed in, that's it!

Users will still manually be able to opt in to passwords if they go through the create account flow.

A `useSetPassword` flag has also been added so we can preserve and test the legacy behaviour if needed.

## Changes

Adds a `POST /idp/idx/skip` method, which can be called when there is a `skip` remediation object in the IDX API response.

This is used to tell the IDX API to skip the current step if the remediation exists.

Since we've made passwords optional in Okta for users, in https://github.com/guardian/identity-platform/pull/806, the IDX API returns a `skip` remediation after the user verifies their email with a passcode.

This allows us to skip the `select-authenticator-enroll` remediation, which we'd normally do to set a password, which will no longer be the default.

To fascilitate this behaviour, we've added a `skipPasswordEnrollment` method, which will call the `skip` endpoint, make sure the user is authenticated (`isChallengeAnswerCompleteLoginResponse`), set the `emailValidated` flag (and curcially **NOT** the `passwordSetSecurely` flag), and finally redirect the user to the `loginRedirectUrl` which set's the user session and redirects the user back to where they started the authentication process from.

In the Gateway `POST /register/code` controller, we now check for `hasSkipSchema`, and if so (and the `useSetPassword` is not set), we call the `skipPasswordEnrollment` method.

We also update the `PasscodeUsed` page to make sure that if the user accidentally clicks back in their browser after setting a passcode, the CTA on that page will direct the user correctly.

Finally tests and documentation were updated to take into account this change.

## Videos

<table>
<tr>
<th> Mobile - Live App Android
<th> Web - Desktop
<tr>
<td>

[Screen_recording_20250403_152749.webm](https://github.com/user-attachments/assets/faedd3f4-b648-45d2-b33e-559fe3f707e6)

<td>

https://github.com/user-attachments/assets/768e5e4e-5457-4c5b-864b-a2b97873b804

</table>

## Tested

- [x] CODE


AshCorr (approved):
LGTM!

The app version looks a bit jarring to me due to the lack of Welcome/Thank you page though

coldlink:
> LGTM!
> 
> The app version looks a bit jarring to me due to the lack of Welcome/Thank you page though

Yeah, unfortunately it's also the current behaviour with passwords, after a user sets a password they don't currently see the `/welcome/review` screen either. I think it was a product and techinical decision at the time, but it's possible to change.

It's a quirk with the way that Gateway works when coming from the Okta hosted login page, where we rely on a `fromURI` parameter provided to us by the hosted login page, and then Gateway has to redirect the user there when the user finishes authentication.

There are ways around this, we probably just have to persist the `fromURI` parameter for a bit longer, i.e until the user interacts with the welcome thank you page, rather than just redirecting them straight away to `fromURI`.

AshCorr (approved):

src/server/lib/okta/idx/shared/submitPasscode.ts
```diff
@@ -221,3 +222,72 @@ export const setPasswordAndRedirect = async ({
...
+			id,
+			ip,
+			updateEmail: true,
+			updatePassword: false,
+		});
+	} else {
+		logger.error(
+			'Failed to set validation flags in Okta as there was no id',
+			undefined,
+		);
```
AshCorr, line 272: Should we throw an exception here?

coldlink: hmm.. I implemented this in the similar way to how the `submitPassword`/`submitPasscode` functions were set up in this same file, which don't throw either.

But i'm wondering if this error is even relevant, it should'nt even be possible to get here if a user didn't have an Okta id!

coldlink: Yeah I've just checked the code and we don't need it! `id` is defined! If `id` is missing it would throw automatically as the zod parsing would've failed.

coldlink: Have removed the `if` check.

AshCorr (approved):

PR #3136, opened 2025-04-08
Revert "Add user-benefits-api" (#3124)

coldlink:
Reverts guardian/gateway#3124

We were seeing a lot of "Error: UserBenefitsAPI returned 403 Forbidden" errors, and the user benefits cookies weren't being set, only the standard IDAPI cookies were.


coldlink:
Closing as it was an incorrect parameter causing the errors

PR #3158, opened 2025-05-02
Log authentication metrics

coldlink:
## What does this change?

Currently it's difficult for us to know exactly which flow and app a user used to authenticate with the Guardian.

To remediate this problem, this PR adds a new log line in the OAuth Authentication Callback so we can track this. using the `extraData` object we can pass to the server side logger, and thus kibana.

However we first need to be able to know which flows we're using. We can do this by adding additional information to the Authorization Code Flow state cookie when a user comes to Gateway. In the `AuthorizationState` we add the `flow` and `appLabel` parameter, which allow us to track the flow the user is authenticating with, and also which app they're authenticating from.

We then make sure that the authorization state in the IDX API `interact` call takes into account these two new parameters and persists them. From there we make sure that at the start of each flow this information is available, i.e at the start of the sign in flow (with passcode and password), social authentication, create account, and reset password (account recovery).

Finally just before we redirect the user in the OAuth Authentication Callback, we add the log messaging in.

`OAuth Authentication via Gateway flow` - is the message used when we know the user authenticated via one of the gateway flows, this will include the `flow` and `appLabel` if available.

`OAuth Authentication without Gateway flow` - is authentication flows that didn't require the user to reauthenticate, e.g. sign in refresh flows, token refreshes etc.

When successful, the event will look like the following in the local gateway logs

```
{"appLabel":"<appLabel>","flow":"sign-in-passcode","ip":"<ip>","level":"info","message":"OAuth Authentication via Gateway flow","requestId":"<requestId>"}
```

You can then query it in Grafana too, with an example lucene query being:

```
app.keyword:identity-gateway AND stage:$stage AND message:"OAuth Authentication via Gateway flow" AND flow:"<flow>" AND appLabel:<appLabel>
```

Possible flows include

```js
'sign-in-passcode'
'sign-in-password'
'social-registration' // the idp parameter can be used to differentiate google/apple
'social-signin' // the idp parameter can be used to differentiate google/apple
'create-account'
'account-recovery' // password reset, and email verification flows outside of create account
```

We also added new metrics with the following structure, so we can have a longer term view of the data (as logs are only available for 12 days). While these metrics will cost a little extra, it won't be as many as if we were to add the app to every metric.

```
`UserFlow-${flow}-${app}`
e.g.
`UserFlow-sign-in-passcode-android_live_app`
`UserFlow-social-registration-apple-ios_feast_app`
```

## Tested

- [x] CODE - `sign-in-passcode`
- [x] CODE - `sign-in-password`
- [x] CODE - `social-registration`
- [x] CODE - `social-sign-in`
- [x] CODE - `create-account`
- [x] CODE - `account-recovery`

Screenshot 2025-05-02 at 16 07 21

### DIFF
--- a/src/server/controllers/sendChangePasswordEmail.ts
+++ b/src/server/controllers/sendChangePasswordEmail.ts
@@ -123,6 +123,10 @@ export const changePasswordEmailIdx = async ({
 				confirmationPagePath: user.credentials.password
 					? '/reset-password/complete'
 					: '/set-password/complete',
+				extraData: {
+					flow: 'account-recovery',
+					appLabel: res.locals.appLabel,
+				},
 			},
 		});
 
--- a/src/server/controllers/signInControllers.ts
+++ b/src/server/controllers/signInControllers.ts
@@ -243,6 +243,10 @@ export const oktaIdxApiSignInPasscodeController = async ({
 					res,
 					authorizationCodeFlowOptions: {
 						confirmationPagePath,
+						extraData: {
+							flow: 'sign-in-passcode',
+							appLabel: res.locals.appLabel,
+						},
 					},
 				});
 
@@ -446,6 +450,12 @@ export const oktaIdxApiSignInController = async ({
 			email,
 			req,
 			res,
+			authorizationCodeFlowOptions: {
+				extraData: {
+					flow: 'sign-in-password',
+					appLabel: res.locals.appLabel,
+				},
+			},
 		});
 
 		// check for the "password" authenticator, we can only authenticate with a password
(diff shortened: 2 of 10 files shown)


coldlink:
> > The downside to this is that we'll effectively be doubling the number of metrics to Cloudwatch, which will incur an increase to overall monthly cloudwatch costs, which is diffucult to estimate using cost explorer.
> 
> Wouldn't we be tripling the number of metrics? Gateway, Android, and iOS? Believe we have ~170 metrics at the moment costing $50 a month, so with this change it'll increase to about $150?
> 
> Would Feast count as another 2 client ids? Wonder if we can be a bit more targetted on which metrics include the appClientId?

Actually that's true, and I think with each app it would qualify as another metric, so we'd be tripling it multiplied by the number of apps we have

coldlink:
I've now refactored this to track only specifc flows using logging, which we can then monitor in Grafana

akinsola-guardian (approved):

PR #3187, opened 2025-06-03
feat(jobs): Notify users that a Jobs account will be created after continuing

AshCorr:
## What does this change?

Now that we're adding `?prompt=login` to Jobs users more of them will be seeing the "Signed in as" page when logging in. Add a message to this page to let them know that by continuing they'll create a Guardian Jobs account.

## How to test

Deployed to CODE and checked the Jobs login flow whilst signed in.

| Before | After |
|--------|--------|
| Screen Shot 2025-06-03 at 15 22 37 | Screen Shot 2025-06-09 at 18 18 06 |

### DIFF
--- a/src/client/pages/SignedInAs.stories.tsx
+++ b/src/client/pages/SignedInAs.stories.tsx
@@ -31,6 +31,16 @@ export const FeastApp = () => (
 	/>
 );
 
+export const Jobs = () => (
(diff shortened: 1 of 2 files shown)


pvighi (dismissed):
looks good

pvighi (approved):
looks good to me

PR #3211, opened 2025-06-23
chore(okta): Specify hosted login page JS hash

AshCorr:
We use jsdelivr to host the JS bundle thats used on the Okta hosted login page, the way this service works is that it directly serves the contents of this repository to the end user.

The way jsdelivr is currently setup is to serve the latest code on the `main` branch, this is great for simplicity but because we don't use a version string it means it can take up to 7 days for an update to be reflected on end user devices due to caching.

When we merge https://github.com/guardian/identity-platform/pull/824 any users running outdated JS will end up stuck in a login loop when logging into the Jobs application. To avoid this lets point jsdelivr to the latest commit on main, instead of main itself, this should bust the cache.

Its worth noting that it is possible to purge the CDN cache on jsdelivr via https://www.jsdelivr.com/tools/purge but this does not clear the local browser cache.
## How to test

Tested in CODE.

image

### DIFF
--- a/scripts/okta/okta-login.html
+++ b/scripts/okta/okta-login.html
@@ -60,6 +60,9 @@
 		</noscript>
 		<div id="okta-login-container"></div>
 		{{{OktaUtil}}}
-		<script src="https://cdn.jsdelivr.net/gh/guardian/gateway@main/scripts/okta/okta-login.min.js"></script>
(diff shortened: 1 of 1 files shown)


pvighi (approved):
looks... 

good?

PR #3231, opened 2025-07-15
Feast App creation Soft opt in (SOI) and terms copy

rBangay:
## What does this change?
Shows the marketing consents toggle on the feast app creation page alongside the feast newsletter consent toggle.

Also shows a new paragraph in the terms (below the consents toggles) for users of all products.

## Images

Before            |  After
:-------------------------:|:-------------------------:
  |


src/client/components/MainForm.tsx
```diff
@@ -309,6 +309,10 @@ export const MainForm = ({
 					{additionalTerms && (
 						<InformationBoxText>{additionalTerms}</InformationBoxText>
 					)}
+					<InformationBoxText>
+						You can change your settings in the Data Privacy section of your
+						Guardian account at any time.
+					</InformationBoxText>
```
rBangay, line 315: This additional terms paragraph gets shown for everyone

src/client/components/RegistrationConsents.tsx
```diff
@@ -102,3 +92,6 @@ export const RegistrationConsents = ({
 		</ToggleSwitchList>
 	);
 };
+
+export const marketingConsentTerms =
+	'You can change your settings in the Data Privacy section of your Guardian account at any time.';
```
AshCorr, line 97: Do we want to link to the Data Privacy settings like we do [here](https://github.com/guardian/gateway/blob/89308d432894a003ce301116f29cb5e159f58147/src/client/pages/NewAccountReview.tsx#L148-L150)?

AshCorr (dismissed):

src/client/components/RegistrationConsents.tsx
theemilyturner: I don't think we can, because at this point they haven't got an account, so the link won't take them there.

cypress/integration/ete/registration_3.8.cy.ts
```diff
@@ -618,14 +618,15 @@ describe('Registration flow - Split 3/3', () => {
...
-				cy.interceptRecaptcha(3);
+				cy.failRecaptchaRequest();
 				cy.wait(1000); // wait for the send again button to be enabled
 				cy.contains('send again').click();
 				cy.contains('Google reCAPTCHA verification failed.');
 				cy.contains('If the problem persists please try the following:');
 
 				const timeRequestWasMade = new Date();
 				cy.wait(1000); // wait for the send again button to be enabled
+				cy.reEnableRecaptchaRequest();
```
rBangay, line 629: Slight refactor of how this test handles recaptca requests .. the test is testing what happens when recpatcha fails and then subsequently works a second time. Instead of intercepting the recpatcha request a set number of times (in this case 3) and returning a 500 response then expecting the 4th request to pass, we now intercept any number of requests and return a 500 up until the point at which we expect the recpatcha to work again where we re-enable the request so that it can pass.

cypress/support/commands/recaptcha.ts
```diff
@@ -3,22 +3,46 @@ declare global {
...
 			method: 'POST',
 			url: 'https://www.google.com/recaptcha/api2/**',
-			times,
+			...(typeof times === 'number' && { times }),
 		},
 		{
 			statusCode: 500,
 		},
 	);
 };
```
rBangay, line 26: This was part of an investigation into the recaptcha request interceptions. Although this new functionality isn't used by the any of the tests, I thought it might be useful to someone in the future so I've left it in here. Basically the test will default to intercept the request one time, you can override that by passing a number to reflect the number of times you'd like the request to be intercepted, or pass through the string 'indefinate' which will intercept any number of requests.

AshCorr (approved):

PR #3233, opened 2025-07-18
add more workers to the cypress e2e tests

rBangay:
## What does this change?

The general aim of this pr is to help make the cypress tests less flaky when run from CI eg we would like them to pass without having to re-run them ... with this in mind the following changes have been made:

- Adds more workers to the cypress e2e tests, bringing the new value up from 8 to 11.
In order to do this I have split the `sign_in_passcode` tests into 2 separate suites, one testing active users and the other testing non active users. 

- Refactored/moved the `sendEmailAndValidatePasscode` function from the original `sign_in_passcode` test into a shared cypress command that can be used by both the active and non active users tests.

- Split the registration e2e tests into 3 steps from 2 and ensured that they run in a dedicated worker in github.

- Change the number of times we intercept recaptcha requests in the registration_3.8.cy.ts test from 1 to 3 in order to catch any request retries. Thanks to @AshCorr for this suggestion :)


cypress/integration/ete/registration_3.8.cy.ts
```diff
@@ -1235,7 +618,7 @@ describe('Registration flow - Split 2/2', () => {
 					timeRequestWasMadeInitialEmail,
 				);
 
-				cy.interceptRecaptcha();
+				cy.interceptRecaptcha(3);
```
rBangay, line 621: Here is where we increase the number of times we intercept recaptcha requests from 1 to 3 ... this is so that we also catch any retry attempts if they happen.

cypress/support/commands/sendEmailAndValidatePasscode.ts
```diff
@@ -0,0 +1,121 @@
...
+/**
+ * Send an email, receive the passcode and validate it on the sign in page
+ * @param emailAddress The email address you want the passcode to be sent to
+ * @param expectedReturnUrl  The url you are returned to after signing in
+ * @param params URL params added to the sign in url
+ * @param expectedEmailBody  What do you want the email body copy to say
+ * @param additionalTests  Additional tests/checks that you might want to chek for as well as signining in
+ * @returns A chainable cypress command.
+ */
+export const sendEmailAndValidatePasscode = ({
```
rBangay, line 23: Moved this out into a shared cypress command so that it can be used by separate github workers now that the test has been split up into multiple files.

AshCorr (approved):

PR #3177, opened 2025-05-27
chore(design): Remove "welcome" image and center pages vertically.

AshCorr:
## What does this change?

image

We've been requested to remove the "slender man" image from the Welcome and New Account page. This PR does that and centers those pages vertically now to make use of the extra white space available on the page and avoid changing the layout too much.

## Images

| Before | After |
|--------|--------|
| Screen Shot 2025-05-27 at 16 08 57 | Screen Shot 2025-05-27 at 16 07 00 |
| Screen Shot 2025-05-27 at 16 08 44 | Screen Shot 2025-05-27 at 16 06 38 |

PR #3247, opened 2025-07-30
feat: Revert the removal of the welcome image and replace with a new one

AshCorr:
## What does this change?

Re-adds the welcome image to the registration flow but replaces it with a new illustration.

## How to test

Ran locally

| Light | Dark |
|--------|--------|
| image | image |


src/client/layouts/MinimalLayout.tsx
```diff
@@ -27,19 +27,18 @@ interface MinimalLayoutProps {
 	pageHeader: string;
 	leadText?: React.ReactNode;
 	imageId?: DecorativeImageId;
-	centered?: boolean;
```
AshCorr, line 30: Adding `centered` was a change made in https://github.com/guardian/gateway/pull/3177 to make the welcome page look a bit less odd without an illustration and now it is no longer required.

rBangay (approved):

PR #3253, opened 2025-08-11
chore: Bump dependencies 11th of August 2025

AshCorr:
## What does this change?

Bumps dependencies to their latest versions except:

- @types/express - Stick to v4, until we upgrade to Express v5
- openid-client - v6 is a major version change, will investigate separately, v5 still supported
- storybook - v9 is a major version, with significant changes

Also Bumps jest to v30 which included some breaking changes to how we import Preact in our tests.

## How to test

Deployed to CODE, created an account, signed out, and signed in.


jest.config.js
```diff
@@ -24,8 +18,16 @@ module.exports = {
...
+		// Don't try to transform any node_modules except Preact and @testing-library/preact
+		// Since updating to Jest 30 trying to import these dependencies has resulted in
+		// unsupported ESM getting imported instead of CJS. Possibly because due to
+		// some siginificant changes in how Jest handles ESM under the hood?
+		// See:  https://jestjs.io/docs/upgrading-to-jest30#esm-module-support-and-internal-restructuring
+		'/node_modules/.pnpm/(?!(preact|@testing-library))',
+	],
 	transform: {
-		'^.+\\.(t|j)sx?$': [
+		'^.+\\.(ts|tsx|jsx|mjs|module.js)?$': [
```
AshCorr, line 30: We now need to transform some of code in the `preact` and `@testing-library/preact` packages which contain `mjs` and `module.js` with unsupported ESM code.

These packages have a CJS version which we were previously using, but for some reason when we updated to Jest 30 it broke our imports and caused them to start using the ESM version instead of the CJS versions. My guess its something to do with [Jests internal changes to how it handles ESM under the hood](https://jestjs.io/docs/upgrading-to-jest30#esm-module-support-and-internal-restructuring) caused it to start importing the wrong modules.

jest.config.js
```diff
@@ -9,12 +9,6 @@ module.exports = {
 		'@/([^\\.]*)$': '<rootDir>/src/$1',
 		'\\.(jpg|ico|jpeg|png|gif|eot|otf|webp|svg|ttf|woff|woff2|mp4|webm|wav|mp3|m4a|aac|oga)$':
 			'<rootDir>/__mocks__/fileMock.js',
-		// need to alias preact to itself in order to use the cjs version, same with testing-library
```
AshCorr, line 12: Alias or not Jest does not want to use the CJS version no matter what I tried. Instead I've resorted to just using the ESM version and transforming it to CJS with `@swc/jest`.

rBangay (dismissed):
Thanks for looking at this Ash ... just a couple of questions and some flaky cypress tests :s but happy to approve :)

src/client/lib/hooks/__tests__/useRecaptcha.test.tsx
```diff
@@ -138,18 +137,16 @@ test('should receive a valid token back when the reCAPTCHA check is successful',
...
-			expect(recaptchaExecutionResult).toBe(false);
-		});
+	const { result } = renderHook(() =>
+		useRecaptcha('public-recaptcha-token', 'render-element'),
+	);
+
+	await waitFor(() => {
+		// Request a token from recaptcha.
+		const recaptchaExecutionResult = result.current?.executeCaptcha();
+		expect(recaptchaExecutionResult).toBe(false);
```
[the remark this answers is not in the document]
AshCorr: I'm not an expert on React, but from what I understand in the docs you're meant to wrap code that renders react components in it, and then run assertions afterwards https://react.dev/reference/react/act#await-act-async-actfn as it apparently makes sure that any updates to the components are complete before the assertions run.

In this case I removed it because we're running the assertions inside the `act` which I believe defeats the point? I think renderHook also wraps [`act`](https://testing-library.com/docs/preact-testing-library/api#act), but the nail in the coffin is that we're voiding the promise returned by `act` which was causing the test to complete before the promise had finalized resulting in effects of this test having impacts on the next tests.

PR #3257, opened 2025-09-01
chore: Replace `ophan-tracker-js` with `@guardian/ophan-tracker-js`

AshCorr:
## What does this change?

Replace the client-side `ophan-tracker-js` library with `@guardian/ophan-tracker-js`. The new library is only 1 version ahead, so it should be a pretty minimal update.

On the server-side we [manually call Ophans API](https://github.com/guardian/gateway/blob/main/src/server/lib/ophan.ts#L133-L163) instead of using the `ophan-tracker-js` library, so we did not need to fix any imports.

## How to test

Ran locally and verified that Ophan was still recording events:

image

I did notice however that upgrading to `@guardian/ophan-tracker-js` does seem to break tracking as it ceases to record events (or atleast I'm not seeing them in the network traffic), I tried to upgrade the library in this PR but couldn't get it to work which is something to keep in mind when upgrading dependencies.

### DIFF
--- a/src/client/static/analytics/ophan.ts
+++ b/src/client/static/analytics/ophan.ts
@@ -22,7 +22,7 @@ const addRefToOphanFollow = () => {
 
 addRefToOphanFollow();
 
-import 'ophan-tracker-js';
+import '@guardian/ophan-tracker-js';
 
 export const init = () => {
 	record({
--- a/package.json
+++ b/package.json
@@ -140,7 +140,7 @@
 		"mjml-browser": "^4.15.3",
 		"ms": "^2.1.3",
 		"openid-client": "^5.7.1",
-		"ophan-tracker-js": "^2.0.2",
+		"@guardian/ophan-tracker-js": "2.0.3",
 		"preact": "^10.27.1",
 		"preact-render-to-string": "^6.6.1",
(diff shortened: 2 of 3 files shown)


src/client/static/analytics/ophan.ts
```diff
@@ -22,7 +22,7 @@ const addRefToOphanFollow = () => {
 
 addRefToOphanFollow();
 
-import 'ophan-tracker-js';
+import '@guardian/ophan-tracker-js';
```
rBangay, line 25: It looks like we still using record from the older import `import { record } from '@/client/lib/ophan';` at the top ... could we do something like this instead:

```
import ophan from '@guardian/ophan-tracker-js';

export const init = () => {
  ophan.record({
```

AshCorr: I had a look at this and it didn't seem as straightforward as it looked! Unforunately theres 2 issues that I found:

1. The version of `@guardian/ophan-tracker-js` is tool old to use the newer import format.
2. Upgrading to the newer version of `@guardian/ophan-tracker-js` silently broke tracking and uses a different schema for events than we do. Even after switching to the new imports tracking was still silently broken.

PR #3259, opened 2025-09-02
feat(google-one-tap): Support `signInEmail` query parameter in Social Sign In to set `login_hint` parameter in Google Oauth flow

AshCorr:
## What does this change?

Allows setting the `signInEmail` query parameter on the `/signin/:social` route to set the `login_hint` parameter in the Google Social Sign In flow. This is used by Google One Tap to automatically log the user into the Guardian using their Google account.

Additionally adds documentation about the approach we took for Google One Tap.

## How to test

Deployed to CODE and navigated to https://profile.theguardian.com/signin/google?signInEmail=ashleigh.carr@guardian.co.uk


docs/google-one-tap.md
```diff
@@ -0,0 +1,134 @@
...
+		// such as One Time Passwords and Public Keys, but in our case we want FedCM credentials which is configured
+		// via the `identity` property
+		identity: {
+			providers: [
+				// In theory the FedCM supports providers other than Google so you could have a "Apple One Tap" or a "Guardian One Tap"
+				// listed here too, but I'm not entirely sure how well this works in practice.
+				{
+					// Couldn't find this endpoint documented anywhere but discovered it
+					// by observing GSI network traffic.
+					configURL: 'https://accounts.google.com/gsi/fedcm.json',
```
pvighi, line 28: how safe is it to use this endpoint if it's not documented anywhere ? what would happen if they suddenly change it ?

src/server/routes/signIn.ts
```diff
@@ -603,6 +604,24 @@ router.get(
...
+				idp,
+				closeExistingSession: true,
+				scopes: scopesForAuthentication,
+				redirectUri: ProfileOpenIdClientRedirectUris.AUTHENTICATION,
+				loginHint: signInEmail,
+				extraData: {
+					socialProvider: socialIdp,
+				},
+			});
+		}
```
pvighi, line 623: this is in `SignIn.ts` but it would also result in an account created if the user doesn't already have one, right ?

docs/google-one-tap.md
AshCorr: Google One Tap would stop working. However the [config.json endpoint is part of the FedCM standard](https://www.w3.org/TR/fedcm/#idp-api-config-file), so if they changed the endpoint they would still need to meet the standard. 

Im going to be adding some metrics when I do the follow up PR to set the users registration platform, so we'll be able to fire an alarm if for some reason people stop signing in / registering with Google One Tap

pvighi (approved):

PR #3267, opened 2025-09-16
Update the comment for the ad free cookie expiry in user-features

cemms1:
## What does this change?

Updates the comment referencing the two day expiry for the ad free cookie in user features

It used to point to the client-side code in `frontend` as this is where it was copied from. 
It now references the `dotcom-rendering` code since this is the code most likely to be running in the browser.

## Why

We don't use `frontend` to render our pages much any more and as such, we rarely run the client-side code.

To help future developers understand what the code is doing in various places in the codebase, it is more helpful to direct them to the code that is most likely to be running on users' machines (ie the better source of truth) rather than the reference of where the code originally came from.


rBangay (approved):
👍

PR #3362, opened 2026-04-09
Supply Chain Hardening + Dependency Updates

AshCorr:
## What does this change?

As per DevX Security's latest email about hardening our supply chain: 

> Project Config and CI
> - In your package.json, pin to a specific version of all your dependencies. Do not allow your package manager to determine a range by using ~ or ^. **Done**
> - Store your lockfile in version control and run npm ci ([clean install](https://docs.npmjs.com/cli/v11/commands/npm-ci)) in CI. This ensures a clean install that matches the lockfile. It can take a lot longer than a regular install, so use your judgement locally. **Was already the case**
> - Utilise minimumReleaseAge configurations, so that new versions of dependencies have a chance to be vetted. We recommend at least 48 hours. These are available in npm, pnpm, and bun. **We need to create a PNPM workspace to do this which breaks our builds**
> 
> External Config
> - Configure cooldowns in Dependabot. This works similarly to minimumReleaseAge above. [Example](https://github.com/guardian/service-catalogue/blob/c4530a26d2260108da50025d210c44479469836b/.github/dependabot.yml#L49). [Docs](https://github.blog/changelog/2025-07-01-dependabot-supports-configuration-of-a-minimum-package-age/#whats-new). **Done for Github Actions, we don't use Dependabot for NPM**
> - Avoid installing dependencies on the fly, for example by using npx. Its default behaviour is to install the most recent version, which may be malicious. **We dont use NPX**
> - Exercise caution when choosing less popular libraries, malicious releases may take much longer to be noticed **We should probably review our dependencies at some point, we have a LOT.**
>
> Gently recommended:
> - Consider disabling lifecycle (i.e preinstall/postinstall) scripts in your projects. They are extremely useful for developers, but equally useful to hackers. This might break things. This can be tricky in npm projects, as it’s an all or nothing switch. pnpm’s support for this is much better, as it’s enabled by default and has granular controls. **We rely on postinstall scripts atm, I might raise a follow up PR to address this control specifically**

Deployed to CODE


cdk/pnpm-lock.yaml
```diff
@@ -2967,11 +2883,11 @@ packages:
...
     engines: {node: '>= 0.4'}
 
-  typescript-eslint@8.55.0:
-    resolution: {integrity: sha512-HE4wj+r5lmDVS9gdaN0/+iqNvPZwGfnJ5lZuz7s5vLlg9ODw0bIiiETaios9LvFI1U94/VBXGm3CB2Y5cNFMpw==}
+  typescript-eslint@8.57.1:
+    resolution: {integrity: sha512-fLvZWf+cAGw3tqMCYzGIU6yR8K+Y9NT2z23RwOjlNFF2HwSB3KhdEFI5lSBv8tNmFkkBShSjsCjzx1vahZfISA==}
     engines: {node: ^18.18.0 || ^20.9.0 || >=21.1.0}
     peerDependencies:
-      eslint: ^8.57.0 || ^9.0.0
+      eslint: ^8.57.0 || ^9.0.0 || ^10.0.0
```
pvighi, line 2890: I noticed there still some dependencies referenced with ^ here in the lock file.. is this expected?

pvighi (approved):
looks good to me

cdk/pnpm-lock.yaml
AshCorr: That is expected yeah, libraries we use can still rely on ranges for their transitive dependencies

PR #3370, opened 2026-04-20
chore: Surpress various Gateway errors

AshCorr:
## What does this change?

We got a critical alarm trigger last night about not getting any registrations for 5 minutes, I don't think it was anything particularly concerning, but trying to investigate the cause was incredibly difficult due to the sheer number of errors in the Gateway logs.

Most Gateway errors that are being logged are completely normal parts of the registration/sign-in flow, or user error, and should either be logged as warnings or not logged at all, this PR cleans up various areas where we log these errors to hopefully reduce the number of errors we see in Gateway and make it easier to diagnose alarms.

This pull request improves error handling and logging throughout the codebase, especially for user errors.

Deployed to CODE.

### DIFF
--- a/src/server/controllers/changePassword.ts
+++ b/src/server/controllers/changePassword.ts
@@ -64,7 +64,7 @@ const oktaIdxApiPasswordHandler = async ({
 	password,
 	path,
 }: {
-	req: Request<{ token: string }>;
+	req: Request<{ token?: string }>;
 	res: ResponseWithRequestState;
 	pageTitle: PasswordPageTitle;
 	password: string;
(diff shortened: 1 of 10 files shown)


src/server/controllers/changePassword.ts
```diff
@@ -64,7 +64,7 @@ const oktaIdxApiPasswordHandler = async ({
 	password,
 	path,
 }: {
-	req: Request<{ token: string }>;
+	req: Request<{ token?: string }>;
```
AshCorr, line 67: We were previously typecasting the `token` query param to always be a string, which is false, because it will be `undefined` if its not set.

src/server/lib/idapi/consents.ts
```diff
@@ -70,7 +76,13 @@ const readUserConsents = async ({
 			options,
 		})) as UserConsent[];
 	} catch (error) {
-		logger.error(`IDAPI Error consents read '/users/me/consents'`, error);
+		const isWarning =
```
akinsola-guardian: Is it worth creating a function for `isWarning`?

akinsola-guardian (dismissed):

akinsola-guardian (approved):

PR #3402, opened 2026-04-27
Passcode sent wording change

ioannakok:
## What does this change?
Changes the copy of the passcode sent page to "Please check your inbox in a separate window" when user triggers this flow on the web. 

## Why?
Our less tech-savvy users sometimes try to retrieve their passcode by going to their email in the same window as the passcode page, and lose the page and repeat the flow.

## Testing

|                     | Before | After |
|---------------------|--------|-------|
| Sign In             | image | image |
| Reset password      | image | image |
| Send passcode again | image | image |

### DIFF
--- a/src/client/pages/IframedPasscodeEmailSent.tsx
+++ b/src/client/pages/IframedPasscodeEmailSent.tsx
@@ -38,7 +38,7 @@ const text: Text = {
 	title: 'Enter your code',
 	sentTextWithEmail: 'We’ve sent a temporary verification code to',
 	sentTextWithoutEmail:
-		'We’ve sent you a temporary verification code. Please check your inbox.',
+		'We’ve sent you a temporary verification code. Please check your inbox in a separate window.',
 	securityText:
 		'For your security, the verification code will expire in 30 minutes.',
 	passcodeInputLabel: 'Verification code',
@@ -193,7 +193,7 @@ export const IframedPasscodeEmailSent = ({
 			{email ? (
 				<MainBodyText>
 					{text.sentTextWithEmail} <strong>{email}</strong>. Please check your
-					inbox.
+					inbox in a separate window.
 				</MainBodyText>
 			) : (
 				<MainBodyText>{text.sentTextWithoutEmail}</MainBodyText>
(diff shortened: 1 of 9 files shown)


src/client/pages/IframedPasscodeEmailSent.tsx
```diff
@@ -38,7 +38,7 @@ const text: Text = {
 	title: 'Enter your code',
 	sentTextWithEmail: 'We’ve sent a temporary verification code to',
 	sentTextWithoutEmail:
-		'We’ve sent you a temporary verification code. Please check your inbox.',
+		'We’ve sent you a temporary verification code. Please check your inbox in a separate window.',
```
ioannakok, line 41: From the PR that introduced this component
* https://github.com/guardian/gateway/pull/3293
I understand this can only be shown on the web so I didn't use the `appName`.

AshCorr:
Theres also an `appClientId` query parameter that should always be set for the native apps, you could do something similar to how we set the registration platform where you can exchange that `appClientId` for an app label and then only show the text if the label is one of the apps? This way you don't need to change anything in MMA / DCR, and it would be consistent with other ways we customize the journey for apps?

https://github.com/guardian/gateway/blob/0dde32733e3eb8816d73d4e3ca42989dd70f5c2e/src/server/lib/registrationPlatform.ts#L23-L27

https://github.com/guardian/gateway/blob/0dde32733e3eb8816d73d4e3ca42989dd70f5c2e/src/shared/lib/appNameUtils.ts#L38

src/server/routes/passcode.ts
```diff
@@ -108,13 +110,21 @@ router.get('/passcode', (req: Request, res: ResponseWithRequestState) => {
 		}
 
 		if (!encrypedCookieState.passcodeUsed) {
+			const registrationPlatform = await getRegistrationPlatform(
+				res.locals.queryParams.appClientId,
+			);
+			const appName = isAppLabel(registrationPlatform)
+				? getAppName(registrationPlatform)
+				: undefined;
+
```
AshCorr: Could turn this into a shared function?

AshCorr (dismissed):

AshCorr (approved):

PR #3405, opened 2026-05-01
feat: Add alarms for newsletter email signup inactivity

AshCorr:
## What does this change?

Adds an inactivity alarm for our newsletter token signup route. This is the route that is used whenever a user is sent an email to verify their newsletter subscription, and then clicks the link to confirm their subscription.

Last week we had an incident where users were seeing a "Link Expired" error due to a bug in Identity API that went amiss as it did not trigger any alarms.

I've picked a window of 6 hours, looking at the incident period this would definitely have triggered the alarm. Unfortunately this route does not get much traffic, and is a bit sporadic, so its somewhat hard to have a more precisce window.

Deployed to CODE

### DIFF
--- a/cdk/lib/identity-gateway.ts
+++ b/cdk/lib/identity-gateway.ts
@@ -529,6 +529,30 @@ systemctl start ${app}
 		});
 		unsubscribeAllInactivityAlarm.addInsufficientDataAction(new SnsAction(alarmTopic));
 
+
+		const newsletterSignupInactivityAlarm = new GuAlarm(this, 'NewsletterSignupInactivityAlarm', {
+			snsTopicName: alarmTopic.topicName,
+			alarmName: `${alarmPriorities.P2} - ${app} ${stage} has had no successful newsletter signups in the last 6 hours - CDK`,
+			alarmDescription: 'No one has successfully signed up a newsletter in the last 6 hours.',
+			metric: new Metric({
+				namespace: 'Gateway',
+				metricName: 'ConsentToken::Success',
+				dimensionsMap: {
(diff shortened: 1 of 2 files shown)


cdk/lib/identity-gateway.ts
```diff
@@ -529,6 +529,30 @@ systemctl start ${app}
...
 		unsubscribeAllInactivityAlarm.addInsufficientDataAction(new SnsAction(alarmTopic));
 
+
+		const newsletterSignupInactivityAlarm = new GuAlarm(this, 'NewsletterSignupInactivityAlarm', {
+			snsTopicName: alarmTopic.topicName,
+			alarmName: `${alarmPriorities.P2} - ${app} ${stage} has had no successful newsletter signups in the last 6 hours - CDK`,
+			alarmDescription: 'No one has successfully signed up a newsletter in the last 6 hours.',
+			metric: new Metric({
+				namespace: 'Gateway',
+				metricName: 'ConsentToken::Success',
```
pvighi, line 539: the metric for a successful newsletter signup is called `ConsentToken::Success` ?

pvighi (approved):

cdk/lib/identity-gateway.ts
AshCorr: Yep! https://github.com/guardian/gateway/blob/0dde32733e3eb8816d73d4e3ca42989dd70f5c2e/src/server/routes/consentToken.ts#L31

Believe its because the route accepts a "consent token" in the URL which contains the details of the newsletter signup request.

PR #3434, opened 2026-07-01
IDAPI newsletters endpoint update

rBangay:
## What does this change?
Migrates the IDAPI newsletters calls from the legacy endpoints to the new(ish) version.

Part of the work to remove legacy API endpoints from [newsletter-api](https://github.com/guardian/newsletters-nx/tree/main/apps/newsletters-api) in the newsletters-nx repo.

### Gimmie some context
Gateway loads an Identity api endpoint which in turn is a proxy (kinda, it does change the shape of the response a little) for a newsletter-api endpoint. The Identity and in turn newsletter-api endpoint that it loaded (before this pr) was a legacy one, a newer API had been created a little while back but not all usages of the legacy version were migrated over to the new one. This pr is part of that migration work.

### Tidbit
The older newsletters IDAPI endpoint returned two fields that both contained the same value `exactTargetListId` and `listId`. They weren't different from each other, the underlying identity codebase serialised them both from the same source field ([exactTargetListId](https://github.com/guardian/identity/blob/dcb95f9bbd17a4346ecf55fbd172e9833614523a/identity-model/src/main/scala/com/gu/identity/model/Serialisation.scala#L249) and [listId](https://github.com/guardian/identity/blob/dcb95f9bbd17a4346ecf55fbd172e9833614523a/identity-model/src/main/scala/com/gu/identity/model/Serialisation.scala#L251)). The v2 newsletters endpoint just exposed the listId property. This codebase was referencing the `exactTargetListId` property, so I've switcheroo'd that over to use listId which should be a safe switch.

### Images

There shouldn't be any visual changes here.

disperse-lesley-nielsen


src/shared/__tests__/queryParams.test.ts
```diff
@@ -60,10 +60,10 @@ describe('addQueryParamsToPath', () => {
 			componentEventParams: 'componentEventParams',
 		};
 
-		const output = addQueryParamsToPath('/newsletters', input);
+		const output = addQueryParamsToPath('/consents', input);
```
rBangay, line 63: previously `/newsletters` was used arbitrarily to test the query parm functionality. Moving it over to another path like `consents` allows us to remove newsletters from the `ApiRoutePaths` type

src/server/lib/idapi/newsletters.ts
```diff
@@ -15,21 +15,28 @@ interface NewsletterAPIResponse {
 	id: string;
 	theme: string;
 	name: string;
-	description: string;
+	signUpDescription?: string | null;
+	signUpEmbedDescription?: string | null;
 	frequency: string;
-	subscribed: boolean;
```
ioannakok, line 20: This field was removed from the new newsletters API response? It's still in the model: https://github.com/guardian/gateway/blob/newsletter-api-update/src/shared/model/Newsletter.ts#L10

src/server/lib/idapi/newsletters.ts
```diff
@@ -15,21 +15,28 @@ interface NewsletterAPIResponse {
 	id: string;
 	theme: string;
 	name: string;
-	description: string;
+	signUpDescription?: string | null;
+	signUpEmbedDescription?: string | null;
```
ioannakok, line 19: What's the difference between these two descriptions?

src/server/lib/idapi/newsletters.ts
rBangay: Hey 👋  `subscribed` here looks like it was always an oversight. The property was never destructured/used anywhere. This newsletters API call is just concerned with non user specific newsletter info, eg what newsletters are available. The `subscribed` property on the `NewsLetter` interface that you linked to is populated by another (user specific) API call to `/users/me/newsletters` that gets information on which newsletters the user is subscribed to.

src/server/lib/idapi/newsletters.ts
rBangay: hey, they are used in different contexts, for example you would see `signUpDescription` if you are nosing through the newsletters on the all newsletters page (https://www.theguardian.com/email-newsletters) and `signUpEmbedDescription` for the sign up widget for the newsletter that you can find on article pages. `signUpDescription` maps directly to the old `description` property and `signUpEmbedDescription` maps to the old nested `emailEmbed.description` property

ioannakok (approved):

PR #3442, opened 2026-07-09
fix: Correctly announce "Show password" interactions with screen reader

AshCorr:
## What does this change?

Update the "Show password" button to announce its state to screen readers by modifying its `aria-label` and `aria-pressed` state whenever pressed.

Ran locally.

https://github.com/user-attachments/assets/255a34ba-0137-42da-bd58-a112a1ef6e69

### DIFF
--- a/src/client/components/PasswordInput.tsx
+++ b/src/client/components/PasswordInput.tsx
@@ -109,9 +109,10 @@ const EyeSymbol = ({
 			type="button"
 			css={buttonStyles}
 			onClick={onClick}
-			title="show or hide password text"
 			data-cy="password-input-eye-button"
-			aria-label="Show password"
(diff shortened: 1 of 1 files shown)


src/client/components/PasswordInput.tsx
```diff
@@ -109,9 +109,9 @@ const EyeSymbol = ({
 			type="button"
 			css={buttonStyles}
 			onClick={onClick}
-			title="show or hide password text"
```
AshCorr, line 112: The screen reader seemed to be prioritizing the title over the label, so I've removed in favour of just having a label

AshCorr: Actually just going to set both, as the title is used for the mouse tooltip

pvighi (approved):

PR #3445, opened 2026-07-09
Remove recaptcha from SOI page

ioannakok:
## Why?
We do not require Recaptcha on this page as by this point in the journey, the user has validated they own the email address they have provided.

This is most likely to hit privacy centric users, and users who are using ad-blockers/VPN settings without realising.

Deployed to CODE and checked network tab to make sure there are no recaptcha requests in the SOI page.

## Images

| Before | After |
|--------|-------|
| image | image |


AshCorr (approved):
LGTM!

PR #3452, opened 2026-07-23
Migrate openid from v5 to v6

brunocatarinoguardian:
## What does this change?

Updates openid-client v5 to v6. This is a complete rewrite and a considerable change where entire classes are removed and replaced, which resulted in some considerable changes to the code.

This was tested locally, but because we return `DevProfileIdClient` in dev and `ProfileOpenIdClient` in other environments, we should definitely test this in CODE to make sure it's all working before going to prod.

This has now been tested in CODE. Login with email and google, and register new email, all worked fine.

We use v6 instead of v5 and nothing is broken. We can still register and login as normal using any of the methods that currently work.

It's a risky change, because we're updating to a library version that completely rewrote the code. 

The code changes were made with assistance from Copilot, so while the AI is likely to catch edge cases a dev without in-depth knowledge of openid would not, there's also plenty of relying on AI, which is a risk.

## Accessibility


src/server/lib/okta/openid-connect.ts
```diff
@@ -134,99 +165,170 @@ export const ProfileOpenIdClientRedirectUris: OpenIdClientRedirectUris = {
...
+ *
+ * Replaces: `new Issuer(metadata)` + `new issuer.Client({...})` + `client[custom.http_options]`
  */
-const ProfileOpenIdClient = (ip?: string) => {
-	const client = new OIDCIssuer.Client({
-		client_id: okta.clientId,
+const buildProfileConfig = (
+	serverMetadata: ServerMetadata,
+	ip?: string,
+): Configuration => {
```
brunocatarinoguardian, line 198: `Issuer.Client` was replaced with `Configuration`, and that's what we use to create our own `OpenIdClient`

src/server/lib/okta/openid-connect.ts
```diff
@@ -104,7 +141,7 @@ const { okta, baseUri, stage } = getConfiguration();
  * we only expose the endpoints here
  */
 const issuer = joinUrl(okta.orgUrl, '/oauth2/', okta.authServerId);
-const OIDC_METADATA: IssuerMetadata = {
+const OIDC_METADATA: ServerMetadata = {
```
brunocatarinoguardian, line 142: Replaced from v5 to v6

src/server/lib/okta/idx/interact.ts
```diff
@@ -102,7 +106,8 @@ export const interact = async (
 		// set up the authorization parameters to send to the /interact endpoint
 		// in order to start the interaction code flow
 		// see https://developer.okta.com/docs/concepts/interaction-code/
-		const authorizationParams: AuthorizationParameters = {
+		// AuthorizationParameters type was removed in openid-client v6; use a plain Record instead
+		const authorizationParams: Record<string, string> = {
```
brunocatarinoguardian, line 108: AuthorizationParameters type was removed in openid-client v6, so have to use a record instead.

src/server/lib/okta/openid-connect.ts
```diff
@@ -53,23 +62,53 @@ export interface AuthorizationState {
...
+	id_token?: string;
+	refresh_token?: string;
+	token_type?: string;
+	expires_in?: number;
+	scope?: string;
+	/**
+	 * Returns the validated ID-token claims extracted from the token response.
+	 * Equivalent to v5's `tokenSet.claims()`.
+	 */
+	claims: () => IDToken;
```
brunocatarinoguardian, line 80: Equivalent to v5's `tokenSet.claims()`

src/server/lib/okta/openid-connect.ts
```diff
@@ -53,23 +62,53 @@ export interface AuthorizationState {
...
 }
 
+/**
+ * @interface OidcTokenSet
+ *
+ * Replaces the v5 `TokenSet` class. Wraps the v6 `TokenEndpointResponse` plain
+ * object and adds back a `claims()` helper so that callers in oauth.ts do not
+ * need to change how they access ID-token claims.
+ */
+export interface OidcTokenSet {
```
brunocatarinoguardian, line 73: Replaces the v5 `TokenSet` class

src/server/lib/okta/idx/interact.ts
akinsola-guardian: Did they have any other type to replace?

src/server/lib/okta/oauth.ts
```diff
@@ -156,7 +156,7 @@ export const performAuthorizationCodeFlow = async (
 		// otherwise we'll use the prompt parameter provided
 		prompt: idp && !loginHint ? 'login' : prompt,
 		// The sessionToken from authentication to exchange for session cookie
-		sessionToken,
+		sessionToken: sessionToken ?? undefined,
```
akinsola-guardian, line 159: Why do we now set undefined if sessionToekn is null?

src/server/lib/okta/idx/interact.ts
brunocatarinoguardian: Not that I could find out, because v6 is more functional and the authorizationUrl now is a function in the OpenIdClient object. As I remember it, AuthorizationParameters was converted to Record before being used anyway, so it's eliminating one transformation now.

src/server/lib/okta/oauth.ts
brunocatarinoguardian: Just for sake of handling it. The `authorizationUrl` on the client implementation filters out undefined. Could leave it as null but then I'd be filtering out null, so it would be the same, I believe (bear in mind my TS knowledge is very limited).

See [here](https://github.com/guardian/gateway/pull/3452/changes#diff-e78840bf846f2d94cb577c9c7d03fbfd24aa140cfcd2201738c520407855fb5cR232)

akinsola-guardian (dismissed):

src/server/lib/okta/openid-connect.ts
```diff
@@ -134,99 +164,139 @@ export const ProfileOpenIdClientRedirectUris: OpenIdClientRedirectUris = {
...
+	}
+
+	return config;
+};
+
+/**
+ * Wraps a v6 `Configuration` in the `OpenIdClient` interface that the rest of
+ * the application expects.
+ */
+const makeOpenIdClient = (config: Configuration): OpenIdClient => ({
```
ioannakok, line 233: As I understand, we were previously getting these four methods from the Client class which no longer exists in v6. How do you know how to implement them?

jest.config.js
```diff
@@ -19,15 +19,16 @@ module.exports = {
...
 	transformIgnorePatterns: [
-		// Don't try to transform any node_modules except Preact and @testing-library/preact
+		// Don't try to transform any node_modules except the packages we know ship ESM
+		// and are imported by our test graph.
 		// Since updating to Jest 30 trying to import these dependencies has resulted in
 		// unsupported ESM getting imported instead of CJS. Possibly because due to
 		// some siginificant changes in how Jest handles ESM under the hood?
 		// See:  https://jestjs.io/docs/upgrading-to-jest30#esm-module-support-and-internal-restructuring
-		'/node_modules/.pnpm/(?!(preact|@testing-library))',
+		'/node_modules/.pnpm/(?!(preact|@testing-library|openid-client|oauth4webapi|jose))',
```
ioannakok, line 28: Why did we have to make this change?

brunocatarinoguardian: It was some compatibility issues between commonjs and ESM. I don't know enough about JS ecosystem to say much more than that. This excludes these libraries so that it still works (I think it tries to import stuff with required and it doesn't like it).

I tried to move to ESM, but it was a nasty PR that I wasn't comfortable with, so this was the easiest / quickest solution. It's just tests here though, so shouldn't affect prod code.

src/server/lib/okta/openid-connect.ts
brunocatarinoguardian: Honest answer: I relied a bit too much on AI to fix this and I'm somewhat nervous about merging this PR, but it worked well in CODE, so it should be okay.

As to the implementation, I suspect it's based on the existing implementation of the previous client.

ioannakok: That's fine. Here's the previous code: https://github.com/panva/openid-client/blob/v5.x/lib/client.js#L17. I think we're missing some error handling so posting the link here in case you'd like to add anything. Leaving it up to you how much of it you want to use or not.

brunocatarinoguardian: I think the reason why it's confusing is that the Client interface doesn't exist anymore, but to maintain compatibility and minimise the changes, we're still using a similar structure.

The docs of what we need to do are [here](https://github.com/panva/openid-client#authorization-code-flow). There's those methods `buildAuthorizationUrl` and `authorizationCodeGrant` which are the only thing we actually need to call for authenticating, then `refreshTokenGrant` for refreshing. 

We could get rid of the `OpenIdClient` interface, but then would need to restructure the code differently and make more changes. Happy to do that, if you prefer.

src/server/lib/okta/openid-connect.ts
```diff
@@ -134,99 +164,144 @@ export const ProfileOpenIdClientRedirectUris: OpenIdClientRedirectUris = {
...
-): headers is OutgoingHttpHeaders => !Array.isArray(headers);
+const toOidcTokenSet = (response: TokenEndpointResponse): OidcTokenSet => {
+	const withHelpers = response as OidcTokenResponse;
+
+	return {
+		...response,
+		claims: (): IDToken => {
+			const claimsData = withHelpers.claims?.();
+			if (!claimsData) {
+				throw new TypeError('id_token not present in token response');
```
brunocatarinoguardian, line 176: Did another pass and this seemed to be missing, but everything else **seems** to be handled.

ioannakok (dismissed):

src/server/lib/okta/openid-connect.ts
brunocatarinoguardian: I don't think we need to validate anything extra in `authorizationUrl` method, because we pass the parameters as a `URLSearchParams` instead of a plainObject and definedEntries is a bidimensional array, which is one of the types allowed by the constructor. I think `buildAuthorizationUrl` does all the magic and we just really have to call it.

Even that filter for undefined we probably don't need it, but erring on the side of caution by keeping it.

src/server/lib/okta/openid-connect.ts
```diff
@@ -134,99 +164,144 @@ export const ProfileOpenIdClientRedirectUris: OpenIdClientRedirectUris = {
...
+const makeOpenIdClient = (config: Configuration): OpenIdClient => ({
+	authorizationUrl(params) {
+		const definedEntries = Object.entries(params).filter(
+			(e): e is [string, string] => e[1] !== undefined,
+		);
+		return buildAuthorizationUrl(config, new URLSearchParams(definedEntries))
+			.href;
+	},
+
+	callbackParams(req) {
```
brunocatarinoguardian, line 242: This one I used AI to help explain, so take with a grain of salt, but it sounds accurate. 

The original method in v5 is for a generic API that supports many scenarios we don't need. We only need to pass the parameters as they are, from my understanding. The only reason this method event exists is to maintain the same abstractions that v5 had to avoid changing too much code, otherwise we would probably have gone with a completely different approach.

src/server/lib/okta/openid-connect.ts
```diff
@@ -134,99 +164,144 @@ export const ProfileOpenIdClientRedirectUris: OpenIdClientRedirectUris = {
...
+		const response = await authorizationCodeGrant(config, callbackUrl, {
+			pkceCodeVerifier: checks.code_verifier,
+			expectedState: checks.state,
+		});
+
+		return toOidcTokenSet(response);
+	},
+
+	async refresh(refreshToken) {
+		const response = await refreshTokenGrant(config, refreshToken);
```
brunocatarinoguardian, line 269: This seems to also be the v5 class handling several scenarios we don't need. The library actually has a getDPoPHandle, but we're never doing anything with this in the old version, so we don't need it in the new version either. We could add these features to the new Client class, but since we don't need it, going YAGNI is a good idea.

src/server/lib/okta/openid-connect.ts
```diff
@@ -134,99 +164,144 @@ export const ProfileOpenIdClientRedirectUris: OpenIdClientRedirectUris = {
...
+		);
+	},
+
+	async callback(redirectUri, params, checks) {
+		const callbackUrl = new URL(redirectUri);
+		Object.entries(params).forEach(([k, v]) =>
+			callbackUrl.searchParams.set(k, v),
+		);
+
+		const response = await authorizationCodeGrant(config, callbackUrl, {
```
brunocatarinoguardian, line 260: The library does this weird thing where the callbackurl can be in the 3rd parameter and if it is, it overrides the 2nd parameter, but because we are passing the right value and no override in the 3rd parameter, we're fine. It's just excessively confusing, but the way we do it now removes that confusion.

The one thing that sent me down a rabbit hole was accounting for where `pkceCodeVerifier` and `expectedState` were used. They seemed to be completely ignored and no matter how much I looked at the code, I couldn't see it, even though the AI kept claiming it did. But it is used on a validation method in v6, so that's good.

**The only thing that seems to be missing is `response_type` being validated as `code`, which the old service did and now it doesn't, but I can't understand why it would be useful if it's always the same value... might need another pair of eyes tomorrow for this one.**

brunocatarinoguardian: From my understanding, there's a code in v6 that does
```
parameters.set(
  'response_type',
  hybrid ? 'code id_token' : implicit ? 'id_token' : 'code',
)
```
and this is defaulting to `code` because hybrid and implicit are both false (or undefined?) - so this case should be handled too.

ioannakok (approved):

PR #3453, opened 2026-07-27
Add new onboarding flow empty

brunocatarinoguardian:
## What does this change?

It includes a new empty Onboarding page - header only atm - but which will be used in the future to add the new sections in the new onboarding flow.

It also adds a new flag `newOnboardingFlow` and navigates to the `/welcome/onboarding` page from the `/welcome/complete-account` page when the flag is true.

Unit tests were written + tested locally pointing to CODE that both of the flows are behaving as expected.

Also tested in CODE.

Shouldn't have impact on errors, as it's a new flow. 

Only risk I can think of is the new rule having some unforeseen bug, but unit tests should mitigate that risk. 

It doesn't affect the existing website otherwise.

## Accessibility

### DIFF
--- a/src/client/pages/Onboarding.tsx
+++ b/src/client/pages/Onboarding.tsx
@@ -1,0 +1,12 @@
+import { MinimalLayout } from '@/client/layouts/MinimalLayout';
+
+export interface OnboardingProps {
+	shortRequestId?: string;
+}
+
+export const Onboarding = ({ shortRequestId }: OnboardingProps) => (
+	<MinimalLayout
+		shortRequestId={shortRequestId}
+		pageHeader="Welcome to the Guardian"
+	/>
+);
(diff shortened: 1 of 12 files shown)


src/client/pages/Onboarding.tsx
```diff
@@ -0,0 +1,12 @@
+import { MinimalLayout } from '@/client/layouts/MinimalLayout';
+
+export interface OnboardingProps {
+	shortRequestId?: string;
+}
+
+export const Onboarding = ({ shortRequestId }: OnboardingProps) => (
+	<MinimalLayout
+		shortRequestId={shortRequestId}
+		pageHeader="Welcome to the Guardian"
```
brunocatarinoguardian, line 10: Empty page for now. This will be updated as we do the rest of the tasks.

AshCorr (dismissed):

akinsola-guardian (dismissed):

akinsola-guardian (approved):

PR #3463, opened 2026-08-21
Openid v6 2ndattempt

brunocatarinoguardian:
## What does this change?

Reverts the revert of my v5 -> v6 changes (putting them back) with some extra logging to try to figure out why it's not working.

I can't reproduce the errors in CODE, so the only thing I can think of is deploying to PROD, let it run for a few minutes, and revert the deploy once we have some logs.

We figure out what the issue with moving from v5 to v6 is and fix it in a future PR.

There will be some users that will get errors on gateway, but last time it happened, it was a minority of users, so it shouldn't affect every user at once.

## Accessibility

### DIFF
--- a/src/server/lib/okta/openid-connect.ts
+++ b/src/server/lib/okta/openid-connect.ts
@@ -1,4 +1,15 @@
-import { Issuer, IssuerMetadata, Client, custom } from 'openid-client';
+import {
+	Configuration,
+	ServerMetadata,
+	buildAuthorizationUrl,
+	authorizationCodeGrant,
+	refreshTokenGrant,
+	customFetch,
+	allowInsecureRequests,
+	type IDToken,
+	type TokenEndpointResponse,
+	type TokenEndpointResponseHelpers,
+} from 'openid-client';
 import { randomBytes } from 'crypto';
 import { Request, CookieOptions } from 'express';
 import { joinUrl } from '@guardian/libs';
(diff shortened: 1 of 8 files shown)


src/server/lib/okta/openid-connect.ts
```diff
@@ -46,6 +46,7 @@ export type UserFlow =
  */
 export interface AuthorizationState {
 	stateParam: string;
+	createdAt?: number;
```
pvighi, line 49: sorry I'm not super familiar with this code but I can see AuthorizationState is used in multiple places including being read from cookies.
 Did you check that adding a new field to it won't cause any issues ?

brunocatarinoguardian: Hum that's a good shout, but I had a look and I think we should be fine. 
It's optional and we always verify for null before using it.

My main concern was saving it in the cookie and then having a "deserialization error" or something when we remove it later, but this is javascript. There's no such thing as deserialization errors... 😆

src/server/lib/okta/openid-connect.ts
```diff
@@ -255,6 +255,23 @@ const makeOpenIdClient = (config: Configuration): OpenIdClient => ({
...
+				if (typeof value === 'string' || value === undefined) {
+					return [];
+				}
+
+				return [[key, Array.isArray(value) ? 'array' : typeof value]];
+			}),
+		);
+
+		if (Object.keys(droppedParamTypes).length > 0) {
+			logger.warn('OAuth callback params contained non-string values', {
```
ioannakok, line 269: Why do we want to know whether params contained non-string values?

ioannakok (approved):

PR #3468, opened 2026-08-28
Iframes show t c

brunocatarinoguardian:
## What does this change?

Adds the Terms & Conditions text to the register and signin iframes.

The Sign-In iFrame was showing it together with social login buttons, so had to make sure that for multiple accounts we only show the t&cs without socials.

The Register was not showing it at all, so simply added it but only when it's multiple accounts, leaving every other flow unchanged - Emily is going to check if this omission is a problem, but that will be tackled later.

Wrote tests to verify these flows work as intended and show the elements we expect.

Also tested locally using the very useful `dev.ts` page with custom links to initiate these flows. And tested in CODE.

The T&Cs appear when it's multiple account sign in and registration.

## Accessibility

### DIFF
--- a/src/client/__tests__/IframedRegisterWithEmail.test.tsx
+++ b/src/client/__tests__/IframedRegisterWithEmail.test.tsx
@@ -1,0 +1,54 @@
+import { render, screen } from '@testing-library/react';
+import '@testing-library/jest-dom/jest-globals';
+import type { ComponentProps } from 'react';
+import { IframedRegisterWithEmail } from '../pages/IframedRegisterWithEmail';
+
+type IframedRegisterWithEmailProps = ComponentProps<
+	typeof IframedRegisterWithEmail
+>;
+
+const baseProps: IframedRegisterWithEmailProps = {
+	recaptchaSiteKey: '',
+	queryParams: {
+		returnUrl: 'https://www.theguardian.com/uk',
+	},
+};
+
+const setup = (extraProps?: Partial<IframedRegisterWithEmailProps>) =>
+	render(<IframedRegisterWithEmail {...baseProps} {...extraProps} />);
+
+const guardianTermsText =
+	'For more information about how we use your data, including the generation of random identifiers';
+
+test('shows GuardianTerms for multiple account flow', () => {
+	setup({
+		queryParams: {
+			returnUrl: 'https://www.theguardian.com/uk',
+			appClientId: 'maj',
+		},
+	});
+
+	expect(
+		screen.getByText(guardianTermsText, { exact: false }),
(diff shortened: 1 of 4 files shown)


src/client/pages/IframedSignIn.tsx
```diff
@@ -128,6 +128,32 @@ const showAuthProviderButtons = (
...
+	socialSigninBlocked,
+	queryParams,
+}: {
+	hideSocialButtons: boolean;
+	isMultipleAccountFlow: boolean;
+	socialSigninBlocked: boolean;
+	queryParams: QueryParams;
+}) => {
+	if (!hideSocialButtons) {
+		return showAuthProviderButtons(socialSigninBlocked, queryParams);
```
brunocatarinoguardian, line 143: This shows both social buttons and T&Cs, that's why I created a separate flow rather than reuse this. We don't want to show social login buttons on multiple accounts invites.

src/client/pages/IframedRegisterWithEmail.tsx
```diff
@@ -66,6 +79,7 @@ export const IframedRegisterWithEmail = ({
 			errorOverride={pageError}
 			overrideTheme="iframe-light"
 		>
+			{renderGuardianTerms(isMultipleAccountFlow)}
```
ioannakok: Just a suggestion: This would be a more React way of doing this since the logic is not too complex to extract it in a separate function.

```suggestion
			{isMultipleAccountFlow && (
				<InformationBox>
					<GuardianTerms />
				</InformationBox>
			)}
```

src/client/__tests__/IframedRegisterWithEmail.test.tsx
```diff
@@ -0,0 +1,58 @@
...
+	render(<IframedRegisterWithEmail {...baseProps} {...extraProps} />);
+
+const guardianTermsText =
+	'For more information about how we use your data, including the generation of random identifiers';
+
+test('shows GuardianTerms for multiple account flow', () => {
+	setup({
+		queryParams: {
+			returnUrl: 'https://www.theguardian.com/uk',
+			appClientId: 'maj',
```
ioannakok, line 31: This is sent by [support.theguardian.com](https://support.theguardian.com/) in this flow, right?

brunocatarinoguardian: I think so, although the guy who made the changes and test it was martech (which I don't think it's supporter revenue, so a bit confused). 

I tested the iframe content in isolation, but not the integration.

ioannakok (dismissed):

ioannakok (approved):

PR #3469, opened 2026-09-01
Revert "Revert "Openid v6 2ndattempt""

brunocatarinoguardian:
Reverts guardian/gateway#3467.

> Data seems to show the errors we're getting is only in the unsubscribe flows, and that they normalise after 15 minutes.
> 
> The theory is that we generate the link before deploying, so the path params are valid at that point, but once we move to v6 and redeploy, those path params become invalid, so any links generated before the deploy will fail and the user will have to retry.
> 
> The idea is to re-deploy these changes early in the day and monitor if the error rate is higher after the first 15 minutes or if it normalises. If it does, then the error is not the PR, but an invalid state between incompatible versions of v5 and v6 and possibly the way the code challenge is generated between versions.

Ignore that, last test showed that this problem might actually not be happening and we just had not one, but two coincidence episodes where we got an error spike when we  deployed.

If yday's test was accurate, it should be safe to merge this.

**Tests**

Tested login, sign-up, social login, logout in code.

### DIFF
--- a/src/server/lib/okta/openid-connect.ts
+++ b/src/server/lib/okta/openid-connect.ts
@@ -1,4 +1,15 @@
-import { Issuer, IssuerMetadata, Client, custom } from 'openid-client';
+import {
+	Configuration,
+	ServerMetadata,
+	buildAuthorizationUrl,
+	authorizationCodeGrant,
+	refreshTokenGrant,
+	customFetch,
+	allowInsecureRequests,
+	type IDToken,
+	type TokenEndpointResponse,
+	type TokenEndpointResponseHelpers,
+} from 'openid-client';
 import { randomBytes } from 'crypto';
 import { Request, CookieOptions } from 'express';
 import { joinUrl } from '@guardian/libs';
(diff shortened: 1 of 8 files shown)


src/server/lib/okta/openid-connect.ts
```diff
@@ -134,99 +163,161 @@ export const ProfileOpenIdClientRedirectUris: OpenIdClientRedirectUris = {
...
+				if (typeof value === 'string' || value === undefined) {
+					return [];
+				}
+
+				return [[key, Array.isArray(value) ? 'array' : typeof value]];
+			}),
+		);
+
+		if (Object.keys(droppedParamTypes).length > 0) {
+			logger.warn('OAuth callback params contained non-string values', {
```
ioannakok, line 258: Why do we want to know whether params contained non-string values?

brunocatarinoguardian: Ah this is a leftover from when I was testing things out. I wasn't sure that we were not dropping parameters, because openid v5 did a ton of things that we don't need, including support non-string parameters, like arrays and objects.

I don't think we are using it in reality, so our implementation at first didn't have any parameters other than strings. But because we "had" inexplicable errors, I added this to confirm we didn't have any non-string parameters being dropped.

According to the logs, we don't, for the time this was in prod, so I can remove this if you want.

ioannakok (approved):

src/server/lib/okta/openid-connect.ts
ioannakok: Thanks, makes sense. I would say let's remove it if we don't need it.

PR #3471, opened 2026-09-04
Show the right newsletter for each country.

brunocatarinoguardian:
## What does this change?

In the non-iframe flow, we show different newsletter options to different countries.

We want to do the same behaviour in the registration iframe. 

New unit tests added to verify we always show the right country newsletter, and that we never show feast and jobs in the newsletter.

Also tested in code with the VPN to different regions.

Asked for Supporter Revenue for help with testing the full flow and was able to verify in CODE that we get the right options. I didn't get the newsletters yet, but will verify I get the Saturday and First newsletters for the US during the weekend.

We allow US users to sign-up to US newsletters. Same for AUS, and any other country.

## Accessibility


src/client/pages/IframedRegisterWithEmail.tsx
```diff
@@ -106,7 +109,9 @@ export const IframedRegisterWithEmail = ({
 						autoComplete="off"
 					/>
 				)}
-				{isMultipleAccountFlow && <RegistrationConsents />}
+				{isMultipleAccountFlow && (
+					<RegistrationConsents geolocation={geolocation} />
```
brunocatarinoguardian, line 113: The RegistrationConsents component already does everything that needs to be done. I just added the missing parameter to do it.

ioannakok (approved):
Looks good to me!

PR #3472, opened 2026-09-04
Adds OnboardingPage component

ioannakok:
## What does this change?
Adds OnboardingPage component

Locally I ran the new onboarding flow from https://profile.thegulocal.com/

## Images

| mobile | tablet | desktop |
|--------|--------|---------|
| image | image | image |


src/client/components/OnboardingCard.tsx
```diff
@@ -0,0 +1,124 @@
...
+				{text}
+			</MainBodyText>
+
+			<div
+				css={css`
+					grid-area: cta;
+					padding: 0 0 ${remSpace[2]} ${remSpace[2]};
+				`}
+			>
+				<>
```
brunocatarinoguardian, line 83: Is this syntactic sugar for something?

src/client/pages/OnboardingPage.stories.tsx
```diff
@@ -0,0 +1,19 @@
+import React from 'react';
+import { OnboardingPage } from './OnboardingPage';
+import { Meta } from '@storybook/preact';
+
+export default {
```
brunocatarinoguardian, line 5: I don't know anything about Storybook, so don't know if there's standard behaviour there's given for free when you define a new file, but do we need to add more tests here?

brunocatarinoguardian (commented):
With my very limited / almost inexistent knowledge of React, it looks good, but yeah, take it with half a Kg of salt... :)

PR #3475, opened 2026-09-08
update dependencies

pvighi:
## What does this change?

update browserslist


brunocatarinoguardian (approved):
Just lock? No package.json needed?

pvighi:
> Just lock? No package.json needed?

no, just did pnpm update for the library and updated the lockfile only

PR #3476, opened 2026-09-08
Dependency updates

ioannakok:
## What does this change?
Patch and minor dependency updates. It also bumps @swc/plugin-emotion to a major version that is compatible with the latest version of @swc/core.

## How has this change been tested?

Deployed to CODE, created account, deleted account, reset password, subscribed to newsletter via /email-newsletters page.

PR #3477, opened 2026-09-08
Dependency updates

ioannakok:
## What does this change?
Patch and minor dependency updates. It also bumps @swc/plugin-emotion to a major version that is compatible with the latest version of @swc/core.

Deployed to CODE, created account, deleted account, reset password, subscribed to newsletter via /email-newsletters page.


cdk/lib/__snapshots__/identity-gateway.test.ts.snap
```diff
@@ -1067,6 +1067,28 @@ exports[`The IdentityGateway stack matches the snapshot 1`] = `
...
+          },
+          {
+            "Key": "Stack",
+            "Value": "identity",
+          },
+          {
+            "Key": "Stage",
+            "Value": "TEST",
+          },
+        ],
```
ioannakok, line 1091: See here: https://github.com/guardian/gateway/pull/3476#discussion_r3958031606

brunocatarinoguardian (approved):

PR #3478, opened 2026-09-08
Adds ability to define when to open external links in separate tabs, …

brunocatarinoguardian:
## What does this change?

During the multiple accounts testing, it was reported the terms and conditions links were not opening. Because we try to open these links inside an iframe, they don't seem to work, so the solution is to open them in a separate tab.

That's what this change does. It opens the terms links in a separate tab when a new parameter is passed on to the component.

We solve two other issues:
- iframe height not getting reduced in the IFramePasscode page.
- confirmation button not visible in web mobile, and no scroll is available, so user can't continue.

Wrote several new unit tests + testing in CODE.

Users can open both of the terms and conditions links when inside the iframe.

Other issues are also fixed.

## Images

Didn't take a picture of the before, but basically there was a huge white gap below the button and customer help email, because the first screen's height was much larger than the second screen's height, and the iframe was not being resized when we navigate. 

This is how it looks after fixed:

Screenshot 2026-09-10 at 16 44 51

## Accessibility


src/server/routes/dev.ts
```diff
@@ -2,6 +2,113 @@ import { Request } from 'express';
 import { rateLimitedTypedRouter as router } from '@/server/lib/typedRoutes';
 import { ResponseWithRequestState } from '@/server/models/Express';
 import { getApps } from '../lib/okta/api/apps';
+import { getConfiguration } from '@/server/lib/getConfiguration';
```
brunocatarinoguardian, line 5: This file only works in local development, so it won't affect prod. Basically all this code is creating an iframe of our own so that I could test the integration behaviour in a simple way without needing to deploy support-frontend. We should keep it for the future, as it allows us to test iframes in isolation in the future.

src/client/components/ExternalLink.tsx
```diff
@@ -3,11 +3,21 @@ import {
 	LinkButtonProps,
 	LinkProps,
 } from '@guardian/source/react-components';
-import React from 'react';
 import ThemedLink from '@/client/components/ThemedLink';
 
-export const ExternalLink = (props: LinkProps) => (
-	<ThemedLink {...props} rel="noopener noreferrer" />
+interface ExternalLinkProps extends LinkProps {
+	openInNewTab?: boolean;
```
brunocatarinoguardian, line 9: Allows to set whether to open a link in a new tab.
Default to false, since we decided to only use it for the iframe. The regular process outside of the iframe uses this set to false and open the link in the page where it is - reason is mobile. In mobile, this would likely break.

src/client/app.tsx
```diff
@@ -37,16 +37,16 @@ export const App = (props: Props) => {
 				styles={css`
 					${fontFaces}
 					html {
-						height: 100%;
+						height: ${isPageIframed ? 'auto' : '100%'};
```
brunocatarinoguardian, line 40: Most of the changes in this file fix a bug found during multiple account testing.

src/server/lib/middleware/helmet.ts
```diff
@@ -23,13 +23,10 @@ const CSP_VALID_URI = {
...
 const idapiOrigin = idapiBaseUrl.replace(/https?:\/\/|\/identity-api/g, '');
 
-// Dynamically build frame-ancestors based on stage.
-// This builds an array of urls that are appropriate for the current stage
-// Filter removes any falsey values
 const frameAncestors = [
 	stage === 'PROD' && 'https://support.theguardian.com',
 	stage === 'CODE' && 'https://support.code.dev-theguardian.com',
-	stage === 'DEV' && 'support.thegulocal.com',
+	stage === 'DEV' && (HELMET_OPTIONS.SELF || 'support.thegulocal.com'),
```
brunocatarinoguardian, line 29: Changes in this file allow the iframe to open a site that belongs to itself, plus it disables the cross origin stuff below.

src/server/routes/dev.ts
brunocatarinoguardian: Needless to say, this file was entirely written by AI, and it's low-risk, so might not be worth getting too deep into it. It's a dev tool, nothing that will run in prod, so should be okay.

src/client/app.tsx
```diff
@@ -37,16 +37,16 @@ export const App = (props: Props) => {
...
 						box-sizing: border-box;
-						${isPageIframed ? 'overflow: hidden;' : ''}
+						${isPageIframed ? 'overflow-x: hidden; overflow-y: auto;' : ''}
```
brunocatarinoguardian, line 42: This specific line fixes another bug found on mobile web where the confirmation button was hidden and there was no scroll to get there.

src/client/__tests__/ExternalLink.test.tsx
```diff
@@ -0,0 +1,42 @@
...
+
+test('ExternalLink renders with target="_blank" when openInNewTab is true', () => {
+	render(
+		<ExternalLink href="https://example.com" openInNewTab={true}>
+			Click me
+		</ExternalLink>,
+	);
+
+	// eslint-disable-next-line @typescript-eslint/no-unnecessary-type-assertion
+	const link = screen.getByText('Click me') as HTMLAnchorElement;
```
ioannakok:
```suggestion
	const link = screen.getByText('Click me');
```

You can do this. [`getByText`](https://testing-library.com/docs/queries/bytext/) returns an `HTMLElement` and is a parent of [`HTMLAlnchorElement`](https://developer.mozilla.org/en-US/docs/Web/API/HTMLAnchorElement).

src/client/__tests__/ExternalLink.test.tsx
```diff
@@ -0,0 +1,41 @@
...
+});
+
+test('ExternalLink renders without target="_blank" when openInNewTab is false', () => {
+	render(
+		<ExternalLink href="https://example.com" openInNewTab={false}>
+			Click me
+		</ExternalLink>,
+	);
+
+	// eslint-disable-next-line @typescript-eslint/no-unnecessary-type-assertion
```
ioannakok, line 28: Same as above. You can remove all the comments disabling this eslint rule in all the tests.

src/client/__tests__/IframedSignIn.test.tsx
```diff
@@ -72,3 +72,49 @@ test('does not show GuardianTerms when social buttons are hidden outside multipl
...
+	setup();
+
+	// eslint-disable-next-line @typescript-eslint/no-unnecessary-type-assertion
+	const termsLink = screen.getByText(
+		'terms and conditions',
+	) as HTMLAnchorElement;
+	expect(termsLink).toHaveAttribute('target', '_blank');
+});
+
+test('opens privacy policy link in new tab when social buttons are visible', () => {
```
ioannakok, line 83: Social buttons are visible in the non-multiple accounts flow, right? Were the T&C and privacy policy links not opening before this change even for the non-multiple accounts flow?

src/server/routes/dev.ts
```diff
@@ -2,6 +2,113 @@ import { Request } from 'express';
...
+	}
+	if (page === 'passcode') {
+		return '/iframed/passcode';
+	}
+	return '/iframed/signin';
+};
+
+if (stage === 'DEV') {
+	router.router.get(
+		'/iframe-harness.js',
```
ioannakok, line 21: Why do we need these extra two routes? I was able to open T & C and privacy policy links from the "Iframed Sign-in (Multiple Account)" and "Iframed Register (Multiple Account)". Isn't this enough for testing?

Image

brunocatarinoguardian: Those don't use the iframe, so I couldn't test one of the bugs found during testing - it didn't happen in those routes, as it had to do with the resize of the iframe due to its content. 

I added this to be able to test the full behaviour within an iframe locally, otherwise I'd have to be deploying to code every little change I did and checking if it worked - would have taken forever with trial and error.

src/client/__tests__/IframedSignIn.test.tsx
brunocatarinoguardian: Yeah this is [that confusing thing where we have two variables that basically do the same](https://github.com/guardian/gateway/pull/3478/changes#diff-ca6bb1122205b40a854c47eb97c139f442ca1ba4a7f7aac6c1a801bd0c9ec7baR142). The test is not new behaviour but to confirm that it didn't change previous behaviour.

ioannakok (approved):

src/server/routes/dev.ts
ioannakok: Thanks for the explanation. I didn't realise those were not using the iframe

